PR #2053, opened 2014-02-20
Coinbase

whit537:
*This is a redo of #2052 to bring the commits into the main repo so we can all hack on it.*

----

Hey there! This is the front-end code (and some back-end stuff) to add Coinbase support to Gratipay through Balanced.

## How it works

People will be able to donate USD gifts, but pay for them via Bitcoins in their Coinbase account.

Screenshots:

screen shot 2014-02-19 at 5 43 32 pm

Currently, everything looks the same here. We added a little note about the new feature, but you may or may not want that.

After choosing an amount and clicking 'confirm,':

screen shot 2014-02-19 at 5 44 46 pm

You have the option of paying via credit card or coinbase.

Choosing Coinbase:

screen shot 2014-02-19 at 5 45 25 pm

A popup window to OAuth to your coinbase account appears. After signing in:

screen shot 2014-02-19 at 5 46 24 pm

This says "Balanced" right now, but will say Gittip in the future. After confirming:

screen shot 2014-02-19 at 5 47 17 pm

Your gift is updated and and it's been connected. On the accounts page:

screen shot 2014-02-19 at 5 48 02 pm

You have the opportunity to see and remove or add the account.

## What works in this PR

Basically, I took screenshots of all of this locally, so that all works. Hooray!

To be clear, this means that nobody is actually receiving Bitcoins, because they are instantaneously converted to USD by Coinbase. This means that individuals receiving tips don't have to opt in, they won't even know how much of their tips were sent via BTC.

## What still needs to be done

The downside is.... it doesn't actually save this information in the Gittip backend. Some work needs to be done to actually save the funding instrument that comes back from Balanced, and charge from the correct account.

Balanced is obviously happy to help make this happen, but we don't know the Gittip codebase really well, so we should all pitch in with this!

This PR relies on https://github.com/gittip/www.gittip.com/pull/2036 landing before it can be merged. This PR is based on top of it, you can see all the other commits below.

Furthermore, it would be nice if https://github.com/gittip/www.gittip.com/issues/1165 could be figured out; this would allow someone to be able to tip in BTC itself, rather than tipping in a certain amount of BTC denominated in USD. We have some mocks for what that UI would look like if you're interested.

## Conclusion

:heart: :heart: :heart:

Let's hash out all the details and get this going!

<a name="todo" />
## To Do

- [x] Fix bitcoin address styling - #2065


Changaco:
This is obsolete, closing in favor of #2841.

PR #2780, opened 2014-09-23
Fix the broken layout on community

kaguillera:
Fix the broken layout on community also reduced size of font for amount for top givers and receivers so it matches font size of age of new members.

### DIFF
--- a/scss/modules.scss
+++ b/scss/modules.scss
@@ -26,7 +26,7 @@
         span.money {
             color: $black;
             display: block;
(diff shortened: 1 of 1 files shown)


clone1018:
This is going to hit a wall when https://github.com/gratipay/gratipay.com/pull/2938 releases

Changaco:
Closing in favor of 2100628579736f83e8ec4bf1217ecb077b08b080 in #2938.

Issue #2584, opened 2014-07-21
Improve the homepage for signed-in users


whit537:
On #2544 we changed the homepage. For anonymous users a static homepage is fine. For signed-in users we need some sort of dynamic timeline or whatevs at /.

Changaco:
Closing in favor of #3183 and #3184.

PR #2809, opened 2014-10-14
Improve the authenticated homepage

Changaco:
Remake of #2651 for #2584.


seanlinsley:
This is probably quite out of date after #2938. Is there work worth recovering from this PR?

Changaco:
I'm tagging this for review. This PR isn't good enough to close #2584 but it's better than what we have now, and it's been open for a long time so let's merge it and start another one for further improvements.

clone1018:
Anyone have a screenshot of this?

whit537:
@Changaco The reason I think this is a step backward from even the really bad homepage we have now is that this comes across as being like one of the other profile pages, but it is sufficiently *unlike* them to be very confusing. When I see this page my first instinct is to click on the subnav under the heading ... and there is no subnav under the heading. Furthermore, the teams and communities sections are duplicates of what we have on the actual profile page, heightening my cognitive dissonance. I don't see this adding value over the main profile page itself. For that reason, I think that a simple redirect to the profile page for authenticated users would be a better workaround than this page.

Changaco:
I've removed the avatar in the sidebar, plus the Teams and Communities sections. The problem now is that the page looks empty…

whit537:
I can live with this if you can, @Changaco.

screen shot 2015-02-16 at 9 15 03 pm

Changaco:
[IRC](https://botbot.me/freenode/gratipay/2015-02-17/?msg=32181530&page=1)

PR #2832, opened 2014-10-21
add a retention chart

whit537:
screen shot 2014-10-21 at 11 12 45 am


seanlinsley:
The benefit of using "out of 10,000" is you can have a round number instead of a fractional percent.

whit537:
I'm closing this PR. I'm glad we've gotten our feet wet with retention analysis, but adopting retention analysis as a key decision-making tool will make a lot more sense after we've mostly built the product we're trying to build. Right now we have a clear picture of what we're trying to build (well, an increasingly-clear; see #3220), and plenty of obvious work to do to get to "version 1." Let's revisit retention after version 1.

PR #2299, opened 2014-04-18
Add a funding rules UI

seanlinsley:
For #1063 

This is still very much a work in progress. Just wanted to get it off of my local machine and out into the fresh air.

PR #2824, opened 2014-10-19
Dogecoin support -- wow

rht:
Someone needs to do it. resolves #2685.

PR #2841, opened 2014-10-25
Add coinbase payin support

rohitpaulk:
Work in progess.

This is a redo of #2053 

> This pull request is too big. Gittip can only chew things in small bites. We need to break this up into smaller PRs.

- [x] Add UI to link coinbase account
- [x] Store coinbase result in DB
- [x] Set `is_funded` correctly by including `last_coinbase_result` too in `participant.py`. 
- [x] Same as above, in fake_payday.sql
- [x] What are the fees for coinbase transactions?
- [ ] Modify payday to first try a credit card, and then a coinbase account if that fails. If #2299 lands before this, this will have to include changes from there. 
- [ ] Modify the missing credit card warnings to display coinbase as an option.
- [ ] Update docs/about/faq wherever needed.


seanlinsley:
It doesn't look like this has been worked in a while, so I'm marking it as blocked.

rohitpaulk:
I'd like to work on this, but I'll need help on integrating the coinbase calls into payday. https://github.com/gratipay/gratipay.com/pull/2914 was a step in this direction, I'm not sure what's next.

cc: @Changaco

Changaco:
I didn't know this was waiting for me. This PR has the same problems as #2824. We don't want to keep adding columns like `last_coinbase_result` to `participants`. We need to generalize the code to handle any number of payment methods. I already have a local branch for that but I haven't done much work on it yet.

techtonik:
>
> We don't want to keep adding columns like last_coinbase_result to
> participants.
>
Does anybody use tools to see (or dump) the current structure of DB
visually? Some time ago looked for some command line tool that outputs into
.png in a predictable way, but didn't find any that worked for me.

rohitpaulk:
> I didn't know this was waiting for me.

That's my fault, I should've asked you once more after we implemented #2914. 

> We need to generalize the code to handle any number of payment methods.

Yes, that makes sense.

rohitpaulk:
Closing this (https://github.com/gratipay/gratipay.com/issues/3245).

PR #3022, opened 2014-12-16
Fix basic auth

Changaco:
This PR changes the credentials we require for HTTP Basic auth. It also fixes https://app.getsentry.com/gratipay/gratipay-com/group/45029121/ and https://app.getsentry.com/gratipay/gratipay-com/group/45029122/

### DIFF
--- a/branch.sql
+++ b/branch.sql
@@ -1,0 +1,3 @@
+BEGIN;
+    ALTER TABLE participants DROP CONSTRAINT participants_api_key_key;
+END;
--- a/branch.sql
+++ b/branch.sql
@@ -1,3 +1,4 @@
 BEGIN;
     ALTER TABLE participants DROP CONSTRAINT participants_api_key_key;
+    ALTER TABLE participants ADD COLUMN old_auth_usage date;
 END;
--- a/tests/py/test_user.py
+++ b/tests/py/test_user.py
@@ -112,29 +112,30 @@ class TestUser(Harness):
         assert SESSION in cookies
 
 
+    # from_id
 
-    # key token
-
-    def test_user_from_bad_api_key_is_anonymous(self):
-        user = User.from_api_key('deadbeef')
+    def test_user_from_bad_id_is_anonymous(self):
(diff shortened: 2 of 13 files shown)


seanlinsley:
I think we should:
- provide backwards compatibility without versioning like done in 7b19b1ed823f0751c3cbcf7c76f3dfc42fd4ffed
- log accounts that use the old version, and contact them
- once there's no one still using the old version, remove it

rohitpaulk:
^ :+1:

whit537:
Rebased on master.

whit537:
Rebased on master again.

Changaco:
I've added a column to keep track of when a participant last used the old auth. Ready for final review and merge.

branch.sql
```diff
@@ -0,0 +1,4 @@
+BEGIN;
+    ALTER TABLE participants DROP CONSTRAINT participants_api_key_key;
```
whit537: What's up with this? What was this constraint doing and why don't we need it anymore?

Changaco: It's the uniqueness constraint on `api_key`. We don't need it anymore because the new basic auth relies on user IDs for uniqueness.

tests/py/test_user.py
```diff
@@ -112,30 +112,31 @@ def test_session_is_regularly_refreshed(self):
...
-    def test_user_can_be_loaded_from_api_key(self):
+    def test_user_can_be_loaded_from_id(self):
         alice = self.make_participant('alice')
-        api_key = alice.recreate_api_key()
-        actual = User.from_api_key(api_key).participant.username
+        actual = User.from_id(alice.id).participant.username
         assert actual == 'alice'
 
 
-    def test_user_from_bad_id_is_anonymous(self):
```
whit537: Looks like this test was simply misnamed.

whit537:
@Changaco Clear to merge if my latest UI commit looks okay.

Changaco:
Looks okay.

PR #3047, opened 2015-01-01
Fix deadlock when trying to close an account

### LINKED ISSUE #3044: Cannot close (delete) my account
I works for a while but then I see an "Application Error" page from Heroku. Ostensibly the operation takes too long and Heroku kills the request.

Changaco:
Fixes #3044.

### DIFF
--- a/gratipay/models/_mixin_team.py
+++ b/gratipay/models/_mixin_team.py
@@ -172,7 +172,7 @@ class MixinTeam(object):
             new = new_takes.get(username, {}).get('actual_amount', Decimal(0))
             diff = new - old
             if diff != 0:
-                r = (self.db or cursor).one("""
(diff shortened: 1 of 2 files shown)


clone1018:
@Changaco Would I need a fake payday for this to matter? It worked locally form I just want to make sure I'm testing it right.

rohitpaulk:
When exactly does this happen? I'm trying to write a test for it, but can't seem to trigger that kind of behaviour. I created a team, added the person as a member, and then tried to close the account. What error should we be expecting?

Changaco:
It happens when the participant has a non-zero balance and is part of a team. I've modified the tests to trigger the problem.

Changaco:
Travis is green, the fix is trivial, I'm merging and deploying.

PR #3073, opened 2015-01-07
a couple more About tweaks

whit537:


whit537:
Okay, this is ready for review.

whit537:
Fund a bug in the test suite, dropping back to WIP.

whit537:
Back to Review.

whit537:
Blech, failure I thought I squished in 20073d3e0ccb0dd38e81ec88f145bb88626b8949 ...

whit537:
Alright, wtf is going on?

whit537:
Hypothesis: I masked an underlying issue with cb2f191864e58eda756bdeebb4563c7530e3edc9. I hypothesize this because before that commit I was getting the body for Do No Track from the request for features/index. What's up with that?

whit537:
**Test:** Drop back to a3db62e0a2312e0a016568ece99cbaf006d1980f and use pdb to observe more closely what was going on.

**Result:** False. I simply hadn't stepped far enough along and was still looking at the result from the previous loop iteration.

whit537:
**Hypothesis:** I recreated `TestPages.yml` incorrectly by running with only a subset of tests. Rerunning with all the tests in `test_pages.py` fails because we don't have all the data we need in `TestPages.yml`.

**Test:** Whack `TestPages.yml` and rerun.

whit537:
Okay, looks like that was it. Back to Review! :-)

cc: @clone1018

whit537:
Rebased on master.

PR #3080, opened 2015-01-08
allow running a single test file w/ pytest

seanlinsley:
As before, `make pytest` will run all of the Python tests.

Now you can do:
```sh
make pytest PYTEST="test_participant.py"
```
To run a single file.

Additionally, you can pass arguments to pytest:
```sh
make pytest PYTEST="test_participant.py -k about"
```


whit537:
Nice!

techtonik:
The only problem is that this stuff is not discoverable.

rohitpaulk:
@techtonik - You could add it to the Readme or IG :)

PR #3086, opened 2015-01-08
fix security email

whit537:
herp derp


whit537:
`zetawb -> zetaweb` is the important part. The rest is just better English.

whit537:
@seanlinsley @clone1018 This one should be trivial to review.

PR #3087, opened 2015-01-08
Fix `charge_volume`

Changaco:
Fixes #3070.

PR #3089, opened 2015-01-08
Fix DB check

Changaco:
Completes #3087.

### DIFF
--- a/gratipay/models/__init__.py
+++ b/gratipay/models/__init__.py
@@ -166,6 +166,7 @@ def _check_paydays_volumes(cursor):
                     and timestamp < ts_end
                     and amount > 0
                     and recorder is null
+                    and (status is null or status <> 'failed')
                 ) as ref
             from paydays
             order by id
@@ -183,6 +184,7 @@ def _check_paydays_volumes(cursor):
                     and timestamp < ts_end
                     and amount > 0
                     and recorder is null
+                    and (status is null or status <> 'failed')
                 ) as ref
             from paydays
             order by id


Changaco:
If you want to fix `charge_volume` in a local DB backup here's the fixed script:

```
BEGIN;

DO $$
DECLARE
    payday record;
    new_ncharges int;
    new_charge_volume decimal(35,2);
    new_charge_fees_volume decimal(35,2);
BEGIN
    FOR payday IN SELECT * FROM paydays LOOP
        CREATE TEMP TABLE our_charges AS
            SELECT *
              FROM exchanges
             WHERE "timestamp" >= payday.ts_start
               AND "timestamp" < payday.ts_end
               AND amount > 0
               AND recorder IS NULL
               AND (status IS NULL OR status <> 'failed');
        new_ncharges := (SELECT count(*) FROM our_charges);
        new_charge_volume := (SELECT COALESCE(sum(amount + fee), 0) FROM our_charges);
        new_charge_fees_volume := (SELECT COALESCE(sum(fee), 0) FROM our_charges);
        UPDATE paydays
           SET ncharges = new_ncharges
             , charge_volume = new_charge_volume
             , charge_fees_volume = new_charge_fees_volume
         WHERE id = payday.id
           AND (ncharges <> new_ncharges
                OR charge_volume <> new_charge_volume
                OR charge_fees_volume <> new_charge_fees_volume);
        IF FOUND THEN RAISE INFO 'updated payday %', payday.id; END IF;
        DROP TABLE our_charges;
    END LOOP;
END
$$;

END;
```

I've already fixed the production DB.

PR #3092, opened 2015-01-09
Clarifies that Gratipay doesn't take a cut of payments

colindean:
This is stated on the [FAQ - "How is Gratipay funded?"](https://gratipay.com/about/faq), but it's not very clear there, nor is it explicitly stated *anywhere* that Gratipay doesn't make money unless users pay it like any other account.

The FAQ probably could be updated, too, if the language I've proposed here is acceptable there, too.

### DIFF
--- a/www/about/index.spt
+++ b/www/about/index.spt
@@ -30,5 +30,13 @@ example = website.db.one("""
     and the <i>total</i> amount you receive is public (you can opt out of
     sharing this info).</p>
 
+    <p title="Gratipay does pass credit card processing fees to donors. See Features for more information.">
+    Gratipay does not take a cut of payments. Recipients receive the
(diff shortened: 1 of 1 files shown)


whit537:
Have you seen http://localhost:8537/about/pricing?

whit537:
Have you seen https://gratipay.com/about/pricing?

colindean:
Yeah. Hmm. I think maybe I should add a link to the pricing information in this blurb. I'll rework it a little.

colindean:
That's a little better, but I think it could be worded even better. I'll think about it for a bit and see what I can come up with.

Changaco:
Ping @colindean.

colindean:
Made some modifications. Better, but I could use some input.

Changaco:
Fixed the links and reworded.

whit537:
wfm

PR #3095, opened 2015-01-09
Adds call to action with "Give to tippee" when tip == 0

### LINKED ISSUE #3093: If I tip someone nothing, prompt me to Give/Donate, not Edit
I'm sitting here looking at a profile page, and there's not really a call to action. 

The left column has something that looks like it: big numbers showing what I give. Below it, it says Edit. OK. Maybe I want to edit that? What does that mean? \*shrug\*

Maybe if I'm already giving something to someone, leave it say "Edit". Otherwise, perhaps it should have a stronger call to action: "Give to `$username`" or "Donate to `$username`".

colindean:
The sole concern I have here is with long usernames overflowing the the column. It flows to a second line if the name is up to ~20 chars, but will look ugly past that.

Fixes #3093.

### DIFF
--- a/templates/your-tip.html
+++ b/templates/your-tip.html
@@ -12,7 +12,9 @@
             <div class="view">
                 <span class="amount">{{ format_currency(tip, 'USD') }}</span>
                 <div class="per-week">{{ _("per week") }}</div>
-                <button class="edit {{ 'not-zero' if tip > 0 }}">{{ _("Edit") }}</button>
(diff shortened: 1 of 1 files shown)


colindean:
screen shot 2015-01-09 at 12 54 34 am

colindean:
OK, I'll look into the Javascript to set that, but I probably won't get around to it until next week.

colindean:
#3166 resolved the above comment.

PR #3097, opened 2015-01-09
automate the weekly tweet

whit537:
Every week during [payday](http://inside.gratipay.com/howto/run-payday) we send out a tweet like this:

> Week 136—2,924 users exchanged $10,581: https://gratipay.com/about/stats.

https://twitter.com/Gratipay/status/553248050957664257

We want to automate that.

### DIFF
--- a/bin/tweet.py
+++ b/bin/tweet.py
@@ -1,0 +1,34 @@
+#!/usr/bin/env python
+# encoding: utf8
+from __future__ import absolute_import, division, unicode_literals, print_function
+from gratipay import wireup
+from gratipay.elsewhere.twitter import Twitter
+
+env = wireup.env()
+
+twitter = Twitter(
+    env.twitter_consumer_key,
+    env.twitter_consumer_secret,
+    env.twitter_callback,
+)
+
+status = "d whit537 Gratipay {npayday:,}—{nusers:,} users exchanged ${volume:,}: " \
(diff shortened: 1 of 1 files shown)


whit537:
API for tweeting is here:

https://dev.twitter.com/rest/reference/post/statuses/update

whit537:
But it looks like we can't pin the tweet programmatically. :-/

https://twittercommunity.com/t/pin-a-tweet-with-the-twitter-api/13048

benhc123:
Yes, that might need to stay manual.

galuszkak:
If it isn't posible through API to pin tweet then, we can always bypass this writing script to pin tweet through CasperJS or something similar.

whit537:
I'll see about using this during payday this week. We don't need to wait for tweet pinning before we land this.

whit537:
I'm getting a 403 with the latest commit and creds from a test app I made.

benhc123:
This might be more trouble than it's worth in terms of effort integrating with the Twitter API.

whit537:
There is no more weekly tweet.

tshepang:
Will it be re-instated? Why (or why not)?

whit537:
@tshepang I believe we shouldn't reinstate it, but we should talk about that. Reticketed as https://github.com/gratipay/inside.gratipay.com/issues/236.

PR #3101, opened 2015-01-09
Add feedback button

rummik:
This adds a feedback button to the side of the page which links to the issue tracker.

It should work back to IE9, which is within our support window (Current + 2 versions back)

### DIFF
--- a/templates/base.html
+++ b/templates/base.html
@@ -74,6 +74,8 @@
             </div>
         </div>
 
+        <a class="feedback" href="https://github.com/gratipay/gratipay.com/issues" target="_blank">Feedback</a>
+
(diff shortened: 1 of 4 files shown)


templates/base.html
```diff
@@ -74,6 +74,8 @@
             </div>
         </div>
 
+        <a class="feedback" href="https://github.com/gratipay/gratipay.com/issues" target="_blank">Feedback</a>
```
whit537: Per IRL discussion, let's go with `issues/new` here.

whit537: Decided not to go with `new` because that prompts for a login right away.

PR #3103, opened 2015-01-09
drop "atomic" metaphor

whit537:
It's better to use direct language for programming constructs, because metaphor introduces a level of interpretation that more easily leads to confusion.


whit537:
Rebased on master.

whit537:
Ready for review, @rummik. :-)

PR #3104, opened 2015-01-09
Add funding goal to side panel on profiles

rummik:
Right now the funding goal is obscured by being placed toward the bottom of the
page.  It's related to receiving amount, so it seems to want to have the number
there.


colindean:
+1

Changaco:
That's a simplistic solution, you've only duplicated the information.

whit537:
@Changaco What's wrong with that?

Issue #236, opened 2012-08-15
should be able to give non-anonymously


whit537:
~~Gittip~~ *Gratipay* does not divulge who exactly gives to whom. The reason we do this is to preserve the "no-strings-attached" nature of the gifts. You know how unfriending someone on Facebook can be a big deal? We wanted to avoid that kind of friction. We also don't want receivers to feel pressured by their larger donors. Like, "if I piss off so and so, I'll lose X dollars." Better not to know in the first place so you're not on eggshells.

Now, when we launched we were very focused on individual-to-individual giving, but since then we've started bringing groups onto ~~Gittip~~ *Gratipay*.

For giving **to individuals**, I expect us to always stay "anonymous in the particulars" to avoid the social friction mentioned above.

For giving **to groups**, I think we can relax the constraint, allowing it to be double opt-in. If you give to a company or organization or project, we should let you advertise that. If a company gives to another company, we should make it possible to see that.

**Update 2017-01-20**: And now as [Gratipay 2.0](https://gratipay.news/gratipay-2-0-2453d3c53077) we are _only_ for giving to groups, i.e., projects.

### Original

Turicas in IRC:

> why tips must be anonymous? I think it should be up to who donates to choose between being anonymous or not in a transaction

As a receiver, I don't really want to know who you are. I do of course have people telling me in person or by other means that they tip me. Maybe it would be opt-in on both sides?
##

rohitpaulk:
[From the retreat video (session 1)](https://www.youtube.com/watch?v=FDOLExNiBnY#t=4277)

Context: Anonymity as a weakness for **Individuals** giving **to Businesses**. 
 
> **Chad**: In the Individual -> Individual case it's purely altruistic. You don't need to know how I am, I don't care what you do with this money. In the Individual -> Business case, it's not purely altruistic. We've been saying no-strings attached for a couple of years, but that applies to the Individual -> Individual case.

whit537:
Thanks, @rohitpaulk! :)

whit537:
I think there was another place where I suggested that the individual to individual case may not even be purely altruistic (your mom gives you money but she still wants you to visit). But we don't necessarily need to get into the philosophy of altruism. :)

whit537:
> A concern with this, would be, well what if I discover Chad donates to political party I hate.

[...]

Issue #3060, opened 2015-01-05
"report user" button


whit537:
Implement an on-site mechanism to report users for [terms](https://gratipay.com/about/terms/) violations. Currently [generic contact channels](https://gratipay.com/about/contact) are the only option.
##

benhc123:
A form?

colindean:
Yeah. Got the idea from how Facebook does it. It could eventually be modal on the userpage and/or have multiple screens, but something flat is fine for now.

I'm getting this error when I submit the form:

```
Traceback (most recent call last):
  File "/Users/colin/Source/gratipay/gratipay.com/env/lib/python2.7/site-packages/algorithm.py", line 288, in run
    new_state = function(**deps.as_kwargs)
  File "/Users/colin/Source/gratipay/gratipay.com/env/lib/python2.7/site-packages/aspen/algorithms/website.py", line 141, in delegate_error_to_simplate
    response = resource.respond(request, dispatch_result, response)
  File "/Users/colin/Source/gratipay/gratipay.com/env/lib/python2.7/site-packages/aspen/resources/simplate.py", line 65, in respond
    response = self.get_response(context)
  File "/Users/colin/Source/gratipay/gratipay.com/env/lib/python2.7/site-packages/aspen/resources/simplate.py", line 219, in get_response
    response.body = render(context)
  File "/Users/colin/Source/gratipay/gratipay.com/env/lib/python2.7/site-packages/aspen/renderers/__init__.py", line 103, in __call__
    return self.render_content(context)
  File "/Users/colin/Source/gratipay/gratipay.com/env/lib/python2.7/site-packages/aspen_jinja2_renderer.py", line 66, in render_content
    return self.compiled.render(context).encode(charset)
  File "/Users/colin/Source/gratipay/gratipay.com/env/lib/python2.7/site-packages/jinja2/environment.py", line 969, in render
    return self.environment.handle_exception(exc_info, True)
  File "/Users/colin/Source/gratipay/gratipay.com/env/lib/python2.7/site-packages/jinja2/environment.py", line 742, in handle_exception
    reraise(exc_type, exc_value, tb)
  File "/Users/colin/Source/gratipay/gratipay.com/error.spt", line 1, in top-level template code
    from __future__ import absolute_import, division, print_function, unicode_literals
  File "/Users/colin/Source/gratipay/gratipay.com/templates/base.html", line 46, in top-level template code
[...]

Issue #1199, opened 2013-07-21
provide monthly invoices


whit537:
**was:** _make receipts look like software invoices_

So @wolfr can [deduct as a business expense](https://twitter.com/wolfr_/status/358959551602630657).

We have receipts. They look like this:

![receipt](https://f.cloud.github.com/assets/134455/832226/af6706e8-f23c-11e2-95d1-b4813124b40c.png)

What needs to look different?

##

whit537:
Invoices mailed on [FD1483](https://gratipay.freshdesk.com/helpdesk/tickets/1483) passed muster with the company in question. Let's do more of that! :-)

whit537:
Dribbble, for comparison:

 https://github.com/gratipay/gratipay.com/blob/68a0fc75990762875b2e93653313f39f881b9816/gratipay/billing/payday.py#L404-L450

whit537:
Dribbble, for comparison:

![screen shot 2015-01-12 at 1 01 26 pm](https://cloud.githubusercontent.com/assets/134455/5707980/37811006-9a5b-11e4-8aab-c5e460b648ae.png)

chrisdev:
@whit537 the "opt ins" are fine. 
The charitable organisation can always give explicit guidance to its supports to ensure select the correct options according to its requirements.

whit537:
+1 https://gratipay.freshdesk.com/helpdesk/tickets/1940

whit537:
> One of the things I do for money, is develop Varnish, and the Varnish Moral License is the vehicle I invented for the paperwork.
>
The way it works is very simple:
>
- Happy varnish users tell me that they want a Varnish Moral License
- I send them an invoice from my company
- They pay the invoice
- I develop Varnish

http://phk.freebsd.dk/VML/

(via https://github.com/gratipay/inside.gratipay.com/issues/732)

whit537:
To: “phk@” at this site
Subject: text of the Varnish Moral License?

> Greetings! For the past four years I've been working on a (weird, open-sourcey) startup called Gratipay, which, I am now [starting](https://github.com/gratipay/gratipay.com/issues/1199#issuecomment-233178150) to realize, may be little more than an attempt to productize your Varnish Moral License. An initial question as I come to terms with the implications of this discovery: What is the actual text of the VML that you send to companies that purchase one? If such a text exists, may I share it with the Gratipay community?

whit537:
[...]

PR #3112, opened 2015-01-13
Break anonymity

whit537:
This PR breaks anonymity with a double opt-in, closing #236.

 - [x] Design privacy permissions UI.
 - [x] Add db fields.
 - [x] Update json endpoint(s).
 - [x] Wire up js.
 - [ ] Review interaction of `is_singular` with `anonymous_receiving`.
 - [ ] Give givers (I guess I'm starting to call them supporters on this PR?) UI to divulge themselves if the receiver allows it. This should go in the tip CTA in the sidebar.
 - [ ] Update payday (etc.?) to set `transfers.is_anonymous` appropriately (double opt-in).
 - [ ] Fill out the Receiving page with a listing of everyone who has given what over time.
 - [ ] Add column(s) on the Giving page to indicate identity divulging status.
 - [ ] Update history page to include details as appropriate.
 - [ ] Wire up behavior when newly requiring non-anonymity: I guess we should prompt and then clear existing tips from people that don't have `tell_supporting`.


whit537:
New knob on Account page:

screen shot 2015-01-12 at 11 40 26 pm

Next steps:

 - [ ] Give givers (I guess I'm starting to call them supporters on this PR?) UI to divulge themselves if the receiver allows it. This should go in the tip CTA in the sidebar.
 - [ ] Update payday (etc.?) to set `transfers.is_anonymous` appropriately (double opt-in).
 - [ ] Fill out the Receiving page with a listing of everyone who has given what over time.

whit537:
I suggest we broaden the "Hide total {giving,receiving} from others" to encompass specific {giving,receiving} as well.

whit537:
P.S. I started this PR because both @weallhave24hrs and @renaantosz need it to really start using Gratipay in earnest, and they're the ones whose use cases I'm personally prioritizing right now. In other words I'm personally shifting from the ["Pay Your Heroes" roadmap](https://medium.com/inside-gratipay/roadmap-to-heroes-bcbd51c63b11), which was focused on the "individual giving to another individual" case, to a new roadmap focused on the "individual giving to a business" use case. This is based on the retreat, which I still need to blog about.

seanlinsley:
+1 in general. I'll work on reviewing this tomorrow

!m @whit537

www/%username/account/index.html.spt
```diff
@@ -202,6 +202,15 @@ emails = participant.get_emails()
             {% endif %}
         </p>
 
+        <h2>{{ _("Supporter Identity") }}</h2>
+        <p>
+            <label class="anonymous-supporters">
+                <input type="checkbox"
+                    {% if not participant.anonymous_supporters %}checked="true"{% endif %} />
+                    {{ _("Allow my supporters to divulge themselves.") }}
```
clone1018: I don't feel like this is very descriptive. If you want to keep it we should at least have hover state help text.

benhc123:
Completely agree with @clone1018, this functionality needs to be explained better in the labels.
Overall though, a great function. !m @whit537!

www/%username/account/index.html.spt
rummik: Just a thought:

Allow anyone:
  - [x] To see how much I give
  - [x] To see how much I receive
  - [ ] To search for me

Allow people I support:
  - [ ] To see I support them
  - [ ] To show I support them on their profile
  - [ ] To be shown on my profile

Allow people that support me:
  - [ ] To let me see they support me
  - [ ] To show they support me on their profile
  - [ ] To be shown on my profile

whit537:
I printed out and am rereading #236. The full implementation of #236 gets complicated. On this PR I want to focus on the minimum necessary for @weallhave24hrs @renaantosz, though I want to do it in a way that makes sense in terms of the bigger picture on #236, so that's why I'm rereading that.

whit537:
I also want to look at the specific use cases mentioned on #236. I'm thinking of @jdorfman @jedgar @ielectric etc.

iElectric:
It covers the use case for Kiberpipa.

whit537:
@iElectric Do you need to know how much someone is giving, or just that they're giving? Does it need to be public that they're giving to you or is it enough if you can see that info?

chrisdev:
@whit537 I've got a client https://gratipay.com/birdsongtt/  that is a government registered charity. They would like to give out receipts to donors for tax purposes. Knowing their donors is a prerequisite for this.
I've got a couple more like this coming on stream in the next couple of months. Hoping that this change fits this use case

whit537:
@chrisdev Do they need amounts given, or just identity of givers?

iElectric:
@whit537 
as Kiberpipa
We should see how much is someone donating to us
So that we can account for member fees.

whit537:
@iElectric Okay, thanks for the input.

chrisdev:
@whit537 ideally we need names and amount. But as a second best option the name is good enough

whit537:
@chrisdev Okay, it sounds like @iElectric needs amount, and I know that @renaantosz does as well. So let's consider amount a requirement for this PR to cover your use cases.

chrisdev:
:dancer:  @whit537

whit537:
Gosh, this is a complicated matrix. It's even more complicated than @balupton's helpful matrix at https://github.com/gratipay/gratipay.com/issues/236#issuecomment-24893623, because for @chrisdev's charities case to work we actually need the ability to *require* non-anonymity. I.e., "Only accept tips from people who are willing to divulge their identity to me."

whit537:
There's which side of the transaction you're on: 1) giver, 2) receiver.

There's what piece of info is shared: 1) identity, 2) amount, 3) reason.

There's where it's shared: 1) privately on the other's profile, 2) publicly on other's profile, 3) publicly elsewhere (community page? third-party site?).

There's sharing restrictions: only allow givers with XYZ configuration.

whit537:
We don't have tip reasons/messages yet and we shouldn't add that on this PR, so let's take that out of the consideration for now.

whit537:
Rebased on master.

whit537:
### Privacy

Allow everyone to see:
  - [x] my profile on search results
  - [x] the total amount I receive
  - [x] the total amount I give
  - [ ] who my supporters are
  - [ ] whom I support

Allow people I support to see:
(You can override this for each payment you set up.)
  - [x] who I am
  - [x] the amount I pay them

- [ ] Require my supporters to to show me who they are and how much they pay me.
- [ ] Allow my supporters to to show me who they are and how much they pay me.

whit537:
### Privacy

Allow everyone to see:
  - [x] my profile on search results
  - [x] the total amount I receive
  - [x] the total amount I give
  - [ ] who my supporters are (but not the specific amounts)
  - [ ] whom I support (but not the specific amounts)

Allow people I support to see:
(You can override this for each payment you set up.)
  - [x] who I am
  - [x] the amount I pay them

Who are my supporters?

- [x] I don't want to know who my supporters are.
- [ ] Allow my supporters to show me who they are and how much they pay me.
- [ ] Require my supporters to show me who they are and how much they pay me.

whit537:
Okay! Been going around and around the horn here at Blue Canary with @rummik @kaguillera @dmk246 @renaantosz @eileenford. https://github.com/gratipay/gratipay.com/pull/3112#issuecomment-71918901 represents the consensus we've come to over here about the privacy options we want to present to users with this PR.

whit537:
Here's a local iteration that was something we looked at during our discussion:

screen shot 2015-01-28 at 4 44 42 pm

clone1018:
Overload!

whit537:
> Overload!

@clone1018 Yeah, we moved beyond that iteration but I wanted to post it since it was part of our convo.

Here are the four options we looked at:

  1. https://github.com/gratipay/gratipay.com/issues/236#issuecomment-24893774
  1. https://github.com/gratipay/gratipay.com/pull/3112#discussion_r23107496
  1. https://github.com/gratipay/gratipay.com/pull/3112#issuecomment-71922937
  1. https://github.com/gratipay/gratipay.com/pull/3112#issuecomment-71918901

We analysed them according to the analysis at https://github.com/gratipay/gratipay.com/pull/3112#issuecomment-71732435. We landed on option 4.

Some points:

 - We decided to prune the possibility of displaying specific payment amounts publicly, because it didn't seem to be an option that would be popular, and pruning it reduced cognitive/UX overhead.
 - We talked a fair bit about framing these options as opt-in vs. opt-out ("Share"/"Show"/"Allow" vs. "Hide"/"Deny"/"Disallow"). We decided to frame as opt-in because of habituation from other sites (Facebook, Google+), but to use "Allow" instead of "Share" to try to hint more at user control of the experience.
 - The "Who are my supporters?" section would be a radio. The rest would be checkboxes.
 - It would be a bit odd to check "who my supporters are" without also checking "Allow my supporters ...," but at this point I think we should allow that combo (would show identity but not amount on my receiving page, both publicly and to myself).

whit537:
Also:

 - We decided not to tackle the "don't allow anyone to support me who also supports XYZ" case. @rummik's suggestion was that we handle that as part of a "block user" implementation (cf #3060).

whit537:
@iElectric @chrisdev Can I ask you to take a look at https://github.com/gratipay/gratipay.com/pull/3112#issuecomment-71918901 and let us know if that set of  permissions will give you what you need for Kiberpipa and Birdsong?

whit537:
One thing that's not possible with https://github.com/gratipay/gratipay.com/pull/3112#issuecomment-71918901 is allowing other supporters to see their fellow supporters, without allowing *everyone* to also see who the supporters are. Let's let that use-case go for now since I believe the initial demand for this feature doesn't require it.

whit537:
Hey @wrought @sudoroom, may I loop you in on this conversation? We're working on enabling Gratipay receivers to either allow or require their supporters to divulge who they are and what amount they give. **tl;dr** Is this feature of interest to you, and will the implementation prototyped at https://github.com/gratipay/gratipay.com/pull/3112#issuecomment-71918901 give you what you need in this regard?

whit537:
Another case I don't think is handled by https://github.com/gratipay/gratipay.com/pull/3112#issuecomment-71918901 is if a company gives to an individual and requires that the individual they're paying know that the payment is coming from their company. So far on #1199 we've been able to avoid this for the case or two where we've hacked an invoice, but as we build out invoices we may run into resistance around the anonymity of our current self-billing plan.

timothyfcook:
awesome!

whit537:
I simplified the "Who?" section by making it entirely about both identity *and* amount, rather than have the second two options include amount but not the first. This rationalizes the situation in point 4 of https://github.com/gratipay/gratipay.com/pull/3112#issuecomment-71926749.

From:

screen shot 2015-01-30 at 10 32 21 am

To:

screen shot 2015-01-30 at 10 32 04 am

whit537:
Current status:

screen shot 2015-01-30 at 10 49 19 am

whit537:
screen shot 2015-01-30 at 11 02 18 am

whit537:
Privacy preferences UI is ready for review. Now to make the preferences actually change things. :-)

whit537:
Do we need "who I support" and "who supports" me to be overridable per-tip? We might.

whit537:
Is this what we need, then?

screen shot 2015-01-30 at 3 19 35 pm

whit537:
The build is breaking at Travis because we don't have the json type in Postgres?

whit537:
Restarted, now seeing failures properly (due to inversion of privacy booleans). Also seeing failures locally.

rummik:
I'm not sure on the `Who pays you what, exactly?` text, since it feels like the wording is too different from the above options.  I'd suggest something more along the lines of:

Information about your supporters:
_Who they are, and how much they pay you._
- [x] Do not show me information about my supporters.
- [ ] Allow my supporters to choose.
- [ ] Require my supporters to tell me.

whit537:
How about this for the tip override UI?

screen shot 2015-02-02 at 11 13 39 am

whit537:
> the wording is too different from the above options

This section *is* different from the above options. If anything, we should set it off with a different heading. We raised but didn't settle the question of whether this is even properly a "privacy" option. It's not about my privacy, it's about my supporter's privacy. "Sharing" could encompass both.

whit537:
As I'm implementing the per-tip overrides, it occurs to me that the implementation at https://github.com/gratipay/gratipay.com/pull/3112#issuecomment-72275370 has it that I'll be able to individually set who of my supporters is displayed publicly. I guess that happens on the receiving page? We can also email people when they get new supporters. I guess `public_supporters` encodes the default, then.

whit537:
Or is that too complicated? Should we give both giver and receiver the ability to override individual tips? Or should we just give that ability to givers?

whit537:
What if you require knowledge and then unrequire it? Do you get your former supporters back? Maybe within a certain timeframe?

rohitpaulk:
> What if you require knowledge and then unrequire it? Do you get your former supporters back? Maybe within a certain timeframe?

I think the simplest way to implement this would be to handle it like we handle changes from a Team to Individual account. i.e When they click on 'require details', show them a warning that says that all their anonymous tips will be cancelled if they make this change.

rummik:
@whit537 What I mean is that I feel like the abrupt question starting the section isn't very clear, and adds confusion to it, since we don't do that anywhere else

synconme:
Just a suggestion regarding the options, maybe it is more convenient to simplify options a little bit, something like that:

Show other users:
* - [ ] My profile
* - [ ] Who I support
  * - [ ] Total amount I give
* - [ ] Who supports me
  * - [ ] Total amount I receive

-------
- [ ] Require my supporters tell me who they are and amount they give

From my point of view, if receiver wants to know who donates her, it is more likely she want to know the amount as well.

whit537:
I'm scrapping this for parts.

Issue #2543, opened 2014-07-03
Send an email when people you pledge to join Gratipay


seanlinsley:
It'd be nice if we sent an email to anyone who pledges to someone when that person joins Gittip. For example, I've pledged to https://www.gittip.com/on/twitter/hardcorehistory/. I'd love to get an email if they ever sign up.

Depends on #2312, etc.

rohitpaulk:
Done in #3136.

whit537:
Great work @rohitpaulk @Changaco! Our first notification! :dancer:

PR #3136, opened 2015-01-23
Notify patrons via email when user opts-in

rohitpaulk:
#2543 

- [x] Add opt-out checkbox on account page
- [x] Handle patrons without `email_address`
- [x] Tests
- [x] Review email content
- [x] What if we're sending emails to 100 patrons? Don't block request on sending emails
- [x] What if the account was taken over by another one?


whit537:
Nice! Our first transactional message! :dancer: 

We'll need to include opt-out with this. Probably a "Email Notifications" section on the Account page, parallel to "Privacy." Eh?

rohitpaulk:
Current status: 

screenshot from 2015-01-24 12 07 16


screenshot from 2015-01-24 12 07 28

Changaco:
286c19fa08c4a0c2ce6da8c17d620fe465a09fda seems to be unnecessary, I suggest dropping it from this PR.

rohitpaulk:
> 286c19f seems to be unnecessary

Yes, that's true. I initially thought that we could batch the emails in one call (have `_render_email` called just once in `send_emails_to`), but later realized that we had to identify the language for each user before sending an email.

> I suggest dropping it from this PR.

Even though it's functionally the same, I think it looks cleaner.

Changaco:
> Even though it's functionally the same, I think it looks cleaner.

It looks a little weird to me. If this code was on master and I was working on an email PR I would be wondering why it's been cut into two functions this way, and I would probably end up refactoring it.

gratipay/models/participant.py
```diff
@@ -631,29 +635,62 @@ def remove_email(self, address):
...
+            target=self.send_emails_to,
+            name='email',
+            args=(patrons, 'notify_patron'),
+            kwargs={
+                'check_for': 'notify_on_opt_in',
+                'elsewhere_username': elsewhere.user_name,
+                'elsewhere_platform': elsewhere.platform,
+                'profile_url': self.profile_url
+            }
+        ).start()
```
Changaco: I don't think that's a good idea, if we need an email thread then there should only be one, and it would be better if the queue was persistent so notifications don't get lost if the app crashes.

js/gratipay/account.js
```diff
@@ -135,6 +135,25 @@ Gratipay.account.init = function() {
...
+        jQuery.ajax(
+            { url: '../email-notifications.json'
+            , type: 'POST'
+            , data: {toggle: field}
+            , dataType: 'json'
+            , success: function(data) {
+                $(e.target).attr('checked', data[field]);
+            }
+            , error: function() {
+                Gratipay.notification("Failed to change your notification preference. Please try again.", 'error');
```
Changaco: i18n: please don't put error messages in the JS code, instead send the translated error message in the response.

gratipay/models/participant.py
rohitpaulk: @Changaco - That's the best I could do, I'll let you take over from here. :)

Changaco:
Ready for review.

emails/notify_patron.spt
```diff
@@ -0,0 +1,19 @@
+{{ _("{0} from {1} has joined Gratipay!", user_name, platform) }}
+
+[---] text/html
+{{ _("Your pledge to give {0} every week to {1} will be turned into action now that they have joined Gratipay.",
+     format_currency(amount, 'USD'), user_name) }}
+<br><br>
+<a href="{{ profile_url }}" style="color: #fff; text-decoration:none; display:inline-block; padding: 0 15px; background: #396; font: normal 14px/40px Arial, sans-serif; white-space: nowrap; border-radius: 3px">{{ _("View {0}'s profile", user_name) }}</a>
```
whit537: Can we abstract the button styles in emails? I'd like to be able to tell by looking at this code whether it'll result in the same button we use for the verification email.

whit537: Actually, let's only use buttons for action-oriented emails. This is an informational email, so let's use a link for the profile instead of a button. Done in 214907a970162c7c7ee828cc38fe57be69bf8d7c.

whit537:
Content reviewed:

screen shot 2015-02-19 at 1 45 44 pm

whit537:
We should use `GitHub` in the subject instead of `github`. We have `display_name` available on the platform object.

whit537:
Otherwise than that nit, this PR works for me. !m 8

rohitpaulk:
Other than https://github.com/gratipay/gratipay.com/pull/3136#commitcomment-9828767 (Isn't a blocker, we're nowhere close to 30 emails per minute - it's just related to us being able to estimate our limit), this PR looks good to me. :)

Changaco:
@rohitpaulk 0cd858db26d97c7e7c2a1504934202768946d73f

rohitpaulk:
I've set `DEQUEUE_EMAILS_EVERY` to `60` in production.

Issue #3115, opened 2015-01-13
Nicer error pages on search results needed


benhc123:
E.g. https://gratipay.com/on/twitter/adhsjakdjsdkjsajdhksda/
https://gratipay.com/on/facebook/adhsjakdjsdkjsajdhksda/
https://gratipay.com/on/github/adhsjakdjsdkjsajdhksda/
https://gratipay.com/on/bitbucket/adhsjakdjsdkjsajdhksda/

I'm not sure if this info is being fetched via JSON or something else but we should do something better with these error responses from APIs.

clone1018:
Haha, awesome.

!m @benhc123

kaguillera:
I am working on this but I was want to know if the intention of this issue is to improve the error pages for only the results of the Search or the error pages in general when a response other than a 200 is to be presented via the current error.spt

Changaco:
Done in #3146.

PR #3146, opened 2015-01-28
3115 better search pages error

kaguillera:
Please review for #3115

### DIFF
--- a/tests/py/test_pages.py
+++ b/tests/py/test_pages.py
@@ -192,16 +192,16 @@ class TestPages(Harness):
 
     def test_search_results_error_pages_twitter(self):
         body = self.client.GxT("/on/twitter/adhsjakdjsdkjsajdhksda/").body
-        assert "User does not exist on Twitter" in body
+        assert "The user adhsjakdjsdkjsajdhksda does not exist on Twitter" in body
 
     def test_search_results_error_pages_facebook(self):
         body = self.client.GxT("/on/facebook/adhsjakdjsdkjsajdhksda/").body
-        assert "User does not exist on Facebook" in body
+        assert "The user adhsjakdjsdkjsajdhksda does not exist on Facebook" in body
 
     def test_search_results_error_pages_github(self):
         body = self.client.GxT("/on/github/adhsjakdjsdkjsajdhksda/").body
-        assert "User does not exist on GitHub" in body
+        assert "The user adhsjakdjsdkjsajdhksda does not exist on GitHub" in body
 
     def test_search_results_error_pages_bitbucket(self):
         body = self.client.GxT("/on/bitbucket/adhsjakdjsdkjsajdhksda/").body
-        assert "User does not exist on Bitbucket" in body
+        assert "The user adhsjakdjsdkjsajdhksda does not exist on Bitbucket" in body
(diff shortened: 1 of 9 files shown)


tests/py/test_pages.py
```diff
@@ -166,3 +166,26 @@ def test_new_participant_can_edit_profile(self):
         self.make_participant('alice', claimed_time='now')
         body = self.client.GET("/alice/", auth_as="alice").body
         assert b'Edit' in body
+
+    def test_search_results_error_pages_twitter(self):
+<<<<<<< Updated upstream
```
whit537: You've got a merge conflict here (the `<<<<<<<`). Please resolve.

Changaco:
I don't think messing with `error.spt` is the right approach, I think should modify the `get_account_elsewhere()` function instead.

kaguillera:
I removed the code from error.spt and rewrote a more reasonable error statement in the `check_api_response_status()` method since that is where the Response is raised first when a user does not exist on a platform. I guess it can also be done in the `get_account_elsewhere()` as you suggested @Changaco but that `Response(404)' is only raised when the no platform is given

Changaco:
`check_api_response_status()` is a generic method, it's not limited to user lookups, changing the error message there is incorrect.

kaguillera:
I  noticed that @Changaco but that is the place where the 404 response from the platform APIs request is being caught and the raising of the aspen Response. I will look around again an see if I can catch it before. if you have any other suggestion I am open to it. As I said before

kaguillera:
Did not finish... `get_account_elsewhere()` is not reached because `check_api_response_status()` is called first in the case of the search.

kaguillera:
I finally got to capture and raise the 404 Response in `get_account_elsewhere()'  @Changaco @whit537 could you review and tell me what you think.

kaguillera:
Travis keeps failing can anyone tell me why that is?

Changaco:
This is still not correct: you're turning all responses into 404s, even the ones that aren't actually 404s.

kaguillera:
Thanks did not realize that. Will fix it.

kaguillera:
I figured out the function in the test_elsewhere.py test script that seems to be affecting the test function in test_pages.py that I have been working on. It is the test_user_pages(self): function. My "educated guess” is that after that function does a call to get_user_info it is not successfully cleared and thus retaining the user information in session/cache or something. I was wondering is you could point me in the right direction of if my guess is right and if so who can I clear/reset thing so that I have a clean call to the platform API to get the proper response.

whit537:
> My "educated guess” is that after that function does a call to get_user_info it is not successfully cleared and thus retaining the user information in session/cache or something.

@kaguillera In this situation I find it helpful to throw a `set_trace` in the test so I can inspect the database and in-memory caches to find the bit of info that's not being cleared. Off the top of my head I don't have a strong sense of what it might be that's not being cleared. I think we still have some in-memory structures related to signing in via elsewhere account? @Changaco @rohitpaulk @clone1018 @seanlinsley Do you remember where these are?

Changaco:
Ready for review.

@kaguillera The fix you were looking for is in 4f69b5b470290ac7bde1bd8e98ee70e723196d57.

kaguillera:
@Changaco Please review, is this failure because of something I did?

Changaco:
Rebased on master.

@kaguillera Yes, the merge you did wasn't good.

Changaco:
Rebased on master again.

kaguillera:
Yep looks good, well done Changaco.

Changaco:
I've added a couple more commits for i18n.

whit537:
@kaguillera Since @Changaco has implicitly signed off on your commits, then if you are okay with his commits then it is your prerogative to merge.

Issue #2913, opened 2014-11-08
Fix error page


Changaco:
https://app.getsentry.com/gratipay/gratipay-com/group/49625026/ (was https://app.getsentry.com/gratipay/gratipay-com/group/39866655/)

whit537:
I infer from https://github.com/gratipay/gratipay.com/commit/ac9077437ed59a485105edd92d3c7fd83b9518ed that the problem here was that i18n wasn't configured before an 403 was triggered in `get_auth_from_request`, ya?

Changaco:
@whit537 Correct. `error.spt` extends `templates/base.html` which uses i18n functions.

whit537:
k thx

PR #3151, opened 2015-02-04
Get rid of configure-aspen.py

Changaco:
PR for a branch started by @whit537 3 months ago. I've added commits to fix the tests.


Changaco:
Added a fix for #2913.

Issue #3152, opened 2015-02-04
switch to autoescape


whit537:
Reticketing from #722. See there for initial notes (should copy over here). It's entangled with i18n.

Changaco:
Do we want [partial](https://github.com/gratipay/gratipay.com/compare/autoescape) or [full](https://github.com/gratipay/gratipay.com/compare/autoescape-alt) autoescape ?

whit537:
[IRC](https://botbot.me/freenode/gratipay/2015-02-06/?msg=31421995&page=1)

whit537:
Digesting the difference between partial and full.

In both cases, we are modifying the two text-related i18n functions, `get_text` and `n_get_text`. The crucial difference I see is between:

```python
from markupsafe import Markup
return Markup(s.format(*a, **kw))
```

and

```python
from markupsafe import escape
return escape(s).format(*a, **kw)
```

One operates on the string after it is formatted, the other on the string before it is formatted.

whit537:
What is `Markup`? Why are we using it there instead of `escape` in both cases?

whit537:
http://www.pocoo.org/projects/markupsafe/

whit537:
[`escape` returns a `Markup`](https://github.com/mitsuhiko/markupsafe/blob/0.23/markupsafe/_native.py#L15-L28)

whit537:
```python
>>> from markupsafe import Markup, escape
>>> Markup('foo')
Markup(u'foo')
>>> escape('foo')
Markup(u'foo')
>>> Markup('<foo>')
Markup(u'<foo>')
>>> escape('<foo>')
Markup(u'&lt;foo&gt;')
>>>
```

So calling `Markup` in python is equivalent to using the `|safe` filter in Jinja2? And `escape` is equivalent to `|e`?

Changaco:
Yes.

whit537:
k

whit537:
Alright, so we know the `s` strings are all safe, because we write those. That leaves the `a` and `kw`. Those could potentially be user input that we'd want to escape, but in a number of cases we are passing our own markup in, which is safe.

[...]

Issue #722, opened 2013-03-12
review site for html injections


whit537:
I just discovered that I wasn't escaping statements. :-(

I'm not sure how that happened, but it did. We need to review the whole site for html injections.

whit537:
Here's all the places in our templates where we use Jinja2 text interpolation:

https://gist.github.com/whit537/755c59aec14335b019b9

whit537:
> Why is Autoescaping not the Default?
>
> There are multiple reasons why automatic escaping is not the default mode and also not the recommended one. While automatic escaping of variables means that you will less likely have an XSS problem it also causes a huge amount of extra processing in the template engine which can cause serious performance problems. As Python doesn’t provide a way to mark strings as unsafe Jinja has to hack around that limitation by providing a custom string class (the Markup string) that safely interacts with safe and unsafe strings.
> 
> With explicit escaping however the template engine doesn’t have to perform any safety checks on variables. Also a human knows not to escape integers or strings that may never contain characters one has to escape or already HTML markup. For example when iterating over a list over a table of integers and floats for a table of statistics the template designer can omit the escaping because he knows that integers or floats don’t contain any unsafe parameters.
> 
> Additionally Jinja2 is a general purpose template engine and not only used for HTML/XML generation. For example you may generate LaTeX, emails, CSS, JavaScript, or configuration files.

http://jinja.pocoo.org/docs/dev/faq/#why-is-autoescaping-not-the-default

whit537:
Summarizing:

 - not always HTML
 - not always necessary even in HTML
 - performance

whit537:
Retort:

 - We're only using Jinja2 for HTML.
 - But it doesn't *hurt* to use it when it's not needed.
 - Hmmm ... really?

whit537:
It just seems increasingly stupid to rely on "a human knows" for the [544 places](https://gist.github.com/whit537/755c59aec14335b019b9) where are potentially vulnerable to injection. Performance will have to be *really* bad to justify not switching to autoescaping. No?

whit537:
http://jinja.pocoo.org/docs/dev/api/#autoescaping

whit537:
Autoescape interacts with i18n somehow:

![screen shot 2015-02-04 at 1 15 30 pm](https://cloud.githubusercontent.com/assets/134455/6046414/f7ceb702-ac6f-11e4-9eed-586c3e2eb1ca.png)

whit537:
[...]

PR #3153, opened 2015-02-04
Here's a helper for #722

whit537:
Figured this would be worth having for now until we can switch to autoescape (#3152).


whit537:
Dropping back to WIP while #722 shakes out. [IRC](https://botbot.me/freenode/gratipay/2015-02-05/?msg=31340128&page=2)

whit537:
Closing since we don't need a review step with the way that #3164 ended up.

PR #3157, opened 2015-02-05
save me some typing during payday

whit537:


whit537:
I put this in the root instead of in `bin` because that's where `payday.sh` is, and this is parallel to that.

whit537:
[IRC](https://botbot.me/freenode/gratipay/2015-02-05/?msg=31337514&page=1)

PR #3163, opened 2015-02-06
Drop referer check

### LINKED ISSUE #2929: login requires Referer
This just started recently.  When I try to log in I get a blank page.  Inspecting the response in my console shows I'm getting a 500 from the server.

My browser (ABrowser): Mozilla/5.0 (X11; Linux i686; rv:33.0) Gecko/20100101 Firefox/33.0

Changaco:
I was going to close #2929 as "won't fix", but then I realized that we probably don't need that Referer check, because we never send the CSRF token over insecure connections (this PR modifies a test to make sure of that).

Conflicts with #3022, I'll rebase after it's merged.

### DIFF
--- a/gratipay/security/csrf.py
+++ b/gratipay/security/csrf.py
@@ -11,7 +11,6 @@ See also:
 from datetime import timedelta
 import re
 import urlparse
-from aspen import log_dammit
 
 
 #from django.utils.cache import patch_vary_headers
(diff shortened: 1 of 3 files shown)


gratipay/security/csrf.py
```diff
@@ -99,32 +87,6 @@ def get_csrf_token_from_request(request):
...
-            # http://example.com/ as completely untrusted. Under HTTPS,
-            # Barth et al. found that the Referer header is missing for
-            # same-domain requests in only about 0.2% of cases or less, so
-            # we can use strict Referer checking.
-            referer = request.headers.get('Referer')
-            if referer is None:
-                raise Response(403, REASON_NO_REFERER)
-
-            good_referer = 'https://%s/' % _get_host(request)
-            if not same_origin(referer, good_referer):
```
whit537: Where is `same_origin` coming from? I'm not seeing that this PR removes a definition or import for it. Are we using it elsewhere?

Changaco: Good catch, I've added a commit that removes it.

Changaco:
Rebased on master.

PR #3172, opened 2015-02-13
teams listing

### LINKED ISSUE #1483: provide a way to browse/search teams
I want to be able to easily find teams to contribute to.

<bountysource-plugin>

---
Want to back this issue? **[Place a bounty on it!](https://www.bountysource.com/issues/950213-gittip-com-teams?utm_campaign=plugin&utm_content=tracker%2F85909&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F85909&utm_medium=issues&utm_source=github).
</bountysource-plugin>

whit537:
Closes #1483.

### DIFF
--- a/www/explore/teams/index.spt
+++ b/www/explore/teams/index.spt
@@ -34,9 +34,9 @@ subhead = _("Teams")
     {% for team in teams %}
     <tr>
         <td class=""><a href="/{{ team.name }}/">{{ team.name }}</a></td>
-        <td class="figure">{{ format_currency(team.receiving, 'USD', '#,##0.00') }}</td>
+        <td class="figure">{{ format_decimal(team.receiving, '#,##0.00') }}</td>
         <td class="figure">{{ team.nmembers }}</td>
-        <td class="figure">{{ format_currency(team.dollars_per_member, 'USD', '#,##0.00') }}</td>
+        <td class="figure">{{ format_decimal(team.dollars_per_member, '#,##0.00') }}</td>
     </tr>
     {% endfor %}
 </table>
--- a/www/explore/teams/index.spt
+++ b/www/explore/teams/index.spt
@@ -4,7 +4,12 @@ teams = website.db.all("""
 
     SELECT t.*
          , p.receiving
+         , p.goal
          , p.receiving / t.nmembers AS dollars_per_member 
+         , CASE p.goal
+           WHEN NULL THEN NULL
+           ELSE ((p.receiving / p.goal) * 100)::int
+            END                     AS pfunded
       FROM ( SELECT team            AS name
                   , count(member)   AS nmembers
                FROM current_takes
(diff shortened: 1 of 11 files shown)


Changaco:
Sooner or later we're going to need pagination for the teams listing.

www/explore/teams/index.spt
```diff
@@ -0,0 +1,43 @@
...
+         , p.receiving
+         , p.receiving / t.nmembers AS dollars_per_member 
+      FROM ( SELECT team            AS name
+                  , count(member)   AS nmembers
+               FROM current_takes
+           GROUP BY team
+            ) AS t
+      JOIN participants p 
+        ON p.username = t.name
+  ORDER BY dollars_per_member DESC, nmembers ASC
```
Changaco: Since the focus is on receiving we should filter out the teams that don't accept tips.

whit537: Hmm ... we allow such a thing? Okay.

Changaco:
It looks weird to have a completely empty sidebar.

www/explore/teams/index.spt
whit537: Done in 1c3bc7c77c749b1068828706a4bac3471dd62bc7. I also added `goal` and `% funded` columns:

screen shot 2015-02-13 at 4 00 27 pm

Changaco: You put the goals in the receiving column, and vice versa.

whit537:
> It should be `goal >= 0`.

Why? When goal is 0 they are a patron. The primary purpose of this page is to enable individuals to find teams that they could join on Gratipay to do work and take funding. If we're going to preclude `goal < 0` then we should preclude `goal = 0` as well.

whit537:
And we definitely should *include* `goal is null`.

whit537:
(But maybe you weren't suggesting we drop the `OR p.goal IS NULL`.)

whit537:
> It looks weird to have a completely empty sidebar.

Okay, well, there's precedent.

screen shot 2015-02-13 at 4 31 40 pm

whit537:
@Changaco I think we shouldn't block on the empty sidebar, since that's not unique to this page. Do you want to block on pagination?

whit537:
I'm not going to get pagination done today. I think we should abstract out a nice javascript table library like we have for charts, and use that for the teams listing as well as the team members listing and possibly in other places as well.

whit537:
Maybe elsewhere team listings, maybe communities, maybe giving and receiving on profiles. I think building that abstraction should be out of scope here.

Changaco:
>> It should be goal >= 0.
>
>Why?

Because that's how `accepts_tips` is defined.

whit537:
> Because that's how `accepts_tips` is defined.

But why should we use that condition here?

Changaco:
> But why should we use that condition here?

Because the idea was to filter out teams who can't receive, not teams who say they're here to give but can still receive.

whit537:
> Because the idea was to filter out teams who can't receive, not teams who say they're here to give but can still receive.

Done in e9e030305d2898cf25bb75ab4866861f68570668.

Issue #3127, opened 2015-01-20
downgrade communities to tags


whit537:
@wilkie on [Twitter](https://twitter.com/wilkieii/status/557594476563398656):

> I don't think @Gratipay can ever say they're for supporting worthwhile communities when the biggest receiver is 8chan 

![b7z5oifiaaaaju4](https://cloud.githubusercontent.com/assets/134455/5822525/e084fd8c-a0a2-11e4-830a-6ee4724b5d02.png)

I agree. For communities on Gratipay to be meaningful, they need to be curated. I've been thinking that we should partner with relevant organizations to curate each of the communities we have. So the [OSI](http://opensource.org/) would curate the "open source" community, jQuery Foundation -> jQuery, PSF -> Python, etc.

What should we do if we can't find a partner? Should we simply turn off communities that are large enough to matter but for which we can't find curators? Should we leave them as-is but clearly mark them as "uncurated"?

whit537:
[IRC](https://botbot.me/freenode/gratipay/2015-01-20/?msg=30040612&page=1)

whit537:
I had thought that "open source" might be a trademark owned by OSI, but that's not obvious to me from skimming [their trademark guidelines](http://opensource.org/trademark-guidelines).

whit537:
Nor from http://en.wikipedia.org/wiki/Open_source.

colindean:
I think there's room for three kinds of communities: sponsored communities, moderated communities, and open communities. Sponsored and moderated communities have some gatekeeper, while open communities have no gatekeeper and are meant for ideas, concepts, or technologies that have no centralized body that can authoritatively act as a gatekeeper. 

Sponsored communities can be pushed by a company or organization wanting to recognize volunteers, without having to maintain a team account and deal with the politics of distributing team take. Gratipay can, perhaps for full consideration, put a logo on the sponsored community page and Gratipay says that people can ask to join the community or only be invited.

I do like the idea of effecting that anyone can join open communities, and that they are unmoderated and Gratipay isn't responsible for the membership of those communities.

Maybe another kind of community is a private community. Only people within that community can see who else is a member thereof, and the community can be visible or invisible on Gratipay. Not sure how this would work in Gratipay's goal of keeping as much transparent/public as possible, though.

Changaco:
I suggest renaming communities to "tags" and limiting the number of tags per user.

rummik:
[...]

PR #3077, opened 2015-01-08
Search Opt Out

clone1018:
This PR should add the is_searchable functionality, allowing users to control their privacy settings for our internal search and public search (Google, Yahoo, etc).

I also moved the metadata in the profile head that should be shared across all pages for that user.

PR #3066, opened 2015-01-06
Search

Changaco:
This PR implements search over profile statements as well as user and community names. It doesn't modify the homepage because that will be done later (#2809).

PR #3180, opened 2015-02-16
opt out of search should also opt out of browse

whit537:
Just after we implemented search (#3066), we implemented the ability to opt out of search (#3077). This PR extends the `is_searchable` boolean to also govern inclusion in the new teams listing (#3172). When we convert communities to tags (#3127), we should extend this to govern inclusion in tag browsing as well.


whit537:
I considered tweaking the language on the privacy setting that controls this, but decided not to, because I want to fold Explore and Search together into one interface.

PR #3195, opened 2015-02-20
Properly erase cookies

Changaco:


gratipay/utils/__init__.py
```diff
@@ -149,13 +151,11 @@ def set_cookie(cookies, key, value, expires=None, httponly=True, path='/'):
     cookies[key] = value
     cookie = cookies[key]
     if expires:
-        if isinstance(expires, datetime):
-            pass
-        elif isinstance(expires, timedelta):
+        if isinstance(expires, timedelta):
             expires += utcnow()
-        else:
-            raise TypeError('`expires` should be a `datetime` or `timedelta`')
```
rohitpaulk: What if `expires` isn't a `datetime` or `delta`?

Changaco: Then it's expected to be `bytes` representing a date in the correct format, like `BEGINNING_OF_EPOCH`.

PR #3200, opened 2015-02-23
More idiomatic hooks

Changaco:
Split from #3189, requires more work. If I'm reading it right it breaks the CSRF check, yet tests are green, that's a problem.

### DIFF
--- a/gratipay/main.py
+++ b/gratipay/main.py
@@ -117,7 +117,8 @@ algorithm.functions = [
     i18n.set_up_i18n,
     authentication.start_user_as_anon,
     authentication.authenticate_user_if_possible,
-    csrf.get_csrf_token_from_request,
+    csrf.extract_token_from_cookie,
+    csrf.reject_forgeries,
 
     algorithm['dispatch_request_to_filesystem'],
 
@@ -133,7 +134,7 @@ algorithm.functions = [
 
     gratipay.set_misc_headers,
     authentication.add_auth_to_response,
-    csrf.add_csrf_token_to_response,
+    csrf.add_token_to_response,
     cache_static.add_caching_to_response if website.cache_static else noop,
     x_frame_options,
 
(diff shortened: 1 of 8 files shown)


whit537:
@Changaco Is there a reason for 24b8cf51e7776618a6f8e829322953316db29701 that I'm missing? If you're good with 124e7221f226b83462ce4a80c4c75d4dd753e0f6 then I'm good to merge this.

Changaco:
I've modified the tests to show why we can't use `return {}` for the CSRF hook.

Changaco:
We were entirely skipping the CSRF stuff for `/assets/` and `/callbacks/`, with your version we're generating and returning tokens for these requests.

whit537:
> We were entirely skipping the CSRF stuff for `/assets/` and `/callbacks/`, with your version we're generating and returning tokens for these requests.

Is that a big deal? Why?

Changaco:
> Is that a big deal? Why?

It's bad for performance, but more importantly it could break asset caching because we're adding `Cookie` to the `Vary` header.

Changaco:
Actually we don't, because `add_caching_to_response` overrides `Vary`. That's a bug.

whit537:
> It's bad for performance

Its effect on performance is negligible, because we have a CDN in front of `/assets/`, and `/callbacks/` is low-traffic.

> but more importantly it could break asset caching because we're adding `Cookie` to the `Vary` header.

Fair enough. Fixed in e26b91a865c5bd0ab2ee7bcab56c2878e7ab0583.

whit537:
> Actually we don't, because `add_caching_to_response` overrides `Vary`. That's a bug.

Is that a regression on this PR or is it a separate issue?

Changaco:
> Is that a regression on this PR or is it a separate issue?

Separate issue, this PR doesn't modify `cache_static.py`.

whit537:
Okay, I kicked Travis. We good here?

Changaco:
> We good here?

I think so, yes.

Issue #3162, opened 2015-02-05
move SQL functions out of schema.sql


whit537:
Following on from #2360, and in view of the hack at https://github.com/gratipay/gratipay.com/pull/3154#issuecomment-73041912, what if we moved SQL functions to a separate file from schema.sql? If we had one file per function we could automate the process of updating those functions during deployment, and we'd get sensible diffs on PRs because we wouldn't have to use branch.sql as a go-between (branch.sql would remain for table changes).

techtonik:
Decouple stored procedures from data structure seems like a logical way to do.

PR #3206, opened 2015-02-26
Separate SQL functions

Changaco:
Implements #3162.


whit537:
Maybe move `branch.sql` into `sql/` as well?

sql/upsert_community.sql
```diff
@@ -0,0 +1,46 @@
+CREATE OR REPLACE FUNCTION upsert_community() RETURNS trigger AS $$
```
whit537: Why do we use `CREATE OR REPLACE` here but just `CREATE` for `enumerate`, above? It seems that we should have one pattern. I think I would now expect `CREATE OR REPLACE` for functions.

Changaco:
Anything else ?

whit537:
Yeah, I don't understand how this interacts with whatever we're doing now to keep dev and prod in sync.

whit537:
I also don't understand how this affects deployment. If I want to modify `upsert_community`, do I still use branch.sql? If not, how do my changes get deployed?

Changaco:
> I don't understand how this interacts with whatever we're doing now to keep dev and prod in sync.

`make schema-diff` doesn't care about what file(s) the schema is in. I've run it from this branch, it doesn't detect any drift.

> If I want to modify `upsert_community`, do I still use branch.sql?

Yes, you make a `branch.sql` with an `\i sql/upsert_community.sql` inside.

whit537:
> I've run it from this branch and it doesn't detect any drift.

I get the following error from `make schema-diff`:

```shell
[...]
pg_dump -sO `heroku config:get DATABASE_URL -a gratipay` >prod.sql
env/bin/honcho run -e defaults.env,tests/test.env,tests/local.env sh -c 'pg_dump -sO "$DATABASE_URL"' >local.sql
diff -uw prod.sql local.sql
--- prod.sql    2015-03-02 09:08:05.000000000 -0500
+++ local.sql   2015-03-02 09:08:06.000000000 -0500
@@ -24,20 +24,6 @@
 
 
 --
--- Name: pg_stat_statements; Type: EXTENSION; Schema: -; Owner: -
---
-
-CREATE EXTENSION IF NOT EXISTS pg_stat_statements WITH SCHEMA public;
-
-
---
--- Name: EXTENSION pg_stat_statements; Type: COMMENT; Schema: -; Owner: -
---
-
-COMMENT ON EXTENSION pg_stat_statements IS 'track execution statistics of all SQL statements executed';
-
-
---
 -- Name: pg_trgm; Type: EXTENSION; Schema: -; Owner: -
 --
 
make: *** [schema-diff] Error 1
$
```

Changaco:
[`schema.sql` does execute these two commands](https://github.com/gratipay/gratipay.com/blob/separate-sql-functions/sql/schema.sql#L12-13). Maybe they're failing on your system for some reason, do you see an error message ?

whit537:
Indeed:

```
ERROR:  could not open extension control file "/usr/local/pgsql/share/extension/pg_stat_statements.control": No such file or directory
ERROR:  extension "pg_stat_statements" does not exist
```

whit537:
I built and installed the extension and now `make schema-diff` exits without error.

Issue #3210, opened 2015-02-26
stats is json by default now


whit537:
Might be a regression due to #3189?

https://gratipay.com/about/stats

whit537:
I believe if we change the order of the content pages so that HTML is first then that will fix it. But why did this start happening? Is it indeed a #3189 regression, or has Chrome changed it's `Accept` header? What other simplates are affected?

untitaker:
Search is broken too.

whit537:
```
$ find www -name \*.spt|grep -v \\..*\\.
www/%username/account/close.spt
www/%username/elsewhere/refresh.spt
www/%username/history/record-an-exchange.spt
www/about/charts.spt
www/about/contact.spt
www/about/details.spt
www/about/dnt.spt
www/about/error.spt
www/about/faq.spt
www/about/features/dnt.spt
www/about/features/index.spt
www/about/features/teams/index.spt
www/about/fraud/2012-11-05.spt
www/about/index.spt
www/about/me/%redirect_to.spt
www/about/me/index.spt
www/about/me.spt
www/about/policies/index.spt
www/about/policies/privacy.spt
www/about/policies/terms-and-conditions.spt
www/about/policies/terms-of-service.spt
www/about/pricing.spt
www/about/privacy/index.spt
www/about/security/hall-of-fame.spt
www/about/security/index.spt
www/about/stats.spt
www/about/teams/index.spt
www/about/terms/index.spt
www/about/terms/of-service.spt
www/callbacks/balanced.spt
www/dashboard/index.spt
www/explore/teams/index.spt
www/on/%platform/associate.spt
www/on/%platform/redirect.spt
www/search.spt
www/security.spt
```

whit537:
```
$ find www -name \*.spt|grep -v \\..*\\.|xargs grep json_dump
www/about/stats.spt:[----------------------------------------------------------] application/json via json_dump
www/search.spt:[---] application/json via json_dump
$
```

whit537:
Fixed in #3213. Working now, @untitaker?

untitaker:
Yeah!

whit537:
@untitaker Huzzah! Thanks for the report. :-)

PR #3213, opened 2015-02-26
Work around content negociation bug

Changaco:
Temporary fix for #3210.

### DIFF
--- a/tests/py/test_privacy_json.py
+++ b/tests/py/test_privacy_json.py
@@ -43,7 +43,7 @@ class Tests(Harness):
 
     def test_participant_doesnt_show_up_on_search(self):
         self.hit_privacy('POST', data={'toggle': 'is_searchable'})
-        assert 'alice' not in self.client.GET("/search?q=alice").body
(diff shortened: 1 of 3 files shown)


whit537:
I believe search is the only other affected page:

```
$ find www -name \*.spt|grep -v \\..*\\.|xargs grep json_dump
www/about/stats.spt:[----------------------------------------------------------] application/json via json_dump
www/search.spt:[---] application/json via json_dump
$
```

whit537:
Ready when you (and Travis) are, @Changaco.

Changaco:
Added a commit to fix the failing test.

whit537:
I don't understand why that test passed before.

Changaco:
> I don't understand why that test passed before.

Because the test client doesn't send an `Accept` header, so it always gets the first page of a negotiated simplate when the requested path doesn't contain an extension.

whit537:
Good call. Reticketed upstream as https://github.com/gratipay/aspen-python/issues/422.

PR #3215, opened 2015-02-26
Fix content negociation

Changaco:
gratipay/aspen-python#422


whit537:
Good find, but why did the test fail to fail?

Changaco:
I'm guessing because I pushed the second commit too fast and Travis used it for both builds.

PR #3226, opened 2015-03-04
Get rid of `website.connect_tokens`

### LINKED ISSUE #715: move caching out of memory
There's one or two places in the app where we have an in-memory cache that will prevent us from scaling. I believe the two places are:

 - /on/twitter/ - we store state in memory during the round-trip to Twitter for oauth
 - /on/confirm.html - something about "connect tokens"

<bountysource-plugin>

---
Want to back this issue? **[Place a bounty on it!](https://www.bountysource.com/issues/316435-move-caching-out-of-memory?utm_campaign=plugin&utm_content=tracker%2F85909&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F85909&utm_medium=issues&utm_source=github).
</bountysource-plugin>

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/316435-move-caching-out-of-memory?utm_campaign=plugin&utm_content=tracker%2F85909&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F85909&utm_medium=issues&utm_source=github).
</bountysource-plugin>

Changaco:
Closes #715.

### DIFF
--- a/www/on/%platform/associate.spt
+++ b/www/on/%platform/associate.spt
@@ -64,13 +64,9 @@ elif action == 'connect':
     try:
         user.participant.take_over((platform.name, account.user_id))
     except NeedConfirmation, obstacles:
-        # XXX Eep! Internal redirect! Really?!
-        # XXX In this case, we can get away with reusing the same dispatch_result. When upstreaming
-        #     we should probably generalize by making a new DispatchResult.
-        request.internally_redirected_from = dispatch_result.match
-        resource = resources.get(website, website.www_root+'/on/confirm.html.spt')
-        state.update(locals())
-        raise resource.respond(state)
+        token, expires = account.make_connect_token()
+        response.set_cookie(b'connect_%s' % account.id, token, expires)
+        response.redirect('/on/confirm.html?id=%s' % account.id)
 
 elif action in {'lock', 'unlock'}:
     account.participant.update_goal(-1 if action == 'lock' else None)
--- a/www/on/confirm.html.spt
+++ b/www/on/confirm.html.spt
@@ -1,33 +1,19 @@
-"""This is called as an internal redirect from /on/blah/associate
-"""
-import uuid
-
 from aspen import Response
-from gratipay.models.participant import NeedConfirmation, Participant
 
-website.connect_tokens = {}
+from gratipay.models.account_elsewhere import AccountElsewhere
 
 [--------]
 
 if user.ANON:
-    raise Response(404)
+    raise Response(403)
 
-if 'fake' in request.qs:
-    account = next(iter(user.participant.get_accounts_elsewhere().values()))
-elif not getattr(request, 'internally_redirected_from', None):
-    raise Response(404)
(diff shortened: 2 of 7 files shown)


whit537:
Tests pass locally and UI appears to work. Reading the code ...

www/on/%platform/associate.spt
```diff
@@ -64,13 +64,9 @@ elif action == 'connect':
     try:
         user.participant.take_over((platform.name, account.user_id))
     except NeedConfirmation, obstacles:
-        # XXX Eep! Internal redirect! Really?!
-        # XXX In this case, we can get away with reusing the same dispatch_result. When upstreaming
-        #     we should probably generalize by making a new DispatchResult.
-        request.internally_redirected_from = dispatch_result.match
-        resource = resources.get(website, website.www_root+'/on/confirm.html.spt')
-        state.update(locals())
-        raise resource.respond(state)
```
whit537: @Changaco You're removing the internal redirect on principle, because it's a "mess"?

Changaco: I removed it because it's messy and because we don't need it.

whit537:
When I add an account that's linked to a different Gratipay account, I end up on `/on/confirm.html` with no querystring. From reading the code, I expect to end up on `/on/confirm.html?id=2345678`. I put a `set_trace` right before the redirect and that code is being run. If I hit back in Chrome I get back to an URL with the querystring. What's going on? Is this some weird Chrome behavior?

whit537:
Oh! Okay ... same behavior in Firefox. What I'm seeing is the Facebook landing cleanup.

whit537:
https://github.com/gratipay/gratipay.com/blob/1804/templates/base.html#L108-L114

whit537:
Probably the `pushState` there should include the querystring (and maybe be `replaceState` instead).

www/on/confirm.html.spt
```diff
@@ -1,33 +1,19 @@
...
-connect_token = uuid.uuid4().hex
-connect_key = (user.participant.username, account.platform, account.user_id)
-website.connect_tokens[connect_key] = connect_token
-
+    raise Response(403)
 
-other = Participant.from_username(account.participant.username)
+account_id = request.qs['id']
+account = AccountElsewhere.from_id(account_id)
+token = request.headers.cookie.get('connect_%s' % account_id)
```
whit537: What are the security implications of using a cookie here instead of the internal redirect? If an attacker stole this cookie what could they do with it?

Changaco: If you can steal cookies you can steal the session cookie, so I don't think it matters much that you can also have a connect token.

whit537:
Querystring clobbering reticketed as #3252.

PR #3230, opened 2015-03-05
Clean up HTTP headers

Changaco:
Fixes https://github.com/gratipay/gratipay.com/pull/3200#issuecomment-77206397 and other related issues.

### DIFF
--- a/gratipay/main.py
+++ b/gratipay/main.py
@@ -8,7 +8,7 @@ from gratipay import canonize, utils
 from gratipay.cron import Cron
 from gratipay.models.participant import Participant
 from gratipay.security import authentication, csrf, x_frame_options
-from gratipay.utils import cache_static, erase_cookie, i18n, set_cookie, timer
+from gratipay.utils import erase_cookie, http_caching, i18n, set_cookie, timer
 from gratipay.version import get_version
 from gratipay.renderers import jinja2_htmlescaped
 
@@ -98,8 +98,8 @@ algorithm.functions = [
 
     algorithm['dispatch_request_to_filesystem'],
 
-    cache_static.get_etag_for_file if website.cache_static else noop,
-    cache_static.try_to_serve_304 if website.cache_static else noop,
+    http_caching.get_etag_for_file if website.cache_static else noop,
+    http_caching.try_to_serve_304 if website.cache_static else noop,
 
     algorithm['apply_typecasters_to_path'],
     algorithm['get_resource_for_request'],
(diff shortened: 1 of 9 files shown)


whit537:
It looks like this removes the `Vary` header entirely. Why?

Changaco:
> It looks like this removes the `Vary` header entirely. Why?

Because we don't need it. Our static assets don't vary on anything, and our dynamic assets potentially vary on everything (we use `Cache-Control: no-cache` for them, it's better than `Vary: *`).

gratipay/main.py
```diff
@@ -109,10 +109,10 @@
     tell_sentry,
     algorithm['get_response_for_exception'],
 
-    gratipay.set_misc_headers,
+    gratipay.set_version_header,
     authentication.add_auth_to_response,
     csrf.add_token_to_response,
-    cache_static.add_caching_to_response if website.cache_static else noop,
```
whit537: We don't need the noop path anymore because if `not etag` then `add_caching_to_response` is a noop anyway?

gratipay/security/authentication.py
```diff
@@ -82,9 +81,7 @@ def add_auth_to_response(response, request=None, user=ANON):
     if request is None:
         return  # early parsing must've failed
     if request.line.uri.startswith('/assets/'):
-        return  # assets never get auth headers and have their own caching done elsewhere
-
-    response.headers['Expires'] = BEGINNING_OF_EPOCH # don't cache
```
whit537: We're using `Cache-Control: no-cache` for all dynamic resources, so we don't need to use `Expires` anymore?

whit537: > we don't need to use `Expires` anymore

... for authed resources, since authed resources are a strict subset of dynamic resources?

gratipay/main.py
Changaco: The reason we're removing the `noop` path here is that we always want to add the proper caching headers to the response, whether or not `cache_static` is turned on. `add_caching_to_response` isn't a noop when we don't have an `etag`, it adds `Cache-Control: no-cache`.

gratipay/security/authentication.py
Changaco: Yes, we don't need `Expires` here, and we don't want it here either: a function that handles auth shouldn't set caching headers, it's not its job.

gratipay/main.py
whit537: Ah, right: because now `add_caching_to_response` is scoped to all content, not just static content. Yes?

Changaco: Correct.

PR #3232, opened 2015-03-05
Fix a reference error in `payments.js`

Changaco:
Looks like I broke the credit card and bank account pages in #3218 (5667f656cf4047dc9e70a53e12ca82a5a751a413).


Changaco:
So, our JS tests regularly fail for no reason, but they don't detect simple "variable not in scope" bugs…

PR #3240, opened 2015-03-10
Improve our docker experience

rohitpaulk:
- Our [docker image on Docker Hub](https://registry.hub.docker.com/u/citruspi/gittip/) is outdated. Unless we setup [automated builds](http://docs.docker.com/docker-hub/builds/) or find someone who is willing to update the image regularly, a pre-built image doesn't make sense. So, I've removed the documentation for pulling from Docker Hub.

- Our `Dockerfile` now copies the repository files from context, not by cloning our GitHub repository. Before this, we were asking the user to clone our GitHub repository first, and then clone it again during the build process.

- It's now possible to execute commands such as `make schema` and `make data` within the container, documentation has been added too. Also added documentation for mounting a local folder, that's the only way people can actually hack on Gratipay using Docker.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,16 +1,16 @@
 # Dockerfile to build and run Gratipay
-# Version 0.1 (April 15, 2014)
+# Version 0.2 (March 10, 2015)
 
 ################################################## General Information ##################################################
 
-FROM ubuntu:12.04
+FROM ubuntu:14.04
 MAINTAINER Mihir Singh (@citruspi)
 
 ENV DEBIAN_FRONTEND noninteractive
 
 ################################################## Install Dependencies #################################################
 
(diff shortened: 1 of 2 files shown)


rohitpaulk:
cc: @citruspi

citruspi:
> Our Dockerfile now copies the repository files from context, not by cloning our GitHub repository. Before this, we were asking the user to clone our GitHub repository first, and then clone it again during the build process.

That definitely makes sense.

> Unless we setup automated builds or find someone who is willing to update the image regularly, a pre-built image doesn't make sense. 

I would be totally interested in setting up automated builds with Docker Hub.

As far as the Dockerfile itself goes, I haven't run it myself, but the changes look solid.

&mdash; @citruspi

PR #3247, opened 2015-03-14
Fixes for the credit card and bank account pages

Changaco:


rohitpaulk:
Added a commit.

Changaco:
It wasn't really necessary since the CC `<input>` won't let you enter anything that isn't a number.

rohitpaulk:
@Changaco - Better safe than sorry! :)

Issue #3236, opened 2015-03-06
clean up Profile/Account IA


whit537:
Over on #3220 it has surfaced that our IA around profiles and accounts is confused. Currently we treat Account as a private subset of Profile, when it would be more user-centric to treat Profile as a public subset of Account (cf. https://github.com/gratipay/gratipay.com/issues/3220#issuecomment-77561304). Any given user experiences Gratipay as a whole encompassing both public and private parts, so we should organize our IA with Account as the whole and Profile as a part thereof.

Current:

```
Profile
  Profile
  Members
  Receiving
  Giving
  History
  Widgets
  Account
  Events
```

Initially Proposed:

```
Account
  Dashboard
  Profile
    Basics
    Giving
    Receiving
    Members
  History
  Widgets
  Settings
  Events
```

Evolving Target:

```
Dashboard /
  Profile       /$username
  Members               /members
  Giving                /giving
  Receiving             /receiving
  Account → Settings
    Settings            /settings
    History             /history
    Widgets             /widgets
    Events              /events
```

Steps:
- [x] Rename Account to Settings (#3254)
- [x] ~~Move Dashboard, History, Widgets, Settings, and Events up to second-level nav, which is currently unused under Account~~
- [ ] move Settings, History, Widgets, and Events under Account (#3256)
- [ ] implement a proper dashboard at `/` (disentangled from Profile)
##

whit537:
I believe this would also address the Dashboard issue lingering from ... #2938, was it?

ESWAT:
By "second-level nav" do you mean in the sidebar, or should a new `.nav` be created somewhere near the existing one to act as a parent?

whit537:
@ESWAT I'm referring to the sidebar nav, yes. I think we'll want to revisit the presentation of that nav downstream of #3063, but that's out of scope here.

Changaco:
The profile pages use the sidebar so if you put the Account nav there then when you click on Profile the nav disappears. That's bad UI design IMO. I think modifying the top nav is a better idea, using dropdowns where necessary.

whit537:
Account renamed to Settings in #3254.

whit537:
Here are some other IA for comparison:

### GitHub
```
News Feed /
  Settings      /settings
    Account settings    /admin
    Emails              /emails
    Notification center /notifications
    Billing             /billing
    SSH keys            /ssh
    Security            /security
    Applications        /applications
    Repositories        /repositories
    Organizations       /organizations
[...]

PR #3256, opened 2015-03-17
aaaaaand account

whit537:
Brings back Account as a container for Settings, History, Widgets, and Events.

screen shot 2015-03-16 at 10 25 20 pm


whit537:
Builds on #3254; part of #3236.

Changaco:
I'm not comfortable merging this as it is, I think it makes our nav even more confusing.

whit537:
@Changaco Okay, I'll revisit after #3258.

PR #3259, opened 2015-03-17
group about templates

whit537:
Quickie. Includes #3256.


whit537:
Rebased on master since we closed #3256.

PR #3262, opened 2015-03-18
Use container-based testing for Travis

techtonik:
http://docs.travis-ci.com/user/workers/container-based-infrastructure/

PR #3263, opened 2015-03-18
Add caching for pip

techtonik:
Container-based builds allow to cache files. There is support
for caching pip files with `cache: pip`, but because of
https://github.com/travis-ci/travis-ci/issues/3239 it gets disabled
if `.travis.yml` provides own install section. Therefore adding
cached dir manually.


techtonik:
I don't see Travis kicking in to test that. Seems like #3262 needs to be merged first.

Changaco:
I don't see how that would help. Almost all of the dependencies are in the `vendor/` dir, pip doesn't download them. It's compiling the C(++) libs that slows down our builds, we could try caching `env/lib/python2.7/site-packages`.

Changaco:
Needs to be redone on top of master.

techtonik:
How can I restore the branch to rebase on top?

PR #3265, opened 2015-03-18
execnet is a missing pytest_cache dep from requirements_tests.txt

techtonik:
See https://travis-ci.org/gratipay/gratipay.com/builds/54857342#L507


techtonik:
Cool. Thanks.

PR #3264, opened 2015-03-18
Add Travis caching

techtonik:
Container-based builds allow to cache files
http://docs.travis-ci.com/user/caching/

For `pip` packages there is `cache: pip`, but it doesn't work,
because we have a custom `install:` section. See
travis-ci/travis-ci#3239 for details.

PR #3274, opened 2015-03-20
Cache python dependencies across Travis builds

Changaco:
Alternative to #3264.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,6 +11,7 @@ cache:
     - env/bin
     - env/lib/python2.7/site-packages
 install:
+ - if [ "${TRAVIS_BRANCH}" = "master" -a "${TRAVIS_PULL_REQUEST}" = "false" ]; then rm -rf env; fi
  - touch requirements.txt package.json
  - make env
  - make node_modules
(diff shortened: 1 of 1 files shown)


techtonik:
This will cache gratipay application files also.

Changaco:
Ready for review. Build time is around 2 min 30 sec, down from around 3 min 20 sec without caching.

techtonik:
Do you guarantee that this won't cache any files that is installed by Gratipay?

Changaco:
Caching parts of `env/` between builds is similar to what happens on my machine and those of the other developers where `env/` is almost never deleted. If it was problematic we would have noticed it a long time ago.

techtonik:
I can't approve that, because I haven't finished WIP over #3264 yet

whit537:
I'm fine with this on branches. On master we should make a clean build, just as on master we hit live API endpoints instead of using the VCR cache.

whit537:
As I read this it would apply to master as well as other branches, correct? Can we scope this to branches other than master?

Changaco:
> Can we scope this to branches other than master?

Done.

Issue #3025, opened 2014-12-18
fix travis


chadwhitacre:
Travis is failing on #3011 and #3020 with this error:

```
$ make test
env/bin/honcho -e defaults.env,tests/test.env,tests/local.env run ./recreate-schema.sh
==============================================================================
Recreating public schema ... 
psql: FATAL:  database "travis" does not exist
make: *** [test-schema] Error 2
The command "make test" exited with 2.
Done. Your build exited with 1.
```

What's up? [IRC](https://botbot.me/freenode/gratipay/2014-12-17/?msg=27829353&page=2)

PR #3275, opened 2015-03-20
Use released honcho version

techtonik:
The fix needed for 1813762f should be already there.


techtonik:
@Changaco, can you clarify why the build fails? The failure seems related to https://github.com/gratipay/gratipay.com/issues/3025

Changaco:
We can't use a honcho version from upstream, they don't support `honcho run PROCESS` like https://github.com/gratipay/honcho does.

techtonik:
Bad day.

https://honcho.readthedocs.org/en/latest/using_procfiles.html#no-honcho-run-target

I don't get it. What's the difference between running a command (which is implemented) and running a process (which is Gratipay fork)?

techtonik:
/me is reading http://blog.daviddollar.org/2011/05/06/introducing-foreman.html

techtonik:
https://stackoverflow.com/questions/21129788/what-is-the-difference-between-foreman-start-and-foreman-run-rails-server

Ok. `process` is `command configured in Procfile`.  `foreman start` is used to exec process. `foreman run` is used to run shell command in the same conditions. Why extend `foreman run` to run processes?

@whit537 ?

PR #3279, opened 2015-03-23
update security email and PGP key

### LINKED ISSUE #3174: update PGP key
Mine expired a couple days ago:

http://pgp.mit.edu/pks/lookup?op=vindex&search=0x6EE18A8DC47977C2

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/8563630-update-pgp-key?utm_campaign=plugin&utm_content=tracker%2F85909&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F85909&utm_medium=issues&utm_source=github).
</bountysource-plugin>

whit537:
Closes #3174.

You can see here that I'm proposing that we never use PGP to encrypt messages or authenticate our identity, but only to decrypt messages we might receive. If we need to communicate further, we use a private GitHub repo for that.

PGP is (well-known to be) a usability nightmare, and I'd like to use it as little as possible. By only using it to decrypt initial reports, we can avoid worrying overmuch about our expiry and revocation policies.


whit537:
You can see that I also decided not to publish our key on a keyserver. It's enough to have it on our server. Better, actually, since the keyserver we were using is HTTP-only.

Issue #1928, opened 2014-01-19
Prevent abuse of the leaderboards, possibly remove them entirely


seanlinsley:
@galuszkak and I were able to do this, ostensibly because we have credit cards attached.

<img width=30%" src="https://f.cloud.github.com/assets/688886/1948712/f9b3e67c-80a7-11e3-97b3-7291f97daa1c.png"/>

This ticket is here to theorize about the possible ways users could abuse the statistics we provide, both on leaderboards and on profile pages.

[IRC](https://botbot.me/freenode/gittip/msg/9944897/)

whit537:
After landing #3281 we could address this ticket by tracking giving amounts from the previous payday separately from the upcoming payday, and including both columns in the new explore table from over there.

whit537:
Leaderboards are gone.

Issue #3222, opened 2015-03-04
unify tables/listings


whit537:
We achieved consistent charts in #3067. Let's do the same for our tables/listings:
- explore
- search
- members
- receiving
- giving
- history
- events
- elsewhere teams
##

Changaco:
Is charts.js really an example ? [I protested](https://github.com/gratipay/gratipay.com/pull/3067#issuecomment-69002496) when you threw out my d3 code and what I said still stands.

What exactly would tables.js library do ? What would it do that a generic jQuery table plugin doesn't ?

(Also, there are no tables on the search page.)

whit537:
> (Also, there are no tables on the search page.)

Not yet. In #3221, I'm proposing that we unify the Explore and Search interfaces.

whit537:
> Is charts.js really an example ? I protested when you threw out my d3 code and what I said still stands.

Yes. Here's your protest:

> The new tip distribution charts look terrible, and I'm not convinced that writing our own code instead of using d3 is a good idea.

There are two concerns here:

 1. Design of the user experience.
 1. Choice of tools.

As much as possible, (1) needs to drive (2), and not the other way around.

When we *do* get to tooling decisions, we find a trade-off between control and convenience. We err on the side of control. Yes, we could use Django or Rails with their ecosystem of convenient plugins and ORMs. Instead, we use Aspen and PostgreSQL, because that gives us more control. Yes, that means we're responsible for the bugs in Aspen and Postgres.py, but it's worth it to us for the increased control. Some day we'll port to Haskthon for the same reason. ;-)

Re: D3 in particular, #1278 was actually the second instance of the tail wagging the dog (and afaict the D3 implementation of the tip distribution chart was primarily @cakey's, not yours). The first was #937. I had suggested in #792 that we switch to D3 without changing the UX, because I also wondered whether "writing our own code instead of using d3 is a good idea." We tried to switch to D3 in #937, and we failed to achieve the user experience we wanted, so we abandoned the attempt. D3 did not give us the control we want.

> What exactly would tables.js library do ? What would it do that a generic jQuery table plugin doesn't ?

It would do exactly what we want it to do. It would give us more control than a generic jQuery table plugin.

When you said that the new tip distribution charts look terrible, @clone1018 asked you to say more, and you didn't answer:

> [H]ow does the chart look terrible? Is it the data or the actual design of the chart?

*That's* the primary conversation we should be having, about both charts and tables: what user experience do we want to deliver?

[...]

PR #3185, opened 2015-02-17
find friends

whit537:
Here's a proposal for how to implement #3183 (really, I'm looking at #536).

 - [ ] cache friend calls
 - [x] add pagination to table
 - [ ] add search filter to table
 - [ ] add sorting to table
 - [x] abstract for other elsewheres
 - [x] implement or stub other elsewheres
 - [x] figure out how to handle Facebook limitations
 - [x] add elsewhere logo as sidebar avatar
 - [x] link to attached networks from dashboard

PR #3281, opened 2015-03-23
replace communities leaderboards with a table

### LINKED ISSUE #3127: downgrade communities to tags

whit537:
Partially addresses #3222, and sets us up for #3127.

### TODO

 - [ ] paging (cf. #3185)
 - [ ] clean up formatting a bit

### DIFF
--- a/gratipay/models/community.py
+++ b/gratipay/models/community.py
@@ -53,16 +53,27 @@ class Community(Model):
 
     def get_members(self, limit=None, offset=None):
         return self.db.all("""
-            SELECT p.*::participants
-              FROM current_community_members c
-              JOIN participants p ON p.id = c.participant
-             WHERE c.slug = %s
-               AND c.is_member
-               AND p.is_suspicious IS NOT true
-          ORDER BY c.ctime
-             LIMIT %s
-            OFFSET %s;
-        """, (self.slug, limit, offset))
+        SELECT row_number() OVER () AS n
+             , *
+          FROM ( SELECT p.username
+                      , (CASE p.anonymous_giving WHEN true THEN p.giving ELSE NULL END) AS giving
+                      , (CASE p.anonymous_receiving WHEN true THEN p.receiving ELSE NULL END) AS receiving
+                      , p.goal
+                      , p.number
+                      , c.ctime AS first_tagged
+                   FROM current_community_members c
+                   JOIN participants p ON p.id = c.participant
+                  WHERE c.slug = %s
+                    AND c.is_member
+                    AND p.is_suspicious IS NOT true
+                    AND p.is_searchable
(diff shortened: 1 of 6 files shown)


whit537:
screen shot 2015-03-23 at 6 29 11 pm

Changaco:
This is about #3222, not #3127. I assume you're planning to make the table sortable, so it doesn't really fix #1928 either.

whit537:
> This is about #3222, not #3127. 

The UI will need to change at least a little bit as part of #3127. For one, the sidebar doesn't make sense in the context of tags: one doesn't generally "organize" or "join" a tag; one isn't a "member" of a tag; the avatar no longer fits. We could simply drop the sidebar as on other pages (Search, Explore > Teams), but we also need to get the word "Member" out of "New Members."

However, I think #3127 actually requires a strong visual break from the current communities page. The point of #3127 isn't just to change `communities` to `tags` in our codebase, the point is to change the user's expectation from one of strongly identifying with others in the same *community* to weakly identifying with others using the same *tag*.

Since I've put #3222 out there as a general goal, I do see a chance to move us forward on both #3222 and #3127 at the same time by adopting a table layout for tags. Implementing #3222 for tags gives us a significant break from the current communities implementation.

It could make sense to do them as two separate PRs, but once the #3222 portion of this PR is done, the remaining work will largely be under the hood, and I think it won't bloat this PR too much to do it here.

> it doesn't really fix #1928 either.

Addressed on #1928.

whit537:
I've narrowed the scope of this to be about #3222.

whit537:
Turned off communities in #3706.

Issue #3268, opened 2015-03-18
upgrade virtualenv, setuptools, and pip


whit537:
```
"/Users/whit537/bin/python2.7" "./vendor/virtualenv-1.11.6.py" \
				--unzip-setuptools \
				--prompt="[gratipay] " \
				--extra-search-dir=./vendor/ \
				--always-copy \
				./env/
New python executable in ./env/bin/python2.7
Also creating executable in ./env/bin/python
Installing setuptools, pip...
  Complete output from command /Users/whit537/perso...om/env/bin/python2.7 -c "import sys, pip; sys...d\"] + sys.argv[1:]))" setuptools pip:
  Ignoring indexes: https://pypi.python.org/simple/
Downloading/unpacking setuptools
  Running setup.py (path:/Users/whit537/personal/gratipay/gratipay.com/env/build/setuptools/setup.py) egg_info for package setuptools

    warning: no files found matching '*.c' under directory 'tests'
    warning: no files found matching '*.pyx' under directory 'tests'
Downloading/unpacking pip
  Running setup.py (path:/Users/whit537/personal/gratipay/gratipay.com/env/build/pip/setup.py) egg_info for package pip
    /Users/whit537/lib/python2.7/distutils/dist.py:267: UserWarning: Unknown distribution option: 'entry_points'
      warnings.warn(msg)
    /Users/whit537/lib/python2.7/distutils/dist.py:267: UserWarning: Unknown distribution option: 'extras_require'
      warnings.warn(msg)
    /Users/whit537/lib/python2.7/distutils/dist.py:267: UserWarning: Unknown distribution option: 'zip_safe'
      warnings.warn(msg)
    /Users/whit537/lib/python2.7/distutils/dist.py:267: UserWarning: Unknown distribution option: 'tests_require'
      warnings.warn(msg)
    usage: -c [global_opts] cmd1 [cmd1_opts] [cmd2 [cmd2_opts] ...]
       or: -c --help [cmd1 cmd2 ...]
       or: -c --help-commands
       or: -c cmd --help

    error: invalid command 'egg_info'
    Complete output from command python setup.py egg_info:
    /Users/whit537/lib/python2.7/distutils/dist.py:267: UserWarning: Unknown distribution option: 'entry_points'

  warnings.warn(msg)

/Users/whit537/lib/python2.7/distutils/dist.py:267: UserWarning: Unknown distribution option: 'extras_require'

  warnings.warn(msg)

[...]

PR #3283, opened 2015-03-24
Upgrade virtualenv 1.11.6 -> 12.0.7 (#3268)

techtonik:
Apart from issue #3268 this is also needed to experiment with caching wheels in Travis.
pip 1.5.6 still unpacks our vendorized distribution if they are already installed.


techtonik:
Thanks!

Issue #1459, opened 2013-09-15
get tarballs out of repo


whit537:
We have tarballs for all Python dependencies in our repo under vendor. The reason we do this is to avoid network access when building the software for the second and subsequent times, and to make it easier to collaborate when working in potentially network-constrained environments such as sprints. However, these tarballs take up space in our repo, especially as we upgrade dependencies over time. The one time we force-pushed to master was to remove an overly large dependency (#919).

I've discovered [a way](http://stackoverflow.com/questions/13698374/how-to-create-a-standalone-pip-package-all-deps-included) to use pip to download the tarballs as a separate step from installing them, and I propose we implement that in our build process. We can do this in the Makefile or wait until we land a pure-Python build system (#1134).

##

Changaco:
Now that we have caching for Travis builds this could be done without risk of slowing them down. @techtonik you want to take this on?

techtonik:
I kind of converted myself into thinking that storing dependencies in version control repository is a Good Thing :tm:. Because you need not to fear that something is broken in the middle, the version is gone, or replaced or not the version that was actually meant to be used.

Is it possible to do download as a part of checkout process without extra steps that are not really obvious? So that it is still possible to work offline. I am not familiar with git well enough to know how to do this. In Mercurial it is http://mercurial.selenic.com/wiki/LargefilesExtension but it needs to be enabled in extensions IIRC.

techtonik:
The problem may be less actual if we switch to .zip and wheels (which are also .zip). Because they are not [solid](https://en.wikipedia.org/wiki/Solid_compression) and each file has its own compressed block, it is not a big surprise to see this:

```
 rename vendor/{setuptools-3.6-py2.py3-none-any.whl => setuptools-14.3.1-py2.py3-none-any.whl} (55%)
 rename vendor/{virtualenv-1.11.6.py => virtualenv-12.0.7.py} (80%)
```

Changaco:
Zips/Wheels won't solve the repo size problem.

techtonik:
Which size we need?

whit537:
https://pip.pypa.io/en/latest/user_guide.html#fast-local-installs

I'm implementing this on whit537.org; will report back.

whit537:
Looks like we're already using `--no-index --find-links`. We're not using `--download` with that, and that's what would get the tarballs out of the repo. However, `--download` is insecure; that's where `peep` comes in (cf. gratipay/inside.gratipay.com#30):

[...]

PR #3292, opened 2015-03-26
Remove paths in requirements.txt with requirement specs

techtonik:
This allows to store dependencies everythere else (#1459)
and may lead to better performance if we cache wheels,
because it looks like pip still unpacks every tar.gz regardless
of if mentioned version is already installed.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -23,8 +23,8 @@ env: requirements.txt requirements_tests.txt setup.py
 				--extra-search-dir=./vendor/ \
 				--always-copy \
 				./env/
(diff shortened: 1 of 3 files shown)


techtonik:
Saved another 7 seconds. No need for wheels, except that they are better compressed in repository.

Changaco:
We should try caching `node_modules/`.

techtonik:
At least JS downloads need to be cached. For `node_modules/` I still don't know it works. Not enough confidence to do the change.

Issue #761, opened 2013-03-19
don't use Google Analytics


whit537:
Per @buchuki.

whit537:
Piwik PRO is $65/mo and afaict is reasonable wrt user privacy. From the [TOS](https://cloud.piwik.pro/tos/):

> 6.2.2 Piwik PRO may collect data on behalf of Customer while hosting Piwik Software on Piwik PRO servers; this data is fully owned by Customer and Piwik PRO does not make any use of such data and do not share it with third parties.

whit537:
Piwik trial started:

![screen shot 2015-03-12 at 5 43 46 pm](https://cloud.githubusercontent.com/assets/134455/6628867/62ebeccc-c8df-11e4-87b1-cfdc765940de.png)

whit537:
I believe that anyone with access to the shared folder on LastPass should be able to access https://gratipay.piwik.pro/. Can anyone confirm?

solarimix:
Folks should do further analysis of Google's Analytics platform re: privacy concerns and keeping your data secure from malicious "non state actors" like big corporations who care nothing about small entrepreneurs or small businesses or orgs.

AwStats, and Piwik are great alternatives.  hoping to see more adoption of GNU in new tech and more web statistics analysis

I'm also partial to StatCounter
http://statcounter.com/about/mission/

...both sitemeter and statcounter also provide free and paid web analytics tiers, not sure how usable that is for an app or a web app.

cheers!

webmaven:
@whit537, just tried to access, and it didn't work:

```
Error: Wrong Username and password combination.
```

whit537:
I'm unable to access Piwik as well. I can get into our Piwik Pro account management interface, but not the Piwik instance itself.

clone1018:
Maybe you have to make actual users for the piwik instance

whit537:
@clone1018 Yeah, I did make one but I lost the password. Reset now, with a second entry in LastPass. 

Try again with "Piwik (Analytics Dashboard)"?

webmaven:
Works now with the 2nd acct.

whit537:
@webmaven Yay! Do you know enough about Piwik to say that it's adequate for our needs? Or would you need to do some evaluation vis-a-vis Google Analytics first?

webmaven:
Would need to do some evaluation personally (and I can't give you a timeline for that right now), but by most accounts I've heard/read it should be adequate.

whit537:
[...]

PR #3299, opened 2015-03-30
Drop Google Analytics; #761

whit537:


www/about/policies/privacy.spt
```diff
@@ -93,16 +93,16 @@ the recipients of your gifts, so that particular gifts are anonymous. We do
 publicly share aggregated information about your giving and receiving. You may
 opt out of publicly sharing your aggregate giving and/or receiving.</p>
 
-<p>We also share your personal information with these third party vendors who
-work on our behalf:</p>
+<p>We may also share your personal information with these third party vendors
```
whit537: "May" because we're not actually sending data to Optimizely right now.

PR #3305, opened 2015-04-01
Fix an i18n bug

Changaco:
Currently `_("Member")` returns `('Membre', 'Membres')` (with the french locale), this PR works around that Babel defect by modifying our `get_text` function to return only the first string of the tuple.


whit537:
Worth [opening](https://github.com/mitsuhiko/babel/issues/new) a Babel ticket, if you haven't already?

Issue #3308, opened 2015-04-02
Add an account type selector to the bank account page


Changaco:
So we can support `savings` accounts in addition to `checking`. [FD1867](https://gratipay.freshdesk.com/helpdesk/tickets/1867)

PR #3282, opened 2015-03-24
Exchange routes

Changaco:
Closes #1681 and #2446. Unblocks #2824 and #3245.

PR #3309, opened 2015-04-02
Fix bank account page

Changaco:
Implements #3308, also fixes a regression and a broken link.


Changaco:
The regression comes from 101b87fd3025d0e45452dd59d778ea3bf4b0b55d in #3282. The problem is that you can modify your bank account if you've already connected one, but you can't add one otherwise.

Changaco:
[IRC](https://botbot.me/freenode/gratipay/2015-04-02/?msg=35691150&page=1)

PR #3311, opened 2015-04-02
Fix cc/ba forms wrt xxxxx123

whit537:
We were filling the form fields with bad data, which results in errors for the cc form, and (what's worse) silent failures for the ba form.


whit537:
[IRC](https://botbot.me/freenode/gratipay/2015-04-02/?msg=35691793&page=1)

Issue #2779, opened 2014-09-23
backfill the `status`, `route`, and `ref` columns of the `exchanges` table


whit537:
Right now the `exchanges.status` column has NULL values. It would be great to constrain that column to non-NULL, but it would take some effort with the Balanced API to determine the proper value for previous exchanges. [IRC](https://botbot.me/freenode/gratipay/2014-09-23/?msg=22175953&page=2)
##

Changaco:
Added to the [Balanced shutdown](https://github.com/gratipay/gratipay.com/milestones/Balanced%20shutdown) milestone.

whit537:
We have a user getting bitten by #3302, which we decided to fix with this ticket.

https://gratipay.freshdesk.com/helpdesk/tickets/1867

whit537:
I'm working on manually fixing this for the 1867 case, and am blocking payday (#3307) on it.

whit537:
[IRC](https://botbot.me/freenode/gratipay/2015-04-02/?msg=35700392&page=2)

whit537:
I believe the `destination` parameter here identifies the route for the most recently failing deposit:

https://dashboard.balancedpayments.com/#/marketplaces/MP12Xw5lL6iaILtqImIoroDL/events/EVfe85f7b4d70211e4b84a0663eb3e293e

However, the same parameter on the previous nine deposits (some successful, some not) is not present in `exchange_routes.address`:

https://dashboard.balancedpayments.com/#/marketplaces/MP12Xw5lL6iaILtqImIoroDL/events/EV3c627cc8d15011e496a0065bec47279b
https://dashboard.balancedpayments.com/#/marketplaces/MP12Xw5lL6iaILtqImIoroDL/events/EV9610ff60c92111e4b8dc02da29bdd175
https://dashboard.balancedpayments.com/#/marketplaces/MP12Xw5lL6iaILtqImIoroDL/events/EV864e9340c69111e4a8dc02da29bdd175
https://dashboard.balancedpayments.com/#/marketplaces/MP12Xw5lL6iaILtqImIoroDL/events/EV58831afac2bb11e49819065bec47279b
https://dashboard.balancedpayments.com/#/marketplaces/MP12Xw5lL6iaILtqImIoroDL/events/EV2f43cbb2b89811e4a512065bec47279b
[...]

PR #3314, opened 2015-04-02
Fix handling of exchange routes

Changaco:
For https://github.com/gratipay/gratipay.com/issues/2779#issuecomment-89053807 etc.

### DIFF
--- a/gratipay/billing/exchanges.py
+++ b/gratipay/billing/exchanges.py
@@ -209,7 +209,7 @@ def capture_card_hold(db, participant, amount, hold):
     assert participant.id == int(hold.meta['participant_id'])
 
     route = ExchangeRoute.from_address(participant, 'balanced-cc', hold.card_href)
-    assert route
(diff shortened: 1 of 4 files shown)


whit537:
Test?

Changaco:
Sleep first. :-)

whit537:
!m @Changaco 

:sleeping:

:-)

Changaco:
Tests added.

whit537:
I've confirmed [in Sentry](https://app.getsentry.com/gratipay/gratipay-com/group/60437761/) that the route is not `None` for the FD1867 case.

whit537:
What is it about `ExchangeRoute` that `assert route` evaluates false?

whit537:
```python
    def __bool__(self):
        return self.error != 'invalidated'
```

https://github.com/gratipay/gratipay.com/blob/1818/gratipay/models/exchange_route.py#L11-L12

whit537:
Running tests locally ...

Issue #1963, opened 2014-01-30
automate bitcoins payins


sim6:
Please add support for automate bitcoins payins.

rohitpaulk:
No more bitcoin, closing this.

sim6:
More info at https://github.com/gratipay/inside.gratipay.com/issues/201#issuecomment-102010517

Issue #3295, opened 2015-03-30
Add exchange routes to record-an-exchange.spt


rohitpaulk:
Reticketed from #3282

Issue #2178, opened 2014-03-24
give Coinbase access to others


whit537:
Reticketing from #2177, which is about PayPal and got slightly side-tracked for Balanced. The trouble with Coinbase is that we don't get user management. I can't even hand out a password, because I've got two-factor auth set up and so I get SMS on my cell phone when logging in. I've asked for user management on [Twitter](https://twitter.com/whit537/status/447226450190422016), need to probably escalate.

whit537:
Moot with https://github.com/gratipay/inside.gratipay.com/issues/201#issuecomment-102010517.

PR #3317, opened 2015-04-05
Bitcoin payin

rohitpaulk:
#1963

- UI Work
  - [x] Add bitcoin to settings page
  - [x] Create bitcoin-payin page
  - [x] Create coinbase button on request, embed iframe
  - [x] Handle messages from iframe on success/failure
  - [x] Fix how bitcoin payins are shown on the 'history' page
  - [x] Modify receipts to show bitcoin payins too
- [x] Handle coinbase callback
  - [x] Update DB, balance
  - [x] ~~Handle mispayments - overpaid, underpaid.~~ Skip for now, handle via support.
  - [x] Send emails
  - [x] Check for a secret_key
- [x] Handle fees properly
- [x] Handle both sandbox and non-sandbox mode
- Think about security
  - [ ] Do we need identity verification to allow a bitcoin payin?


rohitpaulk:
> A mispayment is an order payment that was received under abnormal circumstances. Two triggers will cause a mispayment to occur:

> The customer’s payment was received after the 10-minute order timeout expired.
> The customer’s payment was greater than or less than the actual order amount.

Mispayments are kinda tricky. We have three options -

### # 1:  accept all mispayments and add them to the user's balance

We do receive a [callback](https://developers.coinbase.com/docs/merchants/callbacks) everytime a mispayment is made. 

screenshot from 2015-04-06 00 24 44

Coinbase's UI will always show something like ^^ though, so it might be weird if we just silently ignored the problem and credited the user's account. 

Maybe send an email with details?
> "We received an amount different that you intended to pay, we've added it to your balance. Email support if you'd like a refund". 

Under the hood, this will make stuff a bit complicated though. We link every `exchange` to a `route` that contains the order ID as the `address`. All mispayments belong to the same order (`route`, in our case), so when we do create new exchanges for each mispayment, where would we store the `mispayment_id`? In the exchange's `note` column?

### # 2:  refund all mispayments as soon as we receive them

We'd have to collect a **refund address** from the user for this. This isn't required for someone with a coinbase account, but that's unlikely - people who have a coinbase account would pay **using** coinbase, which would never cause a mispayment to happen in the first place. Should we ask the user to add a bitcoin address to his profile before he can make a bitcoin payin?

### # 3: Don't do anything, handle through support.

That'd be very bad UX.

rohitpaulk:
A few screenshots:
screenshot from 2015-04-10 14 47 53

screenshot from 2015-04-10 14 47 22

screenshot from 2015-04-10 14 48 55

screenshot from 2015-04-10 14 34 00

rohitpaulk:
Notes on deployment: 

- Generate a random string and add it as `COINBASE_SECRET_KEY` in heroku config
- In Coinbase, set the default callback URL to `https://gratipay.com/callbacks/coinbase?secret_key=the_secret_key_generated_above`
- Set `COINBASE_SANDBOX_MODE` to `false` in heroku config
- Fill our merchant profile on coinbase, add logos etc.
- Fill all redirect URLs in merchant settings page to `/about/me` or something similar.

www/callbacks/coinbase.spt
```diff
@@ -0,0 +1,51 @@
...
+    # Check if a route corresponding to the order ID already exists.
+    route = ExchangeRoute.from_address(participant, 'coinbase-payin', order['id'])
+
+    # If not, create one.
+    if not isinstance(route, ExchangeRoute):
+        route = ExchangeRoute.insert(participant, 'coinbase-payin', order['id'])
+
+    status = order['status']
+
+    # We're ignoring mispayments/expired orders here.
```
Changaco: It doesn't seem very wise to completely ignore errors.

www/%username/routes/bitcoin-payin.json.spt
```diff
@@ -0,0 +1,35 @@
...
+    )
+    out['code'] = result.code
+
+    if website.coinbase_sandbox_mode:
+        out['iframe_base_url'] = "https://sandbox.coinbase.com/checkouts/"
+    else:
+        out['iframe_base_url'] = "https://www.coinbase.com/checkouts/"
+
+except APIError:
+    raise Response(500, "We're having trouble communicating with Coinbase, please try again.")
```
Changaco: 500 → 502 and i18n.

Changaco:
I don't like the changes you've made to `exchanges.py` and `test_billing_exchanges.py`. `record_exchange` should not do things differently depending on the route, it's only supposed to consider the status and the amount.

gratipay/utils/history.py
```diff
@@ -69,8 +69,12 @@ def iter_payday_events(db, participant, year=None):
 
     username = participant.username
     exchanges = db.all("""
-        SELECT *
-          FROM exchanges
+        SELECT *,
+              ( SELECT r.*::exchange_routes
+                   FROM exchange_routes r
+                  WHERE r.id = e.route
+              ) AS route
```
Changaco: Not very important, but the indentation is inconsistent here.

rohitpaulk:
> I don't like the changes you've made to exchanges.py and test_billing_exchanges.py. record_exchange should not do things differently depending on the route, it's only supposed to consider the status and the amount.

What if we want to record an exchange that we know is successful? Would we have to call `record_exchange` first, and then call `record_exchange_result`?

And `record_exchange` was already doing different things depending on the route, it was updating the balance for bank transfers but not for credit card charges. Just that the check was for `amount < 0` and not the route.

www/callbacks/coinbase.spt
rohitpaulk: Expired orders don't matter to us at all, but we could do more with mispayments.

We do show client side error messages like shown in the last screenshot in  https://github.com/gratipay/gratipay.com/pull/3317#issuecomment-91491278. Does sending an email sound good to you? The other options are listed in https://github.com/gratipay/gratipay.com/pull/3317#issuecomment-89835622, I think they're unnecessary and too much of work for now.

rohitpaulk:
@Changaco - Also, this change will allow us to reuse `record_exchange` in `record-an-exchange.spt` once https://github.com/gratipay/gratipay.com/issues/3295 is done.

techtonik:
Is there a diagram or an algorithm about what's going on (mispayment etc.)?

rohitpaulk:
@techtonik - No, there isn't. I can explain a bit though. 

When a user clicks 'pay' on the `routes/bitcoin-payin.html` page, we send coinbase a request to create a [button](https://developers.coinbase.com/api#buttons), and we use the response to insert an iframe into the page. Now, three things can happen. 

1. The user successfully completes his order.
2. The order expires (10 min window) and the user doesn't pay anything.
3. The user sends a wrong amount, or sends an amount after the 10 minute window. (Mispayment).

For each of these cases, we receive a callback at `www/callbacks/coinbase.spt`. The callback code only hits our db if the order is **successful**. It creates an exchange, updates the balance of the user, and sends them an email too. 

**Case 2** (expiry) isn't a problem for us, the user probably didn't intend to pay at all. 

**Case 3** is tricky, as mentioned in https://github.com/gratipay/gratipay.com/pull/3317#issuecomment-89835622. We just show the user an error message (coinbase does the same in the iframe) that asks them to contact support to retrieve the mispaid bitcoin. Refunding the mispaid bitoin is something that one of us (only @whit537 for now, see #2178) can do through Coinbase's UI. 

screenshot from 2015-04-10 14 34 00

Changaco:
> What if we want to record an exchange that we know is successful? Would we have to call `record_exchange` first, and then call `record_exchange_result`?

Either that or modify `record_exchange` properly.

> And `record_exchange` was already doing different things depending on the route, it was updating the balance for bank transfers but not for credit card charges.

No, that's not what it's doing. Although it's not implemented it is possible to do payins via `balanced-ba` and payouts via `balanced-cc`. Your changes are breaking `record_exchange`'s handling of these potential cases. The route is irrelevant, what matters is the exchange's `amount` (i.e. whether it's a payin or payout), and its `status`.

rohitpaulk:
@Changaco - https://github.com/gratipay/gratipay.com/commit/efef5891418ba561c9c61e2c404d9f1c17c67f29, https://github.com/gratipay/gratipay.com/commit/b6e13e2331f10e9d307a038b9ffe3a0e07915c79

www/callbacks/coinbase.spt
```diff
@@ -0,0 +1,52 @@
...
+if order:
+    # Fetch the participant from the order's 'custom' attribute.
+    participant = Participant.from_id(int(order['custom']))
+
+    # Check if a route corresponding to the order ID already exists.
+    route = ExchangeRoute.from_address(participant, 'coinbase-payin', order['id'])
+
+    # If not, create one.
+    if not isinstance(route, ExchangeRoute):
+        route = ExchangeRoute.insert(participant, 'coinbase-payin', order['id'])
```
Changaco: That's a race condition. You should try the `insert` first, and do the `from_address` if you get an `IntegrityError`.

www/callbacks/coinbase.spt
```diff
@@ -0,0 +1,52 @@
...
+request.allow('POST')
+
+if request.qs.get('secret_key') != website.coinbase_secret_key:
+    raise Response(401)
+
+order = request.body.get('order', None)
+
+if order:
+    # Fetch the participant from the order's 'custom' attribute.
+    participant = Participant.from_id(int(order['custom']))
```
Changaco: Should probably `assert participant` here.

rohitpaulk: Fixed.

rohitpaulk:
We're dropping bitcoin, closing this.

Issue #3107, opened 2015-01-10
Adjust tips from /%username/giving summary


colindean:
It'd be great if I could adjust tips in one central location instead of having to go to each account's profile.
##

whit537:
We had that functionality on the Giving page, but lost it in #2938. I agree we should bring it back.

rohitpaulk:
+1 on [FD#1603](https://gratipay.freshdesk.com/helpdesk/tickets/1603)

whit537:
+1 on https://github.com/gratipay/inside.gratipay.com/issues/131.

whit537:
+1 from @colinta on gratipay/inside.gratipay.com#131.

whit537:
+1 from @JeffSpies in https://github.com/gratipay/gratipay.com/issues/3177:

> In the "Giving" section, I can no longer adjust amounts in batch, but I have to click on each user and then edit my giving there. My first thought is that this is purposeful inconvenience, which doesn't sit well with me. I doubt it is purposeful, knowing this team, but that's probably not the desired intuition. Takeaway:
> 
>  1. I strongly prefer the previous interface where you can adjust giving all on one page.

whit537:
+1 from [UkuleleRod](https://gratipay.com/UkuleleRod/) IRL.

nobodxbodon:
Merging with [Make it easier to add new recipients on the "Giving" screen](https://github.com/gratipay/gratipay.com/issues/716). Maybe a more thorough approach is to allow all of CRUD.

rohitpaulk:
Here's how the giving page looks right now: 

![screen shot 2017-06-17 at 12 22 18 pm](https://user-images.githubusercontent.com/3893573/27250873-aae813f2-5357-11e7-8add-d37d80ec2fb8.png)

We'll need more area to offer CRUD functionality, might have to scrap the table in favor of rows. Something like: 

![screen shot 2017-06-17 at 12 17 42 pm](https://user-images.githubusercontent.com/3893573/27250879-c324ce56-5357-11e7-9acf-e4ced8c39f3b.png)

![screen shot 2017-06-17 at 12 18 00 pm](https://user-images.githubusercontent.com/3893573/27250880-c32d301e-5357-11e7-9011-02a2f49ba1c4.png)

whit537:
I like the idea of moving from the tabular format to a style more parallel to what we do with projects on the homepage, search, etc. as well as packages on /on/npm/.

PR #3325, opened 2015-04-10
Improve giving page - part 1

rohitpaulk:
This is cleanup before fixing #3107.

### DIFF
--- a/js/gratipay/giving.js
+++ b/js/gratipay/giving.js
@@ -1,25 +1,29 @@
 Gratipay.giving = {}
 
 Gratipay.giving.init = function() {
-    Gratipay.giving.activate_tab('active');
-    $('.giving #tab-nav a').on('click', Gratipay.giving.handle_click);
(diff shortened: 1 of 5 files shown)


rohitpaulk:
Before:
screenshot from 2015-04-11 00 43 39

After:
screenshot from 2015-04-11 00 42 54
screenshot from 2015-04-11 00 42 58
screenshot from 2015-04-11 00 43 02

Changaco:
I don't see why you're using server-side tabs. On the history page it's because there is too much data to send at once, but here it's not the case.

rohitpaulk:
@Changaco - Hmm, yh. I was thinking of expanding each row in the table to a bigger item with the user's avatar + buttons to edit or cancel, so wanted to reduce the amount of info we show at a time on the page.

You're right, client side tabs would definitely make more sense.

whit537:
wfm

!m @rohitpaulk

PR #3330, opened 2015-04-12
Clean up profile where bitcoin buttons are wrapping

skj3gg:
The styling for the buttons in the add a bitcoin receiving option was a bit wonky.

Before:
screen shot 2015-04-12 at 1 26 12 pm

After:
screen shot 2015-04-12 at 1 32 54 pm


whit537:
Thanks for the PR, @skj3gg! :-)

Issue #3331, opened 2015-04-13
The bottom nav is invisible


techtonik:
The bottom bar is hidden through scrollback, so nobody sees it. Discovered the fact only because was looking how to deal with #2850.

![nobottom](https://cloud.githubusercontent.com/assets/515889/7111475/c168941a-e1c8-11e4-8949-85442deace5d.png)

whit537:
Can you be more specific? I thought maybe you were referring to the footer, but I can see that just fine:

![screen shot 2015-04-13 at 7 34 28 am](https://cloud.githubusercontent.com/assets/134455/7114835/902c540e-e1af-11e4-9aa0-e284fd07b1cc.png)

techtonik:
Yes. I am referring to footer. It is shown only when you scroll down no matter how small the page content is.

whit537:
Ah, gotcha.

whit537:
Addressed in #3718.

Issue #3165, opened 2015-02-10
Failing tests when using Vagrant


techtonik:
`master` is currently failing for Vagrant.

```
=================================== FAILURES ===================================
________________________ TestPages.test_anon_can_browse ________________________

self = <test_pages.TestPages testMethod=test_anon_can_browse>

    def test_anon_can_browse(self):
>       self.browse()

tests/py/test_pages.py:30: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
tests/py/test_pages.py:24: in browse
    r = self.client.GET(url, **kw)
env/lib/python2.7/site-packages/aspen/testing/client.py:81: in GET
    def GET(self, *a, **kw):    return self.hit('GET', *a, **kw)
env/lib/python2.7/site-packages/aspen/testing/client.py:105: in hit
    , return_after=return_after
env/lib/python2.7/site-packages/aspen/website.py:63: in respond
    , _return_after=return_after
env/lib/python2.7/site-packages/algorithm.py:288: in run
    new_state = function(**deps.as_kwargs)
env/lib/python2.7/site-packages/aspen/algorithms/website.py:95: in get_resource_for_request
    return {'resource': resources.get(website, dispatch_result.match)}
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

website = <aspen.website.Website object at 0x2b8d835cefd0>
fspath = '/vagrant/www/about/me/%redirect_to.spt'

    def get(website, fspath):
        """Given a website and a filesystem path, return a Resource object (with caching).
        """
    
        # XXX This is not thread-safe. It used to be, but then I simplified it
        # when I switched to diesel. Now that we have multiple engines, some of
        # which are threaded, we need to make this thread-safe again.
    
        # Get a cache Entry object.
        # =========================
    
        if fspath not in __cache__:
            entry = Entry()
            __cache__[fspath] = entry
    
        entry = __cache__[fspath]
    
    
        # Process the resource.
        # =====================
    
        mtime = os.stat(fspath)[stat.ST_MTIME]
        if entry.mtime == mtime:  # cache hit
            if entry.exc is not None:
                raise entry.exc
        else:  # cache miss
            try:
[...]

PR #3339, opened 2015-04-15
Fixed the bottom footer. Made it fixed on scrolling.

kaguillera:
Please review if this is what you want for #3331 @techtonik

### DIFF
--- a/scss/layouts/layout.scss
+++ b/scss/layouts/layout.scss
@@ -164,6 +164,12 @@ body {
     text-align: right;
     font-size: 13px;
     line-height: 13px;
+    
+    position: fixed;
+    left: 0px;
+    bottom: 0px;
+    height: 40px;    
+    width: 100%;
     @include clearfix;
 
     li {
@@ -181,3 +187,4 @@ body {
         }
     }
 }
+


techtonik:
I am still struggling to get my working environment at #3165, so I can't test ATM.

kaguillera:
No problem

techtonik:
There is still a scrollbar even when there is no content. 

gratipay_-_scrollbar

kaguillera:
What browser are you using. I may have to test it on all (or as much as I possibly can)

techtonik:
I have only Chrome and Firefox, and I believe that was Chrome. In any case it seems that the issue is caused by the way the header is added to the page, because the height of scrolled area looks equal to header height.

techtonik:
Also, fixed border may eat up precious space on mobile devices. So it would be nice to just fix the screen height calculation to avoid the margin / padding / whatever makes this taller than it is,

techtonik:
Found a way to fix our layout http://stackoverflow.com/questions/8257311/modern-way-to-markup-100-height-layout-in-html5-and-css3

kaguillera:
!m @techtonik I wil try it

techtonik:
@kaguillera, go ahead! =)

I played a bit with CSS table layout in Google Chrome dev tools, but I am not sure how it will work in other browsers. I gave <body> a  `display:table`, moved fixed size <header> and <footer> out of #wrapper (I don't know why as extra #wrapper is needed there), and set `display:table-row` on them. Then I have to wrap central #main into additional <div>, because #main should be a `display:table` for managing central columns height to 100%, and top level needs `display:table-row` to redistribute its height among children.

I also used `border-collapse: collapse` on body table, because border became invisible. And then there was something to do with bottom footer to fix content alignment due to changed model from div to table row.

whit537:
Obsolete with #3718 .

Issue #3343, opened 2015-04-16
unblock charge notification emails


whit537:
First production run got stuck. Reticketing from #3341.

whit537:
1097 mails in the `email_queue`. There was a Sentry error about this ...

whit537:
https://app.getsentry.com/gratipay/gratipay-com/group/62939278/

whit537:
Looks like we're simply queueing mails for participants who don't have an email attached.

whit537:
We should probably add that check to `queue_email`.

whit537:
Aaaand I suppose an alternate fix would be to perform the check in `send_email`. That would have the added benefit of clearing out our `email_queue` for us.

whit537:
Email queue is empty! :-)

whit537:
https://twitter.com/Gratipay/status/589073378016899072

PR #3345, opened 2015-04-16
only queue mail if we have an address

whit537:
Fixes root cause of #3343.


whit537:
Closing in favor of alternate approach.

PR #3346, opened 2015-04-16
silently skip emails when we don't have an address

whit537:
This is a fix for #3343.

### DIFF
--- a/gratipay/models/participant.py
+++ b/gratipay/models/participant.py
@@ -546,10 +546,11 @@ class Participant(Model, MixinTeam):
         username = self.username_lower
         quoted_email = quote(email)
         link = "{scheme}://{host}/{username}/emails/verify.html?email={quoted_email}&nonce={nonce}"
-        self.send_email('verification',
+        r = self.send_email('verification',
                         email=email,
                         link=link.format(**locals()),
                         include_unsubscribe=False)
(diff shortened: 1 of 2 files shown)


whit537:
Alternate to #3345.

whit537:
I considered skipping the `sleep` call when we don't actually make an API call, but if I'm reading the rest of the email infrastructure right that wouldn't actually speed up queue processing at all. It looks like we call `dequeue_mails` every 60 seconds with max 60 messages, so even if we only mailed 30 of them, we would still wait another ~30 seconds until proceeding to the next batch.

whit537:
Apparently I botched that first commit. Oh well. :-/

whit537:
Ready for review @rohitpaulk @Changaco et al.

rohitpaulk:
> It looks like we call dequeue_mails every 60 seconds with max 60 messages, so even if we only mailed 30 of them, we would still wait another ~30 seconds until proceeding to the next batch.

That could change in the future, so I've changed it to only sleep if we're actually making an API call.

rohitpaulk:
Email queue is down to 0 :dancer:

whit537:
https://twitter.com/Gratipay/status/589073378016899072 :dancer:

PR #3349, opened 2015-04-18
Fix amount failed charges

### LINKED ISSUE #3347: respect $10 minimum in charge failure notification
From [FD1980](https://gratipay.freshdesk.com/helpdesk/tickets/1980) it appears that we're not doing this.

screen shot 2015-04-17 at 10 38 40 am

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/11369919-respect-10-minimum-in-charge-failure-notification?utm_campaign=plugin&utm_content=tracker%2F85909&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F85909&utm_medium=issues&utm_source=github).
</bountysource-plugin>

rohitpaulk:
fixes #3347

### DIFF
--- a/gratipay/billing/exchanges.py
+++ b/gratipay/billing/exchanges.py
@@ -179,6 +179,7 @@ def create_card_hold(db, participant, amount):
     # ===============
 
     cents, amount_str, charge_amount, fee = _prep_hit(amount)
+    amount = charge_amount - fee
     msg = "Holding " + amount_str + " on Balanced for " + username + " ... "
 
     hold = None
--- a/tests/py/test_billing_exchanges.py
+++ b/tests/py/test_billing_exchanges.py
@@ -77,7 +77,8 @@ class TestCardHolds(BalancedHarness):
         assert error == "Foobar()"
         exchange = self.db.one("SELECT * FROM exchanges")
         assert exchange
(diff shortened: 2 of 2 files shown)


gratipay/billing/exchanges.py
```diff
@@ -179,6 +179,7 @@ def create_card_hold(db, participant, amount):
     # ===============
 
     cents, amount_str, charge_amount, fee = _prep_hit(amount)
+    amount = charge_amount - fee
```
whit537: Are you sure this is the right way to do this? On my reading this variable only gets used in the exception case below. How do we call `record_an_exchange` in the success case? What is it expecting for amount? Doesn't this pass different values in the success and error cases?

rohitpaulk: > How do we call record_an_exchange in the success case?

We don't. For successful card holds, we don't record an exchange at all. We create an exchange when we capture them, and [this same statement](https://github.com/gratipay/gratipay.com/blob/master/gratipay/billing/exchanges.py#L215) is present there.

PR #3364, opened 2015-04-25
Change SSN field to password field

SirCmpwn:
It seems to make sense that this field wouldn't be shown on-screen while you're punching your info in.


whit537:
Thanks @SirCmpwn! :-)

Issue #3381, opened 2015-05-04
gravatar image incredibly blurry


whit537:
Reticketed from #3227:

> My gravatar image comes in at 80x80 and is incredibly blurry: https://gratipay.com/Jonathan%20Sundqvist/
##

jonathan-s:
Looking at [this](https://gratipay.com/akshayah3/) page for instance I see that the reference to the image is the following `https://avatars.githubusercontent.com/u/4225167?s=160` I'm assuming that this is stored in a database. I would strip all query tags from the URLs and then implement the solution in my PR. 

The avatars from github follow the gravatar protocol after all. Are there any more image suppliers?

whit537:
Actually, that's not sufficient, because we also use non-Gravatar avatars. I've reverted your PR in #3384. Sorry for the confusion.

jonathan-s:
What are the other image sources that you use?

whit537:
The various social networks we support. Here's a grep:

https://github.com/gratipay/gratipay.com/search?utf8=%E2%9C%93&q=avatar_url

jonathan-s:
Ok, so what I gather is that you pull images from twitter and facebook too. Perhaps it could be solved with some logic in the template to check where the image comes from.

PR #3383, opened 2015-05-04
Avoid blurry images and be retina ready

jonathan-s:
Reference to issue #3381. It's described [here](https://de.gravatar.com/site/implement/images/) that you can resize gravatar images. Since 320px is the double of 160px, the gravatar should also be retina ready.


whit537:
Thanks @jonathan-s! :-)

Issue #3377, opened 2015-04-30
migrate to new processing infrastructure


whit537:
Balanced is going out of business. On #67 we've been designing our new processing infrastructure. This ticket is to coordinate implementation of our design. Here are the pieces:

 - Braintree for vaulting and processing cards (#3287), settling to Citizens
 - Gratipay for storing bank accounts and identity information
 - Citizens for ACH credits (#3366; possibly/probably via OpenACH, #3371)
 - Gratipay for compliance (with professional help: https://github.com/gratipay/inside.gratipay.com/issues/193)

Here's a spreadsheet with a timeline of targets we need to hit to pull this off:

whit537:
To: Balanced

> Greetings, friends. :-)
>
As June 11 approaches, we've been hard at work over here sorting out our migration plan. At this point it looks like we're not going to be moving to Stripe. Instead, we're planning to move credit card processing to Spreedly, and ACH credits and identity verification in-house to Gratipay. We're putting concrete plans together here:
>
https://github.com/gratipay/gratipay.com/issues/3377
>
At this point my question is: Will you be able to support us in this plan? Anything special we should be aware in not following the well-trodden Stripe route?

whit537:
I'm seeing two "tracks" here:

 - **Migrating processing**:
  - start double-storing new data in Balanced and in new stores (Spreedly/Gratipay)
  - export data from Balanced to new stores
  - start processing from new stores
  - stop storing new data in Balanced
 - **Salvaging data**: #2443 and #2779 (depends on #3295, PR is #3356)

whit537:
It's more important to migrate processing than to salvage data, so we should do that first. We can supposedly query transactions until October 9, but we can't process after June 11.

whit537:
One question I have is whether we cut over to Spreedly and Citizens ACH on the same week, or different weeks. It's probably less risky to stage them. Which would we do first, all things considered?

whit537:
We just don't have a ton of time, though.

whit537:
Braintree is ready to go, we're still pending account setup with Citizens ACH. On the other hand, we're storing the ACH-side information in our own db, so we don't have to wait for that. I propose that we sync up double-storing for both cards and bank accounts/identity, and then take the processing migration in two steps:

[...]

Issue #3287, opened 2015-03-26
migrate to Braintree


whit537:
https://www.braintreepayments.com/features/marketplace

Changaco:
> A PayPal Company

That worries me, and their [Acceptable Use Policy](https://www.braintreepayments.com/legal/acceptable-use-policy) isn't reassuring.

chrisdev:
Yep Paypal is a major pita but is this not way we  facilitate most of our payouts to non US users?

whit537:
@Changaco Their AUP looks pretty normal to me. What don't you like about it?

whit537:
I'm looking for things in their docs that we run afoul of.

Links off of https://apply.braintreegateway.com/signup/us:

https://www.braintreepayments.com/legal/payment-services-agreement
https://www.braintreepayments.com/legal/bank-agreement-us
https://www.paypal.com/us/webapps/mpp/ua/privacy-full
https://www.braintreepayments.com/legal/acceptable-use-policy
https://www.braintreepayments.com/features/marketplace
https://www.braintreepayments.com/faq#marketplace

whit537:
> Merchant accounts need to be specially approved for use with Marketplace. For more information, please contact support@braintreepayments.com.

https://developers.braintreepayments.com/ios+python/guides/marketplace/overview

whit537:
I've created a Braintree account.

whit537:
![screen shot 2015-04-09 at 6 52 21 pm](https://cloud.githubusercontent.com/assets/134455/7078699/9e49f5f0-dee9-11e4-8d2e-ca8d1a42f617.png)

whit537:
On the application they asked about our business and I said we provide "Sustainable crowd-funding for open-source projects" or something like that.

whit537:
>Thank you for applying with Braintree! I’ll be working with you to wrap up our application process. I only need a few more things from you before we move onto the next step.
>
1. Did you mean to apply for marketplace? Unfortunately our marketplace feature doesn’t support recurring billing. 
2. You listed your subscription billing as 75% monthly, 20% quarterly, 4% semi-annual and 1% annual. However your website states payments will be made weekly. Can you explain which recurring billing schedule your customers will be charged on?
3. How do you vet the individuals/businesses which will be accepting funds?
4. Can you please provide 3 current months of credit card processing statements to support your volume? These need to show a summary of processing volume, charge-backs and refunds.
[...]

Issue #3387, opened 2015-05-08
double-vault new cards at Balanced and Braintree


whit537:
Reticketed from #3377.

rohitpaulk:
Why double vault? Can't we just send the new cards to Braintree, and migrate old cards there too?

whit537:
@rohitpaulk I was thinking that that introduces a race condition, but since we don't process continuously and in fact aren't processing at all right now, maybe we should just go for it.

whit537:
In order to avoid the race condition, we need to start vaulting new cards at Braintree before the transfer is initiated. The trick is that we want to *complete* the transfer before running payday on Braintree.

rohitpaulk:
Yep. I'm working on vaulting new cards first for #3389. Once that's done, I'll visit the billing/exchanges functions.

whit537:
@rohitpaulk Awesome, thank you. Let's deploy the vault changes ASAP. Let's change billing/exchanges in a separate PR, please.

whit537:
Per https://github.com/gratipay/gratipay.com/issues/3377#issuecomment-100375980 there probably *is* a race condition, due to updates. I don't have time to rethink it all through again but we should.

rohitpaulk:
Simplest solution here is to send all new cards **and** updated cards to braintree. And of course, not run them until we've moved off balanced.

rohitpaulk:
AFAICT, we don't actually 'update' cards, we add a new payment method and [invalidate the old one](https://github.com/gratipay/gratipay.com/blob/master/www/%25username/routes/associate.json.spt#L46-L47).

rohitpaulk:
I'm closing this. #3389 is ready for review. Updating a card vaults it with Braintree, and invalidates the previous card at Balanced.

PR #3389, opened 2015-05-08
Vault new cards at Braintree

rohitpaulk:
#3377, #3287

- [x] Backwards compatibility for invalidation
- Fix credit card page and verify for following cases: 
  - [x] No credit card attached
  - [x] Card attached (balanced)
  - [x] Card attached (braintree)
  - [ ] Updates card (from balanced to braintree)
  - [x] Updates card (from braintree to braintree)
- [x] Fix receipt page


rohitpaulk:
https://developers.braintreepayments.com/javascript+python/guides/credit-cards#options

Looks like Braintree only requires the `postal_code`, and none of the other address fields.

whit537:
Per https://github.com/gratipay/gratipay.com/issues/3387#issuecomment-100373415, let's limit the scope of this PR to vaulting new cards on Braintree instead of Balanced. We want to deploy that ASAP, don't want to block it on `billing.exchanges` work.

whit537:
Per https://github.com/gratipay/gratipay.com/issues/3387#issuecomment-100384080 let's please double-vault to avoid a race condition unless we can think it through again from the top to ensure there isn't one.

whit537:
And !m @rohitpaulk. Happy to be [rustlin' dogies](https://github.com/gratipay/gratipay.com/issues/3377#issuecomment-100380277) with you, friend. :-)

rohitpaulk:
!m @whit537 :)

rohitpaulk:
@whit537 - I haven't gotten to the Braintree tests, but this is ready for you to look at and start testing manually.

whit537:
Awesome, thank you.

whit537:
In terms of timing on this, I think we should wait to deploy until the old cards are migrated to Braintree. Sound right, @rohitpaulk?

rohitpaulk:
> In terms of timing on this, I think we should wait to deploy until the old cards are migrated to Braintree. Sound right, @rohitpaulk?

Nope, this is ready to deploy now - it's backwards compatible with balanced cards.

rohitpaulk:
@whit537 - I'd look at this as a better alternative to #3392.

whit537:
Took a first read through the code. What happens if someone fills out the credit card form after Balanced initiates the export but before Braintree completes it?

rohitpaulk:
If they're filling out the cc form - the new details will be sent to Braintree and **if** there was an old card(Braintree or Balanced), it'll be deleted.

whit537:
> the new details will be sent to Braintree

And then what happens when Braintree imports the old details, which will be included in the snapshot that Balanced took before the user posted the cc form?

rohitpaulk:
Once braintree is done importing, we'll only use the accounts for those who have **not** updated their card details, and delete the others.

whit537:
Interesting. I guess I need to run this code and look at the sandbox to understand how that's possible. I guess when we start saving cards now we'll be generating new Braintree customer ids, right? I guess Braintree will create a second customer record for those customers when it does the import. Presumably Balanced sent along some info that will show up in the customer records at Braintree that will allow us to connect them with participants in our database, and we can deduplicate based on that. Is that it?

whit537:
Two more points:

 - Bring back bank accounts opens up the hole mentioned at https://github.com/gratipay/gratipay.com/pull/3392#issuecomment-100478987 again.
 - Should we include the braintree sandbox login in a comment so devs can inspect as needed?

whit537:
Regression: https://app.getsentry.com/gratipay/gratipay-com/group/65471207/.

whit537:
Steps to reproduce: visit https://gratipay.com/about/me/routes/credit-card.html.

whit537:
P.S. Does Braintree vault for us automatically or do we have to explicitly ask for cards to be stored permanently?

rohitpaulk:
> I guess when we start saving cards now we'll be generating new Braintree customer ids, right?

Yes.

> I guess Braintree will create a second customer record for those customers when it does the import. 
> Presumably Balanced sent along some info that will show up in the customer records at Braintree that will allow us to connect them with participants in our database, and we can deduplicate based on that.

I'm assuming that we'll be provided with a mapping between `balanced_customer_href`'s and `braintree_customer_id`s. We search for a participant via the `balanced_customer_href` and update their `braintree_customer_id` and  card route **only if** they don't have a braintree card route already.

rohitpaulk:
> Bringing back bank accounts opens up the hole mentioned at #3392 (comment) again.

As long as people can close their accounts (and use their bank account as the disbursement strategy), isn't that hole always open?

> Should we include the braintree sandbox login in a comment so devs can inspect as needed?

Hmm, we didn't do that for Balanced. I think it'd be best to wait for someone to ask.

> Does Braintree vault for us automatically or do we have to explicitly ask for cards to be stored permanently?

Automatically :dancer:

whit537:
> Hmm, we didn't do that for Balanced. I think it'd be best to wait for someone to ask.

Balanced did sandboxing differently. But yeah we can wait.

> [I]sn't that hole always open?

Yeah, but it's wider if people can connect new accounts.

>Automatically

Cool.

PR #3393, opened 2015-05-10
Merged Gratipay bootstrap scripts

techtonik:
This basically moves ./bin/install-deps.sh to ./scripts/bootstrap-debian.sh

I need to rework that to understand what is needed by core application and what is needed for tests, and make sure that both work on my machine under both Vagrant (Windows) and LXD (Linux). It appears that jstest dependencies are huge (Java). Should split them into separate bootstrap-debian-jstests.sh ?


techtonik:
!m @whit537

PR #3394, opened 2015-05-10
defaults.env: Remove localhost:8537 from CANONICAL_HOST

techtonik:
For development, if you run Gratipay in a container, the
IP can be any local address, not exactly 127.0.0.1, to
which localhost resolves.

The symptom is when Gratipay URL is redirected to
localhost:8537 and there is no content.


whit537:
Can you not override this in your own `local.env`?

techtonik:
@whit537 I want to create a fastlane for `checkout and run` development workflow, without any additional configuration steps. My workflow is that I do a lot of checkouts on different machines and configuring git is already an annoying hassle.

whit537:
One issue: the fonts only work properly on `localhost:8537`.

techtonik:
Is there an explanation of that?

techtonik:
So, why there is a problem with fonts?

whit537:
> So, why there is a problem with fonts?

It's because we use proprietary fonts, @techtonik. Typography.com performs `Referer` checking.

whit537:
@techtonik How about if we register `gratipay:8537` with Typography.com? That way someone who wants to develop on something besides `localhost` can set `gratipay` in their local `etc/hosts` file and get proper fonts. I'd be up for setting (now it's) `BASE_URL` to the empty string and documenting the font issue. "Hit me on localhost:8537 or gratipay:8537 to get proper fonts."

whit537:
Actually, I take that back. The normal "fast lane" case is serving on `localhost:8537`, so that is the right thing to have set in `defaults.env`. If you want to run somewhere else then you should set `BASE_URL`. However, a) Typography.com won't accept `gratipay` as a domain name, and b) you also need a way to login. Therefore, I've configured `gratipay.dev` at Typography.com, and have configured an alternate GitHub oauth application. I've updated the README in #3615. Closing.

techtonik:
@whit537 we use fallback fonts for development server? I remember I've seen an example for CSS somewhere, but I am not sure how to patch LESS. It will be useful to work on design offline.

whit537:
@techtonik Well, the browser will fall back eventually, but it takes a few seconds. It's a significant drag on development.

techtonik:
Few seconds? I thought that some kind of 403 should return immediately, so that browser can take another style. Maybe that's a problem with Typography that holds request for too long?

whit537:
@techtonik Feel free to explore and open a new PR/issue. :)

techtonik:
Okay. =)

PR #3400, opened 2015-05-11
promote Teams to first-class citizenship

whit537:
This will be at least part of #3399.

PR #3402, opened 2015-05-12
move profiles to /~username/

whit537:
This makes way for Teams to be at /team/. I cherry-picked these out of #3400, which has grown out of control.


whit537:
This is ready for review/merge, @rohitpaulk @techtonik et al.

whit537:
Hrm. This leaves all old profile links 404, because I implemented the redirects further on in #3400, once I had implemented a Team entity.

Issue #3399, opened 2015-05-11
turn the ship


whit537:
Last week, Gratipay [stopped processing](https://medium.com/gratipay-blog/gratipocalypse-42fd0ec0d9e8) payments because of legal concerns with our old terms of service that became apparent after our processor, Balanced, announced that they're going out of business, and Stripe wouldn't take us. So in addition to migrating our processing infrastructure (#3377), we also have to change our business model at the same time. No :sweat:, right? :)

We hashed out our new business model on https://github.com/gratipay/inside.gratipay.com/issues/180 and https://github.com/gratipay/inside.gratipay.com/issues/192. This ticket is to track implementation. See also the [Pivot](https://github.com/gratipay/gratipay.com/milestones/Pivot) milestone. 
##

whit537:
I'm stuck thinking about the relationship between individuals, groups, and projects. I think #3337 is the cleanest way to go:

>Users optionally have Projects.
Projects optionally have Payroll.

Projects would be what you give money to.

But right now Users are what you give money to. We don't want to drop existing funding streams if we can at all help it. We want to give Users a way to migrate to a Project. Forcing migration before we process payments is okay since we're requiring acceptance of our new terms anyway.

whit537:
I'm not sure how to pull of this pivot without #3337. That's a fair chunk of work.

whit537:
Let's think about **Individuals** (formerly Participants; Users above) and **Teams** (vs. Projects above).

How do we migrate giving from Individuals to Teams? We basically need Individuals to create a new Team, to which we will reattach their giving.

A Team is a group of people working to deliver a given product or service.

Teams have Members. Some Members are Owners.

Since giving to Individuals has been ambiguous in regard to products or services delivered, migrating giving to a single-purpose Team will be a judgement call for our users. If I have 100 supporters, and 60 would say they were giving to me because of one product or service I provide, and 30 because of another, and 10 just because they love me, then how should I migrate? I don't even know who my supporters are, so it's not like I can manually review and make assignments ("Oh, this person is a Fizzbuzz user, that's probably why they've been supporting me. Let's move their payment to the new Fizzbuzz Team"). No: I have to guess at the 60-supporter product/service, and migrate all of my giving to a new Team dedicated to that. Then I can make a second Team for the 30-supporter product/service.

whit537:
Can a Team give to another Team?

whit537:
> Can a Team give to another Team?

[...]

PR #3403, opened 2015-05-12
kill pledging

whit537:
With the move to Teams we no longer support pledging to arbitrary social network accounts. Incorporates #3402, works towards #3399.


whit537:
Ready for review/merge, @rohitpaulk @techtonik et al.

whit537:
3ba5d0d c2cf25a are failing at Travis because of a pyflakes violation. Presumably 1eb481e will fail with the same.

whit537:
Yup. Should be fixed in e67a9e1.

whit537:
Hmm ... how do I run pyflakes globally locally?

whit537:
`make pyflakes`

techtonik:
Not LGTM on killing the code that is useful for potential forks.

whit537:
@techtonik Don't worry, we use version control. ;-)

techtonik:
Right, but how somebody is supposed to find that? The branch info is erased and there is no like "feature tagging" to know which revisions still fully supported this feature.

techtonik:
And before killing the DB, I'd like to dump my data from it, for example, to know how generous I am and see which project won't be able to receive my donations anymore because of this pivoting.

PR #3329, opened 2015-04-12
The ultimate transparency is here. Yay! =)

techtonik:
This creates a basis for solving #3326. Please add more info.

PR #3404, opened 2015-05-12
Rename `make data` to `make fake`

whit537:
We introduced a `data` directory in #3329, which masks our phony target for populating the database with fake data. This cherry-picks d6372ed21e370bb83ea99ac45e90605e6074b405 from #3400.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -41,7 +41,7 @@ schema-diff: test-schema
 	diff -uw prod.sql local.sql
 	rm prod.sql local.sql
 
-data:
+fake:
 	$(honcho_run) $(env_bin)/fake_data fake_data
 
 run: env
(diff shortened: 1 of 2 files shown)


whit537:
Ready for review/merge, @techtonik et al.

techtonik:
I'd rename this to `make fakedata`. It purges DB, so it's dangerous right?

Also, what's the point in the level of redirection?
```
/setup.py:                        , 'fake_data=gratipay.utils.fake_data:main'
```
1. Why not just run ./gratipay/utils/fake_data.py directly?
2. Why not copy Gratipay scripts during `make env`?

This will remove 500k dependency from our source tree. I don't see where else it is used.

PR #3405, opened 2015-05-12
kill goal

whit537:
Since users will no longer receive payments directly, there's no need for them to be able to publish a personal goal. Our goal implementation was half-baked anyway. If we want goals for teams we should implement them properly (a la Kickstarter, Patreon, Salt, etc.).

Incorporates #3402 and #3403.


whit537:
At this point I'm planning to rebase this one a couple other PRs are landed first. See https://github.com/gratipay/gratipay.com/issues/3399#issuecomment-101391534.

whit537:
We want to turn off tipping (#3406) before we land this, because without this someone could tip a "patron-no-tips."

sql/branch.sql
```diff
@@ -0,0 +1,5 @@
+BEGIN;
+
+    ALTER TABLE participants DROP COLUMN goal;
```
rohitpaulk: @whit537 - Are you sure we want to start dropping columns right now? Wouldn't it make sense to leave it all in there, and delete after we're done migrating?

rohitpaulk: Okay, I see what you're doing - only deleting the columns that we won't find useful at all.

rohitpaulk:
Tipping has been disabled, I'm merging this.

whit537:
!m @rohitpaulk

PR #3408, opened 2015-05-13
new terms of service

whit537:
Closes #3390.

PR #3412, opened 2015-05-13
Team creation

whit537:
For #3399. cc: @rohitpaulk

### DIFF
--- a/gratipay/models/participant.py
+++ b/gratipay/models/participant.py
@@ -776,6 +776,14 @@ class Participant(Model, MixinTeam):
         """, (self.id,))
         return {r.network: r.address for r in routes}
 
+    @property
+    def has_payout_route(self):
+        for network in ('balanced-ba', 'paypal'):
+            route = ExchangeRoute.from_network(self, 'balanced-ba')
+            if route and not route.error:
+                return True
+        return False
+
 
     def get_balanced_account(self):
         """Fetch or create the balanced account for this participant.
--- a/gratipay/models/team.py
+++ b/gratipay/models/team.py
@@ -35,3 +35,18 @@ class Team(Model):
              WHERE {}=%s
 
         """.format(thing), (value,))
+
+    @classmethod
+    def create_new(cls, owner, fields):
+        return cls.db.one("""
+
+            INSERT INTO teams
+                        (slug, slug_lower, name, homepage, product_or_service,
+                         getting_involved, getting_paid, owner)
+                 VALUES (%s, %s, %s, %s, %s, %s, %s, %s)
+              RETURNING teams.*::teams
+
+        """, (fields['slug'], fields['slug'].lower(), fields['name'], fields['homepage'],
+              fields['product_or_service'], fields['getting_involved'], fields['getting_paid'],
+              owner.username))
+
--- a/js/gratipay/new_team.js
+++ b/js/gratipay/new_team.js
@@ -1,0 +1,28 @@
+Gratipay.new_team = {}
+
+Gratipay.new_team.initForm = function () {
+    $form = $('#new-team');
+    $button = $form.find('button');
+    $button.on('click', Gratipay.new_team.submitForm);
+}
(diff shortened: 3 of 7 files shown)


rohitpaulk:
@whit537 - I feel like we're approaching this the wrong way. We're going to deliver a bad user experience if we attempt to get this done before tomorrow. If only a handful of teams (including us) are going to take part in this payday, why not collect the question/answers from them (via email) manually? Shouldn't we focus on getting a few 'beta' users and processing payments for them first?

I'm suggesting that we let this (building an application process) wait, and focus on how we're going to run Payday this Thursday.

whit537:
I definitely agree that our focus is on getting payday run this Thursday with a few beta users. If we don't widely advertise the `/new` page or link it in the nav then only those users who are paying attention to the relevant tickets here on GitHub will know about it. That'll be effectively the same as collecting the information via email and entering it in the db manually. I think it's okay to have a rough user experience here since we're not broadcasting this capability widely yet.

rohitpaulk:
> I think it's okay to have a rough user experience here since we're not broadcasting this capability widely yet.

Ah, okay :)

whit537:
I've been thinking about i18n for this. I don't think we need a full multi-language implementation like we've got for statements, because a team is going to have one primary language they work in. They can just answer the questions in whatever language they want (though that'll pose an interesting challenge in terms of our ability to review new teams). I think it would be good to have a "primary language" field for Teams (which should be pretty easy to implement), but beyond that I don't think we need to worry about i18n.

whit537:
More important is that the form needs to include a checkbox to accept the terms of service.

rohitpaulk:
Rebased on master.

rohitpaulk:
Not the best UX, but it works. Tagging as ready for review...

whit537:
Awesome! I'm working on terms (#3408), and I'll look at this next.

whit537:
We want to require some things of participants before we allow them to register:

 - a working payout route
 - a verified email
 - really, identity verification, but that's too much for now :-(

whit537:
Alright, I'm satisfied with this as far as it goes. We shouldn't merge this until #3408 lands, because the whole point is to have people accept our new terms. Also, anyone using this will need their tips migrated or it's all for naught. Migrating tips depends on the schema that's to be built on #3414. Therefore, I think we need to hold tight with this PR for now. I guess we may as well implement migration on this PR since we're blocked on it.

whit537:
The migration experience should be that if a user creates a team, and that user has `select count(*) from current_tips` > 0, then we should add a notice to the `/new` form that we'll migrate their tips to the new team. And then we should migrate their tips. :-)

whit537:
> And then we should migrate their tips. :-)

We can do that independently of this PR. Unblocking on migration. Still blocked on #3408.

rohitpaulk:
@whit537 - Who'd be the owner of the Gratipay team? `whit537` or `Gratipay`? What happens if other teams apply with their existing 'team account'? Like sudoroom for example, should they log in with `sudoroom` to apply for a team, or should the owner use their personal account?

whit537:
The `~Gratipay` user will be the owner of the `Gratipay` team. Sudo Room should use the `~sudoroom` user account to apply for the `sudoroom` team.

whit537:
Rebased on master post-#3408.

webmaven:
Error messages should link to the form that needs to be completed (in my case, I hadn't verified my email address on the settings form, but I suspect that filling out the identity form also prevented a similar error).

webmaven:
Ah, and now I need to attach a bank account (also still not linking to the appropriate form).

May I suggest that if several requirements are not met, the error message list *all* of them, rather than making us play whack-a-mole?

whit537:
@webmaven Yeah, we cut UX corners in the interest of time, sorry. :-(

whit537:
Reticketed as https://github.com/gratipay/gratipay.com/issues/3422.

Issue #3378, opened 2015-05-01
charge minimum in arrears, not in advance


whit537:
Right now we charge $10 for a $1 tip and trickle it out over another ~9 weeks. We should do like Patreon (which I learned about [under here](https://github.com/gratipay/gratipay.com/issues/3324#issuecomment-94978041)) and let the $1 charge build up for 10 weeks and *then* charge the tipper. This is a little more jittery for the receiver, but a) the jitteriness will even out for any given receiver as ntippers grows, and b) this will remove a source of escrow (#1383).

As part of this, it may make sense to refund all of the <$10 balances that we can. Aaaaaaand if we want to do that, then we better get it done before Balanced shuts down. They're supporting refunds through October 9.

whit537:
Taking this off the Balanced Shutdown milestone. We don't have the hard zero-escrow requirement as we once thought.

whit537:
Erm, sorry. Reopening. :blush:

rorepo:
Thinking of the following approaches:
i) Add a new field to the participants table, something like 'giving_due'. The value to build up by the total of subscriptions each week
ii) Alter the usage of the 'giving' field in participants table to behave as above
iii) Add a new field to the subscriptions table to indicate the last paid date. Use this date to calculate total due
iv) Create a separate record in the subscriptions table for each week, with a new field to indicate 'paid'. In this case, we can calculate the total of unpaid subscriptions based on this field.

In each case, we check the total due, charging when we reach/pass the minimum charge threshold. In (i) and (ii), we reset the value to 0 after charging. In (iii) and (iv), we set the relevant paid date/flag after charging.

(i) and (ii) appear to be the simplest. (ii) might affect current behaviour e.g. will mean relying on a different mechanism to show a participant's current 'giving' value i.e. the current total of all weekly subscriptions
(iii) could cause difficulties if/when a subscriber's subscription values change
(iv) will mean much higher data volumes in the subscriptions table. However, changes in subscription values will not cause problems and this approach also does provide a quick view of subscription history

whit537:
I think we need to track "giving buildup" separately from both individual intentions to give (`subscriptions`) as well as total amount intending to give (`participants.giving`), because we need to keep track of giving buildup *for each team*.

[...]

Issue #1383, opened 2013-09-05
escrow the minimum possible


whit537:
Today's convo on https://github.com/gittip/www.gittip.com/issues/5#issuecomment-23756765, along with "what to do with interest" (#279) and "look into getting a banking license" (#938), make me think that maybe we don't want to escrow any money at all, if we can help it. @zeeg suggested this on [Twitter](https://twitter.com/zeeg/status/344952755611377667) a while back.

The problem with escrowing money is that then we should be doing something with it, namely, earning interest. That's a whole ball of wax in and of itself (#279), and basically turns us into a bank (#938). It's a lot cleaner, both conceptually and legally, if Gittip isn't a bank. (Gittip may still be a money transmitter, but that's different than a bank; see #938 for now, though.)

In order not to hold money in escrow, we would need to require a bank account (or other funding sink) to be attached before we could move money on any recipients behalf. If the deposit fails then we would need a retry-then-refund algorithm, and if that fails then ... eep (that's why I say "as little as possible.").

We need a real EU story (#126) before we can do this. Manual PayPal isn't going to cut it as a funding sink, and without a funding sink and without escrowing we will be seriously retrograde in terms of value for non-U.S. participants.

We need to change the payday algorithm to settle up each week, rather than holding over from week to week. @carsomyr worked out the basics of this on #449.
##

whit537:
Reopening in light of repercussions of the Balanced shutdown. Storing value is running us afoul of Stripe (https://github.com/gratipay/gratipay.com/issues/3245#issuecomment-91112973) and probably MTL in general (#3321).

whit537:
We have three sources of escrow ([IRC](https://botbot.me/freenode/gratipay/2015-04-15/?msg=36705162&page=2)):

 1. **minimum charges**, in order to support micropayments (if you give $1/wk we charge ~$10/9wk)
 2. **giving holdover** (we hold funds for givers-who-receive week to week to cover next week's gifts, in order to avoid processing fees)
 3. **missing/broken withdrawal route** (we allow receivers to pool their money in Gratipay)

whit537:
Looks like we don't need to get this done before Balanced shuts down, but it's still something we want to keep open.

whit537:
We can't entirely eliminate escrow, because of the asynchronous nature of some payment methods. ACH credits can fail and we don't find out for days. We have to keep track of the balance in order to deal properly with such failures.

whit537:
[...]

PR #3414, opened 2015-05-13
rewrite payday

whit537:
Part of #3399.


whit537:
I've printed out payday.py and started combing through it. The reason why https://github.com/gratipay/gratipay.com/issues/3399#issuecomment-101470445 doesn't result in escrow or complicated graph settling is that it ... doesn't do complicated graph settling: **b** can't feed their $1 back into the system. I think we should run with that, so that we can move towards #1383 with this.

whit537:
I'm hoping we can also pick off #3378 while we're here, another part of #1383.

whit537:
I believe since we're not cycling the graph multiple times per payday, we don't need to use the `is_funded` mechanism for distributing funds. We should just be able to use `card_hold_ok` directly.

whit537:
Basically, the work here is to implement the prototype I wrote in Python, in the SQL inside prepare_db. Most of the Python shell around payday should be able to stay intact.

whit537:
I'm starting with the `test_transfer_tips{_whole_graph}` tests, because that calls `settle_tip_graph()`, which is the heart of payday.

whit537:
There some interaction between `participants.balance` and `payday_participants.card_hold_ok` that I need to understand. It relates to #1383 and how we're going to approach that here.

whit537:
On the one hand it seems like we should avoid tackling the arrears issue (#3378) on this PR. On the other hand ...

whit537:
Okay! I believe with that last commit we've gotten rid of the feedback. :microphone: :dizzy_face: *screeeeeeeeeeeeee*

Now to introduce teams ...

whit537:
P.S. We still make payments from preexisting balances, which means we can avoid tackling #3378 here. Phew.

whit537:
Wow. Caching is a :rabbit:-hole. I'm going to stub it out for now. We'll migrate caching of `receiving`, etc., later.

whit537:
I'm also going to not worry about the payroll side this week. Gratipay is the only one of the half-dozen or so teams we're aiming to run with today that uses payroll. We'll just let the full amount clear to our bank account this week. Sorry, Gratipay team. :-/

whit537:
I believe that's it for this PR.

whit537:
Alright, Travis is green. :white_check_mark: 

Here we go! Nothing ventured, nothing gained!

rohitpaulk:
:dancers:

Issue #3416, opened 2015-05-13
Braintree errors not displayed properly


whit537:
https://gratipay.freshdesk.com/helpdesk/tickets/2130

rohitpaulk:
Closed in #3418.

PR #3418, opened 2015-05-13
Display braintree errors properly

rohitpaulk:
#3416

### DIFF
--- a/www/~/%username/routes/associate.json.spt
+++ b/www/~/%username/routes/associate.json.spt
@@ -46,7 +46,7 @@ if change:
         if result.is_success:
             ExchangeRoute.insert(participant, network, result.payment_method.token)
         else:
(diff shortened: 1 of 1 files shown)


rohitpaulk:
@whit537 - Would it make sense to wrap the error in something like "Our payment processor returned to us with: XXX"? 

Looking at [Braintree's docs](https://developers.braintreepayments.com/ios+python/reference/general/processor-responses/authorization-responses#decline-codes), the error message aren't always very user friendly. For example, "Do Not Honor"

whit537:
> Would it make sense to wrap the error

It could, but even "Do Not Honor" is friendlier than a 500. :-)

rohitpaulk:
Hehe, okay - this is ready for review.

PR #3420, opened 2015-05-14
Add subscriptions table

rohitpaulk:
Following #3409 and #3410.

Issue #3415, opened 2015-05-13
run Gratipay 154


whit537:
[153](https://github.com/gratipay/gratipay.com/issues/3385) was the [Gratipocalypse](https://medium.com/gratipay-blog/gratipocalypse-42fd0ec0d9e8). See #3399 for our efforts to get back on track for 154.

whit537:
I believe the payday rewrite is far enough along for this week (#3414).

whit537:
I'm going to migrate `tips` to `subscriptions` on #3399, and then switch over to here.

whit537:
```
=> select count(*) from subscriptions;
┌───────┐
│ count │
├───────┤
│   984 │
└───────┘
(1 row)
```

whit537:
Who's ready?

whit537:
Account review performed manually for beta users of the new system (see #3399).

whit537:
Droplet spun up, backup started ...

rohitpaulk:
:clap: :clap:

whit537:
Droplet updated.

whit537:
Backup finished and (manually) verified.

whit537:
Script started.

whit537:
Seems to be working. :four_leaf_clover:

tshepang:
yeah, just got notification sms

whit537:
`Captured 110 card holds.`

whit537:
@tshepang Nice!

The script seems to have paused ...

whit537:
Could be because we're canceling so many old card holds?

whit537:
I still see the process:

```
$ ps auxww|grep python
whit537   1504  0.0  2.1  45592 10928 ?        S    21:34   0:00 /home/whit537/gratipay.com/env/bin/python2.7 ./env/bin/honcho run -e /dev/stdin ./env/bin/payday
whit537   1524  1.1  7.5 743124 37988 ?        Sl   21:34   0:09 /home/whit537/gratipay.com/env/bin/python2.7 ./env/bin/payday
```

whit537:
Must be stuck in card holds, because the next log line is `Canceled %i card holds.`, and I haven't seen that yet.

whit537:
I guess I'll give it a couple more minutes? What's our option, comment that out and rerun the script?

whit537:
[...]

PR #3424, opened 2015-05-14
Create a migrate tips function

rohitpaulk:
For #3399, includes #3420

### DIFF
--- a/gratipay/models/team.py
+++ b/gratipay/models/team.py
@@ -97,14 +97,18 @@ class Team(Model):
 
             INSERT INTO subscriptions
                         (ctime, mtime, subscriber, team, amount, is_funded)
-                 SELECT ctime
-                      , mtime
-                      , tipper
+                 SELECT ct.ctime
+                      , ct.mtime
+                      , ct.tipper
                       , %(slug)s
-                      , amount
-                      , is_funded
-                   FROM current_tips
-                  WHERE tippee=%(owner)s
+                      , ct.amount
+                      , ct.is_funded
+                   FROM current_tips ct
+                   JOIN participants p ON p.username = tipper
+                  WHERE ct.tippee=%(owner)s
+                    AND p.claimed_time IS NOT NULL
+                    AND p.is_suspicious IS NOT TRUE
+                    AND p.is_closed IS NOT TRUE
 
         """, {'slug': self.slug, 'owner': self.owner})
 
--- a/gratipay/models/team.py
+++ b/gratipay/models/team.py
@@ -87,6 +87,12 @@ class Team(Model):
                 }[self.is_approved]
 
     def migrate_tips(self):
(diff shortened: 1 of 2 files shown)


whit537:
!m @rohitpaulk

whit537:
Once #3415 settles down let's figure out how to merge https://github.com/gratipay/gratipay.com/issues/3399#issuecomment-102175471 and this.

whit537:
Rebased on master.

gratipay/models/team.py
```diff
@@ -54,3 +54,29 @@ def create_new(cls, owner, fields):
...
+        # Make sure the migration hasn't been done already
+        if subscriptions:
+            raise AlreadyMigrated
+
+        self.db.run("""
+
+            INSERT INTO subscriptions
+                        (ctime, mtime, subscriber, team, amount, is_funded)
+                 SELECT ctime
+                      , mtime
```
whit537: I would leave this out and let `mtime` be set to `now`. No?

whit537:
I suppose we want to run this when we approve a team for the first time.

whit537:
How does this compare with the SQL I used the other day?

https://github.com/gratipay/gratipay.com/issues/3399#issuecomment-102175471

gratipay/models/team.py
rohitpaulk: Hmm, I'm not sure. We use `mtime` to show the time the tip was last modified in the UI (Last set). Do we want that to be the time when the user last set it, or the time when we migrated the tip?

rohitpaulk:
> How does this compare with the SQL I used the other day?

That excludes tips from suspicious/closed/unclaimed accounts and also checks for `amount > 0`. Makes sense to exclude those here too.

rohitpaulk:
> I suppose we want to run this when we approve a team for the first time.

Should we also send the tippers (`subscribers`, now) a notification?

whit537:
@rohitpaulk Yes, definitely. Good call.

whit537:
We should go back and notify the new subscribers from last week as well.

rohitpaulk:
Rebased on master.

rohitpaulk:
@whit537 - I'm going to use this for https://github.com/gratipay/gratipay.com/issues/3486#issuecomment-106622666, can you have a look and merge it in?

whit537:
Yes, just reread it, looks good. Go ahead and merge once Travis runs green.

rohitpaulk:
Deploying...

rohitpaulk:
Deployed.

whit537:
Awesome. Now to call it for the users in question?

rohitpaulk:
I tested this on a backup, I can confirm that tips are migrated and an error is raised if the migration is attempted twice.

rohitpaulk:
> Now to call it for the users in question?

Yep. Heading back to #3486...

PR #3425, opened 2015-05-14
Fix gratipay team slug

rohitpaulk:
Handling team slugs better is reticketed as #3426.


whit537:
I think we should use `Gratipay` for the slug. I've updated it manually in the db since we've got deployed code that depends on it. We should allow teams to edit their slug as we do with usernames for participants.

PR #3430, opened 2015-05-15
Fix payment table semantics

rohitpaulk:
https://github.com/gratipay/gratipay.com/pull/3414#commitcomment-11203627

### DIFF
--- a/gratipay/models/__init__.py
+++ b/gratipay/models/__init__.py
@@ -67,7 +67,6 @@ def _check_balances(cursor):
 
     https://github.com/gratipay/gratipay.com/issues/1118
     """
-    return # XXX Bring me back!
     b = cursor.all("""
         select p.username, expected, balance as actual
           from (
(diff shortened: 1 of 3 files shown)


rohitpaulk:
I'd really like to fix `_check_balances` before we retry #3415. I don't have the time to fix it now, but I should be able to within 3-4 hours.

whit537:
I'm ready to merge if you're okay with 46f047b0bdc3b4aa642a4ae6fc8d1c0f03665527, @rohitpaulk.

whit537:
To be clear: this wasn't the reason why we have no `payments`, correct?

whit537:
This bug is unrelated to that, yes?

rohitpaulk:
> this wasn't the reason why we have no payments, correct?
> This bug is unrelated to that, yes?

Yes, and yes. :)

whit537:
k :)

PR #3434, opened 2015-05-15
constraint payments.schema > 0 in the schema

whit537:
This could've gone in #3430.

### DIFF
--- a/sql/branch.sql
+++ b/sql/branch.sql
@@ -1,5 +1,5 @@
 BEGIN;
 
-    ALTER TABLE payments ADD CONSTRAINT positive CHECK (amount > 0) NOT VALID;
+    ALTER TABLE payments ADD CONSTRAINT positive CHECK (amount > 0);
 
 END;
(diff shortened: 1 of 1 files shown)


whit537:
Eh, @rohitpaulk? We were doing the same for `transfers`.

whit537:
We don't. I copied and pasted from `schema.sql` and didn't stop to understand what that meant.

whit537:
> If the constraint is marked NOT VALID, the potentially-lengthy initial check to verify that all rows in the table satisfy the constraint is skipped. The constraint will still be enforced against subsequent inserts or updates (that is, they'll fail unless there is a matching row in the referenced table, in the case of foreign keys; and they'll fail unless the new row matches the specified check constraints). But the database will not assume that the constraint holds for all rows in the table, until it is validated by using the VALIDATE CONSTRAINT option.

http://www.postgresql.org/docs/9.4/static/sql-altertable.html

whit537:
Removed in 8a1a4d2d131c4889ddc28cb5c2e62029fe36ca82. Better, @rohitpaulk?

rohitpaulk:
Purrrfect. :) Let's wait for Travis, and merge this in :dancer:

whit537:
k

PR #3441, opened 2015-05-15
improves debugging

whit537:
Ping @rohitpaulk.

### DIFF
--- a/gratipay/billing/payday.py
+++ b/gratipay/billing/payday.py
@@ -178,13 +178,6 @@ class Payday(object):
                     csv.writer(f).writerows(payments)
                 raise
         self.take_over_balances()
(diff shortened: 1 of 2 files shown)


rohitpaulk:
I don't think that we should merge this into master, can't we [run it on a backup](https://github.com/gratipay/gratipay.com/issues/3415#issuecomment-102496467)?

whit537:
We could run this on a backup, sure. Don't you think we want this change, though? Making the tables permanent makes debugging easier in development and in production.

rohitpaulk:
Ah, makes sense.

rohitpaulk:
Tests passing, and I don't see any typos. Merging.

PR #3449, opened 2015-05-16
Process payroll - part 1

rohitpaulk:
This involves moving the functions from `_mixin_team.py` to the Team model, and fixing tests.


rohitpaulk:
Deleting the functions in `mixin_team` looks like too much of work for one PR, I'm going to implement both side by side (like we did with subscriptions) and slowly migrate everything over.

Issue #3447, opened 2015-05-15
communicate major site changes ... on the site


colans:
I searched all over the Web UI, but couldn't find a way to cancel a weekly tip/payment I'm making.  Is this not possible?  I would expect there to be a widget on the https://gratipay.com/~$USER/giving/ page allowing me to stop a payment.

mattbk:
~~If you go to the page of the person you're tipping, you can edit the amount on the left sidebar.~~

Never mind, that went away.

deltab:
Confirmed. As of a few days ago, the widget's been removed and there's no way to create, change, or stop payments. Also, this week's payments were skipped.

It's been removed (in pull request #3406) as part of Gratipay's change in direction, described in "[Gratipocalypse](https://medium.com/gratipay-blog/gratipocalypse-42fd0ec0d9e8)" on the blog, gratipay/inside.gratipay.com#180, etc.

@whit537, it's rather disconcerting for such a core component of the UI to just vanish, especially one that controls money going out of our bank accounts. Please restore it with suitable changes, or at the very least put in a few words and a link to a longer explanation. That'll save others from trawling through commit logs or raising issues to find out what happened.

anselmh:
I’d add a vote for also adding a separate button stating "Cancel this tip" as it’s confusing for users to edit the amount to $0. I received some questions on that topic from some of my donators.
And yes, a statement/notice that currently no payments are processed would also be good. Couldn’t find any on the website.

colans:
I opened this ticket because I noticed that folks stopped paying me so I better stop paying others if I want to maintain the equilibrium. It looks like there was no intention behind it. If all payments stopped, then I don't have a problem. 

So yes, the real issue here is that after searching through all of the usual pages, I found no notice or information anywhere on what was actually going on. A short message where these controls normally would be should fix this, with a link to more information.

whit537:
Sorry for the confusion, @colans, et al. :-(

I've added +1s to #3432 for turning the tipping/subscription UI back on.

> put in a few words and a link to a longer explanation.
> And yes, a statement/notice that currently no payments are processed would also be good. Couldn’t find any on the website.
> A short message where these controls normally would be should fix this, with a link to more information.

I've PR'd a global site notice in #3452. Is that suitable for now?

whit537:
I've rescoped this ticket to be about communicating major site changes on the site itself, since we already have a ticket for bringing back the tipping UI (#3432).

whit537:
[...]

PR #3452, opened 2015-05-17
add a global notice about site disruption

whit537:
I tried implementing this using our js notification system, but that doesn't support HTML that I can tell. I wanted to include links. The content area is the best place I could quickly find to put this (works properly across the full range of viewport sizes), but that's not available on the unauthenticated homepage, so as a quick hack this PR additionally redirects `/` to `/about/` for anon.

screen shot 2015-05-17 at 6 31 05 pm

Addresses #3447.


whit537:
Fixing tests ...

whit537:
Since this is time-sensitive, I'm planning to merge this in a couple hours if someone else hasn't merged or blocked by then.

Issue #3419, opened 2015-05-14
announce the new system


whit537:
Reticketing from @webmaven at https://github.com/gratipay/gratipay.com/issues/3390#issuecomment-101873253.

mattbk:
Includes updating the front page at all?

ehmatthes:
> Includes updating the front page at all?

It has to; Gratipay needs to pull people in, both project leaders and potential funders.

whit537:
+1 from @kaguillera on #3428:

> I suggest apart from answering this comment/thread/issue directly that it should be emailed/posted/blogged/tweeted/etc somewhere so that users like myself know what is going on.

whit537:
I put up a simple notice in #3452. This ticket is for a proper blog post.

whit537:
Here's a draft: https://medium.com/@Gratipay/2453d3c53077 (still very much in-process).

whit537:
Blocking on:

 - [ ] turning subscriptions on in the UI (#3432)
 - [ ] self-serve PayPal (#2387)
 - [ ] require identity for PayPal (easy enough, but depends on #3458)
 - [ ] split out revenue model question (#3456)

webmaven:
Left some feedback on the draft.

whit537:
Thanks! :)

whit537:
#3458 is too much work to get done before tomorrow. :-( I'm striking it from the punchlist.

whit537:
Ready for final review: https://medium.com/@Gratipay/2453d3c53077.

clone1018:
@whit537 Every team link in that post 404's for me.

whit537:
@clone1018 Yeah, pending #3464&mdash;added to [punchlist](https://github.com/gratipay/gratipay.com/issues/3419#issuecomment-103870467).

whit537:
Outta time. :-(

I'll be back later tonight ...

whit537:
Okay! We have PRs open for #3432 and #3456. Once we land those we can give https://medium.com/@Gratipay/2453d3c53077 a once-over and then publish. Then we can follow up with lots of users who are waiting for this post.

whit537:
We're live!

https://medium.com/gratipay-blog/gratipay-2-0-2453d3c53077

:dancer:

whit537:
#3476 and https://twitter.com/Gratipay/status/601565078547083265

whit537:
I followed up with everyone in our support queue and in my private email.

[...]

PR #3464, opened 2015-05-20
Create Team profile page

kaguillera:
First attempts at creating the Team profile page. Incomplete and not sure it is the direction required.


whit537:
Rough but good enough for #3419. Ping @rohitpaulk @kaguillera.

whit537:
!m @rohitpaulk

PR #3467, opened 2015-05-21
turn on subscriptions

whit537:
Closes #3432.

PR #3468, opened 2015-05-21
convert Giving to Subscriptions

whit537:


whit537:
Nevermind. I broke most of those caching tests in an earlier commit on #3467, not this commit.

Issue #3465, opened 2015-05-21
run Gratipay 155


whit537:
[154](https://github.com/gratipay/gratipay.com/issues/3415)

whit537:
Alright, we want to get #3419 and #3287 out the door today before running this if we can.

whit537:
@rohitpaulk Wanna get a PR going for #3287? I'm going to try to knock out #3419 asap.

whit537:
I've got PRs open for the blockers on #3419. I'm gonna look at #3287.

whit537:
Okay @rohitpaulk, we're getting down to it on this. Do you think we can land #3470 in the next few minutes, or should we bump that to next week?

rohitpaulk:
We should be able to run on Braintree today. The only part I'm not sure about is how we handle cancelling card holds - I can revisit that after today's Payday. 

I'll push the final commits to #3470 in a few mins

whit537:
Okay. We only have 35 minutes if we want payouts to be deposited by tomorrow.

whit537:
I reviewed one new team: https://gratipay.com/mojolicious/. I'm not bothering with our old review system because it doesn't really impact today's payday since we haven't allowed new subscriptions to the new teams yet.

whit537:
I'm going to run locally instead of on a DO droplet since we had problems with DO last week and we don't have much time this week.

rohitpaulk:
How far along are we with the `braintree_customer_id` migration?

whit537:
Not done. We have 20 minutes. I think we should run on Balanced.

rohitpaulk:
> I think we should run on Balanced.

Hmm, yh me too. Braintree today would be too risky.

whit537:
Okay, proceeding on Balanced ...

rohitpaulk:
> I'm going to run locally instead of on a DO droplet since we had problems with DO last week and we don't have much time this week.

The number of API calls this time will be far more (CC charges are included), think it'll be fine?

whit537:
Hmm, good point ... okay, let's start on DO and keep local as a fall-back.

whit537:
Backup taken and verified.

whit537:
Droplet spun up and being updated ...

whit537:
Updated and script started.

whit537:
Fetching card holds ...

whit537:
`HTTPError: 504 Server Error: GATEWAY_TIMEOUT`

[...]

Issue #3391, opened 2015-05-08
migrate cards from Balanced to Braintree


whit537:
To: Balanced, Braintree

> Greetings, teams. What's our process for migrating Gratipay's card vault from Balanced to Braintree?

From: Braintree

> The process is pretty simple. Just reach out to support@balanced.com and let them know you’d like to export to us. We’ll receive a file from them and run the import on our side. The process takes about a week from beginning to end. We cannot process ACH transactions so unfortunately those customers will not be imported. Let me know if there are any questions.

whit537:
From: Balanced:

> Send this encrypted blob to your guys at Braintree.
>
-----BEGIN PGP MESSAGE-----
[...]
-----END PGP MESSAGE-----
>
It contains the data they need to import all your credit cards and bank accounts into Braintree. It's encrypted with their public key.
>
Please note that the URL within expires in 24 hours so if they don't get to this today they'll likely need to reach out for a new URL. They can email the support team directly and we'll assist them.

Forwarded to Braintree!

whit537:
Okay, we've initiated the migration, though we may need to reinitiate it (though Braintree might take that initiative themselves per Marshall's email?).

That means that anyone who updates their card between now and when we start vaulting on Braintree will not have their new info saved on Braintree, unless we make another export. However, anyone who changes their card on Braintree before the second export lands could potentially have their change clobbered.

One option would be to simply disable the credit card form until cards are migrated.

whit537:
From: Braintree
To: Balanced
CC: me

> Hi Balanced Support,
>
Unfortunately I missed the window for this export. May I request it get run again? Thank you.

whit537:
From: Balanced
To: Braintree
CC: Me

>No worries, []. Here's a new link.

whit537:
From: Braintree
To: Balanced
Cc: Me

>Thanks for resending Marshall, we were able to pull down the file successfully. Chad, I have scheduled Gratispay for the first available opening on our import calendar, which is this Thursday (5/14). Let me know if you have any questions!

To: Balanced
Cc (I think?): Braintree

>Awesome, thank you! :D

whit537:
https://gratipay.freshdesk.com/helpdesk/tickets/2121

whit537:
> Thanks for your patience. We are on track for Thursday. We’ll let you know if there are any last-minute issues.

https://gratipay.freshdesk.com/helpdesk/tickets/2129

[...]

Issue #3366, opened 2015-04-27
Citizens for ACH?


whit537:
http://www.citizensbank.com/commercial/automated-clearing-house-ach.aspx
http://www.citizensbank.com/commercial/accessglobal-ach.aspx
https://www.citizensbank.com/commercial-banking/banking-needs/electronic-payments.aspx

whit537:
I called the 877 number and apparently we don't have the right type of account(?). They're having someone else call me back. I think I'm going to stop in the local office to talk with my contact there.

whit537:
Called my contact at Citizens' local office and she's looking into this for me.

whit537:
"There's an underwriting process, it's not an instant sign-up."

whit537:
Local contact has me in the system and I'm waiting to hear back from the next level up.

whit537:
Just had a call with the next level up. She's going to get me a pricing sheet in the next day or so but sounds like they're offering 20¢ per. I did specify that we want to upload NACHA files programmatically and I think she took that on board (good look on recommending we make that explicit, @sbrendtro). I also emphasized that we're only doing credits, and aren't asking to be underwritten for debits at this point (maybe in the future). Seems to be moving along!

sbrendtro:
As you get talking to more technical people at Citizens, you'll want to get a hold of their specifications for file transmission, file confirmations, and their NACHA format guide.  The NACHA format itself is standard, but each bank seems to like to customize the values transmitted in some of the fields.  And the file confirmation format is completely different for each bank.  All of this information will be needed to build an OpenACH bank plugin specific to Citizens.

whit537:
Meeting next Thursday at 10am to discuss.

whit537:
Per https://github.com/gratipay/gratipay.com/issues/3390#issuecomment-100531709, ask Citizens about bank transfers outside the U.S.

sbrendtro:
FYI, international ACH transfers are available through side banks, but to
my knowledge the country lost no longer includes Euro countries since 2009
and the introduction of SEPA payments.  The transfers could likely be done
via:

1. Bank wire (FedWire) - expensive
2. SEPA - needs euro bank relationship
3. OCT transactions to debit cards (direct to card loads) - Citibank does
these and probably others, but not sure who. Read more here:
https://developer.visa.com/vpp/documents/json/Original_Credit_Transaction__OCT_.html

whit537:
[...]

PR #3470, opened 2015-05-21
charge on Braintree

### LINKED ISSUE #3287: migrate to Braintree

whit537:
Closes #3287.


whit537:
I don't see that we had tests for `fetch_card_holds` that actually hit Balanced. I ran this test and it canceled a few dozen card holds in our Balanced sandbox.

rohitpaulk:
@whit537 - I'll take this over, you could focus on using the Balanced -> Braintree migration data. Sound good?

whit537:
@rohitpaulk Roger that.

whit537:
It looks like [customer id](https://developers.braintreepayments.com/javascript+python/reference/request/customer/create#id) is an editable field. The id for customers brought over from Balanced is their Balanced customer id.

whit537:
In other words, we don't need a separate Participant ID in `custom_fields`. We could just use Gratipay's `participant.id` as the customer id at Braintree.

whit537:
Moving to #3391 ...

rohitpaulk:
@whit537 - https://github.com/gratipay/gratipay.com/commit/59ca2f3d7539b9f6ef3e701c3a5754b27063c7ba

whit537:
Good look.

rohitpaulk:
> The transaction has been submitted for settlement and will be included in the next settlement batch. Settlement happens nightly — the exact time depends on the processor.

Looks like we can't capture card holds immediately like with Balanced.

whit537:
> Looks like we can't capture card holds immediately like with Balanced.

Yeah, Balanced floated us the money (i.e., they made it available to us in our escrow immediately, even though they didn't get the money from the acquiring bank until the next day). This came up in my [meeting](https://github.com/gratipay/gratipay.com/issues/3366#issuecomment-102079055) with Citizens last week. It's not a big deal for now since we have enough in escrow to float the different for a day. As we reduce escrow (#1383) we'll have to maybe rethink our timing of card captures vs. payouts.

whit537:
The main point is that card captures are almost certain to settle without incident, so there's little risk in floating for a day if we can make it work cash-flow-wise. Once we have a hold/authorization, we can count on the funds landing in our bank account sooner or later.

rohitpaulk:
> Once we have a hold/authorization, we can count on the funds landing in our bank account sooner or later.

Cool, so I'll proceed with the assumption that a transaction that is `authorized` and then `submitted_for_settlement` is good enough to add to the user's balance.

whit537:
> Cool, so I'll proceed with the assumption that a transaction that is `authorized` and then `submitted_for_settlement` is good enough to add to the user's balance. 

Yup. We'll keep an eye out, and if we end up getting burned by this assumption down the road then we'll adjust.

whit537:
@rohitpaulk One test failure, ya?

rohitpaulk:
@whit537 - Has to do with how we cleanup stuff in the Braintree test sandbox, I'm working on a fix.

rohitpaulk:
Since I was in a hurry, I neglected changing to Braintree's nomenclature (and I probably left a few `import pdb; pdb.set_trace()`s around). We [aren't running on Braintree today](https://github.com/gratipay/gratipay.com/issues/3465#issuecomment-104435747), I'll revisit this tomorrow.

gratipay/billing/exchanges.py
```diff
@@ -183,14 +184,31 @@ def create_card_hold(db, participant, amount):
...
     try:
-        card = thing_from_href('cards', route.address)
-        hold = card.hold( amount=cents
-                        , description=username
-                        , meta=dict(participant_id=participant.id, state='new')
-                         )
-        log(msg + "succeeded.")
-        error = ""
+        result = braintree.Transaction.sale({
+            'amount': str(cents/100.0),
```
whit537: Are we in danger of float errors here? Why type is `cents`?

whit537:
!m @rohitpaulk

gratipay/billing/exchanges.py
rohitpaulk: `cents` is an `int`.

whit537: You positive? I thought it was a `Decimal` here. Regardless, if Braintree accepts dollars we should just pass dollars. We only convert to cents in `_prep_hit` because Balanced wanted cents (though does `ach_credit` also use `_prep_hit`?

rohitpaulk: If I'm reading correctly, `cents` comes from `_prep_hit` - and https://github.com/gratipay/gratipay.com/blob/master/gratipay/billing/exchanges.py#L288

> We only convert to cents in _prep_hit because Balanced wanted cents

True. I wanted to minimize the code changes here, that's why I reused the function. 

> though does ach_credit also use _prep_hit

No, it doesn't.

whit537: > I wanted to minimize the code changes here, that's why I reused the function.

Yeah, we can clean that up later.

whit537: Reticketed as #3500.

PR #3478, opened 2015-05-22
update transition message

whit537:
We can be a little more optimistic now that Gratipay 2.0 is out the door.


whit537:
Ping @rohitpaulk et al.

whit537:
Self-merging, tiny change and time-sensitive.

PR #3483, opened 2015-05-26
allow all authenticated users to see all teams

whit537:
Our team application and approval process calls for a public community comment period, which means that Gratipay users need to be able to see the teams in question.


whit537:
I'm going to fast-track this, @rohitpaulk et al., in light of https://github.com/gratipay/inside.gratipay.com/issues/210. I'll follow up with another PR for other tweaks to team review.

PR #3489, opened 2015-05-27
Strip required input for new team form

### LINKED ISSUE #3479: strip required inputs
I noticed a couple new teams have the revenue model field blank:

https://gratipay.com/sublimelinter/
https://gratipay.com/pageres/

Could be because they applied before we deployed #3469, but I believe it also could be because they put whitespace in the field. We should strip the inputs when validating for required.

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/16727538-strip-required-inputs?utm_campaign=plugin&utm_content=tracker%2F85909&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F85909&utm_medium=issues&utm_source=github).
</bountysource-plugin>

kaguillera:
Here is my attempt to resolve #3479 ping @rohitpaulk, @whit537

### DIFF
--- a/tests/py/test_teams.py
+++ b/tests/py/test_teams.py
@@ -108,6 +108,13 @@ class TestNewTeams(Harness):
         self.make_team(is_approved=False)
         assert 'The A Team' not in self.client.GET("/explore/teams/").body
 
+    def test_stripping_required_inputs(self):
+        self.make_participant('alice', claimed_time='now', email_address='alice@example.com', last_ach_result='')
+        data = dict(self.valid_data)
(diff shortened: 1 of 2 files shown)


whit537:
Ran the test locally without the fix and indeed, it fails.

!m @kaguillera

whit537:
Should've had you apply the `Review` label to this @kaguillera. Another minor detail for next time. :)

Issue #3000, opened 2014-12-10
jstests fail in Vagrant/VirtualBox


techtonik:
Seems like I forgot some `jstest` dependency, but can't figure out which exactly. Travis works ok and that's strange.

```
vagrant@vagrant-ubuntu-trusty-64:/vagrant$ make jstest
node_modules/grunt-cli/bin/grunt test
Running "jshint:gruntfile" (jshint) task
>> 1 file lint free.

Running "jshint:js" (jshint) task
>> 16 files lint free.

Running "jshint:tests" (jshint) task
>> 5 files lint free.

Running "aspen:start" task
Starting Aspen... started.

Running "webdriver:tests" (webdriver) task

events.js:72
        throw er; // Unhandled 'error' event
              ^
Error: spawn ENOENT
    at errnoException (child_process.js:988:11)
    at Process.ChildProcess._handle.onexit (child_process.js:779:34)
make: *** [jstest] Error 8
```

Any ideas what can go wrong? @rummik

rummik:
@techtonik I think this is related to the helper script to let JS tests get a session

techtonik:
@rummik, I am experiencing trouble locating the entrypoint to unwind this problem. Can you point where is the script exactly and how I can execute/debug it manually without Grunt?

rummik:
@techtonik [tests/js/utils/](https://github.com/gratipay/gratipay.com/tree/a02471f4e9573840a730ae0af73c5f9526405d05/tests/js/utils/)

techtonik:
I don't get where should I "require" that session?

```
vagrant@vagrant-ubuntu-trusty-64:/vagrant/tests/js/utils$ node
> var createSession = require('session.js');
Error: Cannot find module 'session.js'
    at Function.Module._resolveFilename (module.js:338:15)
    at Function.Module._load (module.js:280:25)
    at Module.require (module.js:364:17)
    at require (module.js:380:17)
    at repl:1:21
    at REPLServer.self.eval (repl.js:110:21)
    at repl.js:249:20
    at REPLServer.self.eval (repl.js:122:7)
    at Interface.<anonymous> (repl.js:239:12)
    at Interface.EventEmitter.emit (events.js:95:17)
```

techtonik:
Trying to get some debug info:
```
$ node_modules/grunt-cli/bin/grunt -d -v test
...
Starting Aspen... started.

Running "webdriver" task
[...]

PR #3490, opened 2015-05-28
Update grunt-webdriver 0.4.4 -> 0.5.1 for jstest

techtonik:
If it won't break Travis, using latest package
version will make troubleshooting #3000 easier.

Regenerating npm dependency resolution cache:
$ rm npm-shrinkwrap.json
$ npm shrinkwrap --dev

### DIFF
--- a/npm-shrinkwrap.json
+++ b/npm-shrinkwrap.json
@@ -2,25 +2,56 @@
   "name": "gratipay",
   "version": "0.0.0",
   "dependencies": {
+    "ini": {
+      "version": "1.1.0",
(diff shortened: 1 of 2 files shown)


techtonik:
I finally got what `npm-shrinkwrap.json` is, that it should be regenerated to update dependencies, and how to do that. Wheew.

I feel like we need to host quick intro in JS on GP. I'd expect it to be in docs/ but maybe inside.g.c is a better place?

techtonik:
!m whit537

Issue #3481, opened 2015-05-25
release Gratipay 1.0 balances, case-by-case


whit537:
Having announced [Gratipay 2.0](https://medium.com/gratipay-blog/gratipay-2-0-2453d3c53077), we're hearing from some folks that, understandably, just want to cash out and close their Gratipay account. Today our options for them are:

 1. Apply for a team.
 2. Wait until August when we refund all remaining Gratipay 1.0 balances.

Neither of these are awesome. The first is dumb because it's just jumping through hoops. The second is dumb because we're making them wait a long time, and we're forcing them to refund their balance.  Refunding balances is better then letting balances sit in limbo, but obviously less desirable than paying them out.

~~Can we add a disbursement method to account closing that involves accepting a short agreement that the funds were received legitimately?~~ I've made a private ticket to discuss this with @copiesofcopies.

Let's disassociate this from account closing. Per https://github.com/gratipay/gratipay.com/issues/3481#issuecomment-106013128, we can manually review accounts and pay them out if they had been used for a product or service. That will free people up to close their account if they want to, but we don't need to strictly tie this to that.

##### Notify

 - [x] https://gratipay.freshdesk.com/helpdesk/tickets/2103
 - [x] https://gratipay.freshdesk.com/helpdesk/tickets/2147
 - [x] https://gratipay.freshdesk.com/helpdesk/tickets/2161
 - [x] https://gratipay.freshdesk.com/helpdesk/tickets/2187
 - [x] https://gratipay.freshdesk.com/helpdesk/tickets/2232
 - [x] https://github.com/gratipay/inside.gratipay.com/issues/210#issuecomment-106471300
 - [x] https://gratipay.freshdesk.com/helpdesk/tickets/2090
 - [x] https://gratipay.freshdesk.com/helpdesk/tickets/2315

whit537:
If we *can't* figure this out, then I'm not sure how we're able to pay out balances received under Gratipay 1.0 even to people who migrate their tips to Gratipay 2.0 Teams.

kytrinyx:
For (1) does this mean that one would be allowed to donate what is on an individual account to the new team account?

whit537:
@kytrinyx Team accounts don't carry a balance, only ~user accounts do. See https://github.com/gratipay/gratipay.com/issues/3399#issuecomment-101470445 for a rough sketch (bottom of the comment). Sorry it's confusing, we need some better diagrams. :/

whit537:
[...]

PR #3535, opened 2015-06-11
implement a flag for paying out Gratipay 1.0 balances

whit537:
For use on #3481.

### DIFF
--- a/gratipay/billing/payday.py
+++ b/gratipay/billing/payday.py
@@ -422,7 +422,7 @@ class Payday(object):
                         AND t.is_approved IS TRUE
                         AND t.is_closed IS NOT TRUE
                     ) > 0
-                   OR p.status_of_1_0_balance='pending-release')
+                   OR p.status_of_1_0_balance='pending-payout')
 
         """)
         def credit(participant):
--- a/gratipay/billing/payday.py
+++ b/gratipay/billing/payday.py
@@ -416,12 +416,14 @@ class Payday(object):
           ---- Only include team owners
           ---- TODO: Include members on payroll once process_payroll is implemented
 
-               AND ( SELECT count(*)
+               AND (( SELECT count(*)
                        FROM teams t
                       WHERE t.owner = p.username
                         AND t.is_approved IS TRUE
                         AND t.is_closed IS NOT TRUE
-                   ) > 0
+                    ) > 0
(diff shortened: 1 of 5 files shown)


whit537:
Ready for review, @rohitpaulk, et al. We need this for tomorrow's payday.

gratipay/billing/payday.py
```diff
@@ -408,23 +408,22 @@ def payout(self):
...
             SELECT p.*::participants
-              FROM participants p
-             WHERE balance > 0
-               AND ( SELECT count(*)
-                       FROM exchange_routes r
-                      WHERE r.participant = p.id
-                        AND network = 'balanced-ba'
-                   ) > 0
+              FROM exchange_routes r
+              JOIN participants p ON p.id = r.participant
```
rohitpaulk: If a user had 4 old bank accounts and 1 active one, wouldn't this result in multiple calls to `ach_credit` with the same participant? Maybe changing this to `current_exchange_routes` would solve the problem?

whit537:
Blech, forgot the UI for this. :)

gratipay/billing/payday.py
whit537: Does the same bug afflict MassPay?

rohitpaulk: If it does - that's a serious bug, we could end up paying people double their balance :/ Here (ach payouts in payday), the worst case scenario is a few extra API calls.

rohitpaulk: > we could end up paying people double their balance

```SQL
SELECT COUNT(*)
  FROM participants p
 WHERE (
 SELECT COUNT(*)
   FROM exchange_routes r
  WHERE network='paypal'
    AND r.participant = p.id
) > 1; -- returns 0
```
Hasn't happened till now.

rohitpaulk: I think replacing `exchange_routes` with `current_exchange_routes` in both places should solve the problem.

whit537:
Okay @rohitpaulk, how does it look now?

rohitpaulk:
Looks good now.

whit537:
Okay! I'll let you merge when Travis is green. Diving in on #3538 ...

whit537:
Gah, still forgot the UI for this. :)

whit537:
New PR for that: #3542.

PR #3544, opened 2015-06-11
Drop back to iframe for team review

whit537:
It's helpful to see the actual team page, first because we never added the revenue model question to the review dashboard, but also for context like the owner.


whit537:
Ping @rohitpaulk.

Issue #3533, opened 2015-06-10
null route for recent charge receipt


whit537:
``` python-traceback
Traceback (most recent call last):
  File "site-packages/algorithm.py", line 288, in run
    new_state = function(**deps.as_kwargs)
  File "aspen/algorithms/website.py", line 104, in get_response_for_resource
    return {'response': resource.respond(state)}
  File "aspen/resources/simplate.py", line 57, in respond
    exec self.pages[1] in context
  File "/app/www/~/%username/receipts/%exchange_id.int.html.spt", line 24, in <module>
    card = CreditCard.from_route(route)
  File "gratipay/billing/instruments.py", line 32, in from_route
    address_postal_code=card.billing_address.postal_code
AttributeError: 'NoneType' object has no attribute 'postal_code'
```

https://app.getsentry.com/gratipay/gratipay-com/group/67070275/
##

whit537:
I'm trying to provide an example receipt on https://github.com/gratipay/gratipay.com/issues/3531 and hit this. :(

PR #3547, opened 2015-06-11
Check for null billing_address in billing/instruments.py

rohitpaulk:
#3533

I'm not adding extra tests for this, checked that it works by setting `card.billing_address = None` just before where I made the change.


whit537:
Wow, really? That simple?

PR #3561, opened 2015-06-16
teams are under construction

whit537:
Let's tell people that.


techtonik:
No time for Gimp'ing.. And it would be nice to duplicate the workers sign to really mean teams. =)

Issue #2460, opened 2014-06-04
Clarify what DNT means for Gravatar, Twitter, etc.


xnyhps:
https://www.gittip.com/about/dnt.html says:

> Gittip supports DNT. Enabling your browser's DNT settings ensures that Gittip will disable any use of third-party analytics services.

But with DNT enabled Gittip will still try to load images from Twitter and Gravatar (and maybe more I don't know about). Maybe these sites respect DNT (I haven't checked), but the page should mention that they _might_ track you.

I also think it's ironic you only exclude _thid-party_ analytics services, but don't say anything about how you change your own tracking. :)
##

whit537:
Closing in light of our decision to [shut down Gratipay](https://gratipay.news/the-end-cbfba8f50981).

Thank you all for a great run, and I'm sorry it didn't work out! 😞  💃

PR #3563, opened 2015-06-16
Update dnt.spt

kzisme:
Fix for #2460 - Added links to improve clarity for two other services.

### DIFF
--- a/www/about/features/dnt.spt
+++ b/www/about/features/dnt.spt
@@ -24,6 +24,8 @@ subhead = _("Do Not Track")
     ('http://donottrack.us/images/safari_dnt.png', 'Safari'),
     ('https://support.google.com/chrome/answer/114836', 'Chrome'),
     ('http://www.opera.com/help/tutorials/security/control/#referrer', 'Opera'),
+    ('https://support.twitter.com/articles/20169453-twitter-supports-do-not-track', 'Twitter'),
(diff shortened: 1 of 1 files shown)


techtonik:
> So, what would be the best approach the cover all the other social media/analytic services that don't support DNT? Here is a current list that I found: http://donottrack.us/implementations

I'd just list those that we use and that may track users. Just to be clear. On the hand, being too paranoid is also bad. So, need to explain possible consequences and let people decide if tracking is useful for them. If those flags are only relevant to paranoids, then for them they may be useless, because NSA tracks everybody regardless of DHT flags. =)

kzisme:
That is a sensible approach - I just wanted to get some other opinions as well.  

So, aside from the listed services we have listed so far, but are there any other services  that we use that I can l look into a write a paragraph about for some clarifying information?

techtonik:
We need a way to refer to services as data, so that all links to them can be easily found. Perhaps a certain tag for simplates like **service:Twitter** meaning that it is "Twitter used by Gratipay" and not just Twitter in general.

techtonik:
With such tags in place, we can just grep the sources and get list of all the services to revise.

whit537:
We call these "`elsewhere`s" (see the db schema).

techtonik:
@whit537 The Gravatar falls out of this category, no?

whit537:
Correct.

PR #3573, opened 2015-06-19
Fake teams

rorepo:
Changes made to gratipay/utils/fake_data.py to populate the new 'teams' table.

This is a work in progress. Intend to improve on some of the hardcoded values and also include subscriptions.

Issue #3566, opened 2015-06-18
run Gratipay 159


whit537:
[158](https://github.com/gratipay/gratipay.com/issues/3540)

whit537:
Last day for Balanced ACH.

whit537:
0 users to review! O.O

Reviewing teams on https://github.com/gratipay/inside.gratipay.com/issues/248 ...

whit537:
Approved the teams I was able to on https://github.com/gratipay/inside.gratipay.com/issues/248. Will send out notifications to everyone once payday is done ...

whit537:
```sh
[gratipay] $ heroku config -s | run_dammit /dev/stdin -c ./env/bin/python bin/migrate-tips.py
[...]
'bevry' already migrated.
Migrated tips for 'shields'
Migrated tips for 'kivy'
'gammu' already migrated.
Migrated tips for 'knopproject'
'the-charis-project' already migrated.
'northern-plains-athletics' already migrated.
Migrated tips for 'libravatar'
'debian-lts-by-freexian' already migrated.
Migrated tips for 'doit'
'the-mountain-spirit' already migrated.
'jshttp' already migrated.
Migrated tips for 'aegirproject'
'survivejs' already migrated.
Migrated tips for 'nanoc'
Migrated tips for 'ucoin'
Done.
```

whit537:
Script ran for a minute (0:01:54.136485).

whit537:
Fare thee well, @balanced! :rocket: :guitar: :parking: :rice_ball:

whit537:
Declines look good.

```
[gratipay] $ cat gratipay-159.log | grep "Holding.*failed" | wc -l
       4
[gratipay] $ cat gratipay-159.log | grep "Captured.*cents" | wc -l
      75
[gratipay] $
```

![screen shot 2015-06-18 at 6 18 29 pm](https://cloud.githubusercontent.com/assets/134455/8243619/a07f61a0-15e6-11e5-9068-6805afcb05a9.png)

whit537:
MassPay fee cap error.

whit537:
Fee cap error fixed.

MassPay done and posted back for 20 users.

whit537:
Escrow shuffled from New Alliance to PayPal. Emptying our Balanced escrow reticketed as https://github.com/gratipay/inside.gratipay.com/issues/266.

whit537:
[...]

PR #3582, opened 2015-06-24
fake-data, teams and subscriptions

rorepo:
Fake data script updated to create teams and subscriptions.

Fake data test script updated accordingly.

### DIFF
--- a/gratipay/utils/fake_data.py
+++ b/gratipay/utils/fake_data.py
@@ -10,6 +10,7 @@ from psycopg2 import IntegrityError
 from gratipay import wireup, MAX_TIP, MIN_TIP
 from gratipay.elsewhere import PLATFORMS
 from gratipay.models.participant import Participant
+from gratipay.models.team import Team
 from gratipay.models import community
 from gratipay.models import check_db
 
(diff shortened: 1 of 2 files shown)


whit537:
Thanks for the contribution, @rorepo! :dancer:

rorepo:
Following on from the comment in the earlier #3573:

Leaving fake tips around until the migration has been verified seems to make sense. However, how is the migration intended to happen? Is it that new users will only be using subscriptions, while existing users will have their tips records migrated to subscriptions? Presumably, starting with a small batch of existing users?

On removal of tips from fake_data, this is the situation as I see it:
Methods:
fake_tip_amount - this is a generic function to give a dummy amount, the only relation to tips being the MIN_TIP and MAX_TIP values. I've used this function in fake_subscription, so this function itself can't be removed unless a new version is created for fake-subscription.

fake_tip - can be removed, subject to the 'Making tips' bit being removed from populate_db

populate_db - in this:
The 'Making tips' bit can be removed, subject to the usage of the tips collection being removed/replaced in the rest of the code
A small complication I see in doing this is that I'm unclear about the equivalence of tipper, tippee in tips to subscriber, team in subscriptions. If I have understood correct, both tipper and tippee were referring to participant instances, while in subscriptions, the team is obviously referring to a team. And the complication actually arises from the fact that tips.tippee refers to the participant username, while subscriptions.team refers to team.slug - so this will need to be worked around.

Additionally, there is the aspect of transfers - are these still from participant to participant?

Also, a question about methods like fake_tip_amount, (community) slugize - seems likely these (or versions of these) could be used by different modules. Is there a common module such methods could be added to?

whit537:
> However, how is the migration intended to happen?

Migration happens with [a one-off script](https://github.com/gratipay/gratipay.com/blob/1858/bin/migrate-tips.py) every week during payday, e.g., https://github.com/gratipay/gratipay.com/issues/3566#issuecomment-113302914.

whit537:
> Additionally, there is the aspect of transfers - are these still from participant to participant?

Yes, and, as with `tips`, these are historical. We replaced the `tips` and `transfers` tables with the `subscriptions` and `payments` tables in #3414.

whit537:
> Also, a question about methods like fake_tip_amount, (community) slugize - seems likely these (or versions of these) could be used by different modules. Is there a common module such methods could be added to?

Maybe in `gratipay.utils`?

PR #3588, opened 2015-06-30
upgrade Aspen to 0.39+

whit537:
0.39 has the changes we're interested in. Here's the Changelog:

https://github.com/gratipay/aspen-python/blob/master/Changelog.md#040---released-mon-jun-29-2015-by-whit537

PR #3593, opened 2015-07-02
Fix compile_assets for base_url

whit537:
Fixes a regression from #3588.


whit537:
Merging to get master back on track. We are rolled back at Heroku currently.

whit537:
Actually, waiting for Travis ...

PR #3596, opened 2015-07-02
Fake data - subscriptions and payments

rorepo:
Added fake subscriptions and payments. Tests have been correspondingly updated.

Tips and transfers have been left as they are, for removal later once migration is complete.

This is ready for review.

Please note one change in particular that needs to be validated, made for week_tips and week_transfers (and the new week_subscriptions and week_payments):
week_tips = filter(lambda x: date <= x['ctime'] < end_date, tips)

  was

week_tips = filter(lambda x: date < x['ctime'] < end_date, tips)

Seemed to me that transactions at min_date were being skipped in processing due to the 'date < x['ctime']' condition. This wasn't causing any issues, because this would correspondingly not have gone into exchanges as well.

### DIFF
--- a/gratipay/utils/fake_data.py
+++ b/gratipay/utils/fake_data.py
@@ -394,12 +394,12 @@ def populate_db(db, num_participants=100, ntips=200, num_teams=5, num_transfers=
         participant = subscription['subscriber']
         team = Team.from_slug(subscription['team'])
         amount = subscription['amount']
-        if participant != team.owner:
-            paymentcount += 1
-            sys.stdout.write("\rMaking Payments (%i)" % (paymentcount))
-            sys.stdout.flush()
-            payments.append(fake_payment(db, participant, team.slug, amount, 'to-team'))
-            teamamounts[team.slug] = teamamounts[team.slug] + amount
+        assert participant != team.owner
+        paymentcount += 1
+        sys.stdout.write("\rMaking Payments (%i)" % (paymentcount))
+        sys.stdout.flush()
+        payments.append(fake_payment(db, participant, team.slug, amount, 'to-team'))
+        teamamounts[team.slug] = teamamounts[team.slug] + amount
     for team in teams:
         paymentcount += 1
         sys.stdout.write("\rMaking Payments (%i)" % (paymentcount))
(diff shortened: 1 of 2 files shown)


gratipay/utils/fake_data.py
```diff
@@ -167,6 +167,22 @@ def fake_elsewhere(db, participant, platform):
...
                 )
 
+def fake_payment(db, participant, team):
+    """Create fake payment
+    """
+    direction = ['to-team','to-participant']
+    if participant.username == team.owner:
+        direction = 'to-participant'
+    else:
+        direction = random.sample(['to-team','to-participant'],1)[0]
```
rohitpaulk: Since we don't have takes yet - If the participant isn't an owner, the direction will always be `to-team`.

rohitpaulk: Also, since teams don't carry a balance - we need the net payments to sum up to zero. I've added a db check for this in https://github.com/gratipay/gratipay.com/pull/3597 :) Maybe try rebasing on that to make sure?

rorepo:
The requirement of payments adding up to zero for teams is unlikely to be met with the random values being created at the moment. Therefore, this will require the following approach:
- Create 'to-team' payments corresponding to all subscriptions
- Create 'to-team' payments for each distinct team subscribed to in subscriptions, equal in value to the total of all subscriptions received

This will however have the effect of limiting the number of payments to no. of participants * no. of teams.

rorepo:
Logic modified as above.

The number of subscriptions is also now dependent on the number of participants and teams, unless we are to go for multiple subscriptions from the same participant to the same team.
With current defaults, now there are 495 subscriptions and 500 payments in fake data.

whit537:
@rorepo Quick note re: process: Please apply the "Review" label to PRs that are ready for review. (I've updated http://inside.gratipay.com/howto/label-github-issues; waiting for the change to propagate.)

rohitpaulk:
The teams are created, but I can't see them on the Explore page... 

screenshot from 2015-07-07 20 30 53

rohitpaulk:
Oops, sorry - My bad. A server restart fixed it.

rohitpaulk:
Rebased on master.

rohitpaulk:
@techtonik - Any idea what's up with [the tests](https://travis-ci.org/gratipay/gratipay.com/builds/69911732)?

`make: env/bin/honcho: Command not found`

rohitpaulk:
screenshot from 2015-07-07 21 02 30

How do participants end up with that much of a negative balance?

rohitpaulk:
Hmm, still can't figure out the negative balances.

rohitpaulk:
Looks lik the issue is present on master too. This PR is definitely an improvement over what's in master, so I'd merge it in once we can figure out why the tests aren't passing. (https://github.com/gratipay/gratipay.com/pull/3596#issuecomment-119238676)

whit537:
> Any idea what's up with the tests?

Weird. This PR doesn't even modify `Makefile`, just two test files. Can we pin down which file has the changes that introduce the bug? What happens if we temporarily back out the changes to `test_fake_data.py`? Do we still see the missing honcho?

whit537:
Is it something to do with build caching at Travis? I've restarted [the build](https://travis-ci.org/gratipay/gratipay.com/builds/69931790) ...

whit537:
It's odd that we get [requirement already satisfied](https://travis-ci.org/gratipay/gratipay.com/builds/69931790#L149) for honcho. How does that output of `make env` compare to a passing test run?

whit537:
(Restarted build failed, ftr.)

rorepo:
I've figured out the negative balance issue. This is down to participants corresponding to week_payments not having ctime < the end_date for the processing week. These payment values therefore don't get processed and thus cancelled out through exchanges. Happens because participant ctime is also random. Believe this situation would only exist in the dev environment.

Removing the ctime < end_date filter for week_participants (i.e. using participants collection as such) resolves the negative balances - except for one bit. The fee value from exchanges still remains. This is because this value has nothing in payments to cancel it out. Not sure how to handle this.

Please note the above is solely about payments, however. Residual balances will also be a problem as long as transfers remain, as now, because these are still random values. I suppressed transfers while figuring out the above.

whit537:
What's up with this PR, @rorepo @rohitpaulk? What do we need to do to bring this in for a landing?

rohitpaulk:
As far as I remember, the tests were passing on my local. Haven't been able to figure out what's up with Travis

whit537:
Willing to pick up with this again, @rorepo? It's been hanging open for a while. I've restarted the latest Travis build in case it was a spurious error (missing `honcho`!?), but at this point we'll need to rebase on master and resolve conflicts before we can land this.

kaguillera:
oh ok...I need to keep that in mind

kaguillera:
rebased on master. next step is to see what we keep and what we remove. 😱

kaguillera:
Now to remove communities from `fake_data.py` since we no longer use that.

kaguillera:
I think that is good now...could someone review this. @rohitpaulk @aandis @whit537 etc.

whit537:
!m @kaguillera

PR #3616, opened 2015-07-13
update history page

whit537:
Fixes #3455.

Issue #3539, opened 2015-06-11
refund Gratipay 1.0 balances


whit537:
In [Gratipay 2.0](https://medium.com/gratipay-blog/gratipay-2-0-2453d3c53077) we state the following:

> You have until August 19, 2015 (90 days) to migrate your incoming payments from your Gratipay user account to a new Gratipay Team; this will also pay out your current balance. On that date, we intend to clear all remaining payments to Gratipay users, and to clear the balance of any user that is not an owner or member of a team, by refunding the money back to the sources it originally came from.

Since that was published, we've added a way to [cash out 1.0 balances](https://github.com/gratipay/gratipay.com/issues/3481), so that's another option folks have before the big refund. We've also pushed the date back to October 1.

#### Notify
- [ ] https://gratipay.freshdesk.com/helpdesk/tickets/3103

whit537:
We should notify affected people about this 30 days out, on Monday, July 20.

whit537:
##### Draft 1

> You are receiving this message because you are carrying a balance of $`__`.`__` on Gratipay.
>
> Three months ago, prompted by legal concerns with our old terms of service, we wrote new terms of service and relaunched as Gratipay 2.0. As mentioned in [the 2.0 announcement](https://medium.com/gratipay-blog/gratipay-2-0-2453d3c53077), we intend to clear out remaining Gratipay 1.0 balances by sending funds back to the original donors to the extent possible. The date for this will be August 20, thirty days from today.
>
If you would like to withdraw your Gratipay 1.0 balance rather than having it refunded to your original donors, then you have two options:
> 
1. Apply for [a new Gratipay Team]().
2. Apply for [a one-time payout]().
>
In either case, if we approve your application, we will pay out your balance during the next weekly payment cycle. There are five more payment cycles from today before we refund remaining balances. Whatever we are unable to refund to the original donor, we will donate to Watsi.

whit537:
August 19 is a Wednesday. Let's push back a day to August 20, to give people that extra payment cycle to get in on payouts.

whit537:
That would put 30 days out at Tuesday, July 21. We should send a second email a week out, on August 13. And of course we should send a final email when it's done.

whit537:
Here's what we should get done before we send out the 30 day email:

 - Make team applications suck less.
 - Implement a one-time payout application.

whit537:
Should we try and pull of #3378 as part of this?

whit537:
[...]

PR #3619, opened 2015-07-16
rip out balanced-ba references

whit537:
I left the ones in test_history because #3616 takes care of those. Note that I also left `balanced-ba` in the schema and the database, because we still want a record of those routes in the db. We just don't want to
do anything with them anymore.


whit537:
Ready for review @rorepo @kzisme @rohitpaulk et al.

whit537:
I guess let's leave `balanced-cc` alone until after https://github.com/gratipay/gratipay.com/issues/3539.

whit537:
Self-merging ...

PR #3621, opened 2015-07-16
Tweak revenue prompt

whit537:
Based on applications that have come in, I think this is the more essential question we should be asking.


whit537:
S'okay, @mattbk @kzisme et al.?

mattbk:
It's not the same question, but I like it better.

How about
 `<p>{{ _("How do you prompt your users to financially support your team's work?") }}</p>`

whit537:
@mattbk Done, and changed "users" to "customers."

Issue #3475, opened 2015-05-21
stop holding balance to cover next week's payments


whit537:
With the new payroll algorithm (#3415), we can finally stop doing this.

whit537:
This is one element of #1383.

whit537:
Done in #3629.

PR #3629, opened 2015-07-21
Stop holding balance to cover next week's payments

rorepo:
Changes to address issue:
stop holding balance to cover next week's payments #3475


whit537:
Thanks for jumping right on this, @rorepo. Seems a simple enough fix. :-)

Especially now that MassPay is our *only* payout mechanism, we should move the `masspay.py` code under `gratipay.billing` and write some tests for it. We should install the cli as we do with payday itself—in fact, we may want to fold masspay into payday proper. We don't need to block on that here, so I've reticketed it as #3634.

PR #3636, opened 2015-07-22
redirect Security to our HackerOne profile

whit537:
Part of https://github.com/gratipay/inside.gratipay.com/issues/255.


whit537:
Self-merging.

PR #3639, opened 2015-07-23
here's a couple best-practice security headers

whit537:
Ping @rohitpaulk @rorepo @kzisme @techtonik et al.


whit537:
Also ping @greggles @benhc123. :-)

Issue #3645, opened 2015-07-26
"Connect a bank account" still listed in FAQ


mattbk:
Also, PayPal adding is wrong.  https://gratipay.com/about/faq

>How do I withdraw funds from Gratipay? 
>There are two ways to withdraw funds from Gratipay:
>Connect a bank account. Funds will be deposited every Friday ($10 minimum; no fee; U.S.-only).
>Email us to set up a weekly PayPal payout (unlimited payout; 2% fee capped at $20).

PR #3646, opened 2015-07-26
Update payout/withdrawal options

mattbk:
No more direct deposits, only PayPal.  https://github.com/gratipay/gratipay.com/issues/3645


whit537:
!m @mattbk

Issue #3640, opened 2015-07-23
"Account Type" no longer meaningful?


mattbk:
With Gratipay 2.0, "Account Type" on the Settings page doesn't mean anything anymore, does it?  This is confusing now that we have old teams and new teams.
![screenshot_072315_110127_am](https://cloud.githubusercontent.com/assets/3729038/8855403/978c7ec0-312a-11e5-96cf-37de9e2b75fc.jpg)

Suggest it be removed from the UI.

mattbk:
Created a PR because I could.  https://github.com/gratipay/gratipay.com/pull/3647

whit537:
!m @mattbk

PR #3647, opened 2015-07-27
Remove UI for individual vs team

mattbk:
https://github.com/gratipay/gratipay.com/issues/3640


whit537:
I did a brief skim of the codebase for `SINGULAR` and `PLURAL`, I think this is okay. I've reticketed removing all traces of this as #3648.

PR #3652, opened 2015-07-30
rename `subscriptions` to `payment_instructions`

whit537:
https://github.com/gratipay/inside.gratipay.com/issues/117


whit537:
Ping @rorepo @rohitpaulk @techtonik @kzisme et al.

whit537:
Ping @techtonik @tshepang @colindean @mattbk for conceptual review.

tshepang:
see this Comment https://github.com/gratipay/inside.gratipay.com/issues/117#issuecomment-126223270

whit537:
> [A]dding *_instructions* does not help anyone understand these things better, and makes for overlong names.

`instructions` is the base, and `payment_`/`payroll_` is the modifier, rather than the other way around. If we wanted to shorten the table name we could combine them into a single `instructions` table, with an `ENUM('payment', 'payroll')` column. Payroll instructions will include tax withholding instructions, which could be `NULL` where `instruction_type` is `payment`. That'd be similar to [how we do `exchange_routes`](https://github.com/gratipay/gratipay.com/blob/1870/sql/schema.sql#L172-L184), where we have a `fee_cap` that only `paypal` uses.

whit537:
I'm starting to code up that approach, and it feels to me like it brings payments and payroll too close together. We don't want to accidentally conflate them, that'd be bad bugs. I think we should keep them as separate tables.

whit537:
I'm gonna self-merge this pretty soon ...

mattbk:
I trust you know what you're doing. 

On July 30, 2015 8:25:36 AM CDT, Chad Whitacre <notifications@github.com> wrote:
>I'm gonna self-merge this pretty soon ...
>
>---
>Reply to this email directly or view it on GitHub:
>https://github.com/gratipay/gratipay.com/pull/3652#issuecomment-126323501

-- 
Sent from a phone that, although intelligent, is not street smart.

whit537:
I much preferred when we had people around to code review. I think we'll get back there again, but I don't think we can wait until then to make progress. :-(

whit537:
/me takes a backup before deploying ...

whit537:
/me remembers that it's payday ... :)

whit537:
https://github.com/gratipay/gratipay.com/issues/3656 :)

whit537:
Deployed!

Issue #3511, opened 2015-06-01
update subscriptions page


akarelas:
My account (akarelas) sees the following screen on gratipay. As you can see the numbers don't add up:

http://s3.postimg.org/g77tvv9eb/gratipay.png

##### Notify:

 - [ ] https://gratipay.freshdesk.com/helpdesk/tickets/2233
 - [ ] https://gratipay.freshdesk.com/helpdesk/tickets/2398
 - [ ] https://gratipay.freshdesk.com/helpdesk/tickets/2399
 - [ ] https://gratipay.freshdesk.com/helpdesk/tickets/2408
 - [ ] https://gratipay.freshdesk.com/helpdesk/tickets/2405
 - [ ] https://gratipay.freshdesk.com/helpdesk/tickets/2483
 - [ ] https://gratipay.freshdesk.com/helpdesk/tickets/2540

whit537:
Thanks for the report, @akarelas, sorry for the trouble.

whit537:
+1 https://gratipay.freshdesk.com/helpdesk/tickets/2233

> BTW, just noticed another issue: total sum of my payments (in "Giving" on dashboard or under my nickname in the header) is wrong. It seems to include pre-2.0 payments, but not current ones.

whit537:
@timothyfcook on #3570:

>It says I'm giving $10.25 a week, but I see only see one item for $2.00 in my Subscriptions detail.
>
Is this an issue with the old recipients of my gifts not having activated/reconciled their accounts with the Gratipay 2.0 restrictions?
>
![screen shot 2015-06-19 at 11 28 58 am](https://cloud.githubusercontent.com/assets/1823300/8256669/7ddde56e-1676-11e5-9792-cc6106981073.png)

whit537:
+1 from @noffle at https://github.com/gratipay/gratipay.com/issues/3572:

> On my account page it claims "You pay $2.10 per week", though under Subscriptions all I see is a $0.10 / week subscription to Gratipay.
>
According to my history, my $2.10 / week payments stopped going through as of April 30 2015. My credit card information looks up to date.

whit537:
+1 from @joelhooks on #3589:

>http://cl.ly/image/0H431M3L250i
>
It says I am paying, but nothing shows up. What's up with that? I can't affect donations at all. This seems totally broken on my account.
>
> I can't add new payments or see my existing ones, so I removed my card altogether. This makes me sad for the people I was contributing too...
>
[...]

PR #3663, opened 2015-07-31
fix giving amounts in the UI

whit537:
#3599 is a :rabbit2: hole. Here's a quick fix that drops back to computed rather than cached values.


whit537:
This PR resolves the worst error in the UI: #3511. There's still a ton of bad but not egregiously horrible bugs left for #3599 to fix.

whit537:
Self-merging promptly because this is fairly self-contained and Travis is green.

Issue #3579, opened 2015-06-23
occasional spurious test failure


whit537:
```
================================================== FAILURES ==================================================
______________________________ TestNewTeams.test_migrate_tips_to_subscriptions _______________________________
Traceback (most recent call last):
  File "/Users/whit537/personal/gratipay/gratipay.com/tests/py/test_teams.py", line 133, in test_migrate_tips_to_subscriptions
    assert subscriptions[0].subscriber == 'alice'
AssertionError: assert 'bob' == 'alice'
  - bob
  + alice
```

whit537:
Fixed in #3695.

PR #3568, opened 2015-06-19
streamline team review

whit537:
Our team review queue is quite frictive at the moment. Here's what I propose:

 - ~~As part of payday, we create a new milestone every week for the week following.~~
 - Applying for a team automatically creates a new GitHub issue ([`team-review` repo](https://github.com/gratipay/team-review/issues)~~`inside.gratipay.com` repo, `Review` label~~~~, the latest milestone~~)
 - We revise the [content we've been sending in emails](https://github.com/gratipay/inside.gratipay.com/issues/248#issuecomment-111585170), and, rather than sending it in an email, simply put it in the page we display to the user after POSTing to `teams/create.json`.
 - The notification email we send on approve/reject (#3564) should link back to their review issue.

PR #3675, opened 2015-08-09
Charge in arrears

rorepo:
Implemented the core part of the solution for issue "charge minimum in arrears, not in advance #3378" 
Overview of the solution:
i) Added a new column giving_due in payment_instructions. To be populated/incremented with the corresponding amount, each time the amount is not charged due to being under the minimum charge threshold
ii) After creation of payday_payment_instructions, the giving_due value for each participant-team combination is updated with the sum of giving_due values for that combination, from payment_instructions. 
iii) The 'giving_today' value for each participant is calculated as sum(amount + giving_due) from all records for that participant in payday_payment_instructions
iv) The 'giving_today' value is checked against the MINIMUM_CHARGE threshold before creating card holds if at/above the threshold. Have a doubt here - whether it's the basic amount or the upcharged amount that is to be used in the check
v) In the case of participants who have been charged, all giving_due values in payment_instructions for those participants are reset to 0
vi) In the case of participants who have not been charged, the giving_due value in the current record in payment_instructions is updated to amount + giving_due

Will await comments/suggestions before proceeding further. 

Further changes planned:
i) Add a new column giving_due in participants, to cache the total giving_due value for participants
ii) Update the UI to include the above

### DIFF
--- a/gratipay/billing/payday.py
+++ b/gratipay/billing/payday.py
@@ -19,7 +19,7 @@ import braintree
 import aspen.utils
 from aspen import log
 from gratipay.billing.exchanges import (
-    cancel_card_hold, capture_card_hold, create_card_hold, upcharge,
+    cancel_card_hold, capture_card_hold, create_card_hold, upcharge, MINIMUM_CHARGE,
 )
 from gratipay.exceptions import NegativeBalance
 from gratipay.models import check_db
@@ -220,12 +220,17 @@ class Payday(object):
             if p.old_balance < 0:
                 amount -= p.old_balance
             if p.id in holds:
-                charge_amount = upcharge(amount)[0]
-                if holds[p.id].amount >= charge_amount:
-                    return
+                if amount >= MINIMUM_CHARGE:
+                    charge_amount = upcharge(amount)[0]
+                    if holds[p.id].amount >= charge_amount:
+                        return
+                    else:
+                        # The amount is too low, cancel the hold and make a new one
+                        cancel_card_hold(holds.pop(p.id))
                 else:
-                    # The amount is too low, cancel the hold and make a new one
+                    # not up to minimum charge level. cancel the hold
                     cancel_card_hold(holds.pop(p.id))
+                    return
             hold, error = create_card_hold(self.db, p, amount)
             if error:
                 return 1
@@ -253,6 +258,26 @@ class Payday(object):
         log("Processing payment instructions.")
         cursor.run("UPDATE payday_payment_instructions SET is_funded=true;")
 
+    @staticmethod
+    def park_payment_instructions(cursor):
+        """In the case of participants in whose case the amount to be charged to their cc's
+        in order to meet all outstanding and current subscriptions does not reach the minimum
+        charge threshold, park this for the next week by adding the current subscription amount 
+        to giving_due
+        """
+	cursor.run("""INSERT INTO participants_payments_uncharged 
+              SELECT participant 
+                FROM payday_payment_instructions
+            GROUP BY participant
+              HAVING SUM(amount + giving_due) < %(MINIMUM_CHARGE)s
+            """,dict(MINIMUM_CHARGE=MINIMUM_CHARGE))
+
+        cursor.run("""UPDATE payment_instructions 
+              SET giving_due = amount + giving_due
+            WHERE id IN (SELECT id 
+                FROM payday_payment_instructions ppi, participants_payments_uncharged ppu
+               WHERE ppi.participant = ppu.participant)
+                """)
 
     @staticmethod
     def transfer_takes(cursor, ts_start):
@@ -295,7 +320,7 @@ class Payday(object):
             SELECT *
               FROM payday_participants
              WHERE new_balance < 0
-        """)
+        """,dict(MINIMUM_CHARGE=MINIMUM_CHARGE))
         participants = [p for p in participants if p.id in holds]
 
         log("Capturing card holds.")
@@ -340,6 +365,13 @@ class Payday(object):
             INSERT INTO payments (timestamp, participant, team, amount, direction, payday)
                 SELECT *, (SELECT id FROM paydays WHERE extract(year from ts_end) = 1970)
                   FROM payday_payments;
+        """)
+        # Copy giving_due value back to payment_instructions
+        cursor.run("""
+            UPDATE payment_instructions pi
+               SET giving_due = pi2.giving_due
+              FROM payday_payment_instructions pi2
+             WHERE pi.id = pi2.id
         """)
         log("Updated the balances of %i participants." % len(participants))
 
--- a/gratipay/billing/payday.py
+++ b/gratipay/billing/payday.py
@@ -330,7 +330,7 @@ class Payday(object):
             SELECT *
               FROM payday_participants
              WHERE new_balance < 0
-        """,dict(MINIMUM_CHARGE=MINIMUM_CHARGE))
+        """)
         participants = [p for p in participants if p.id in holds]
 
         log("Capturing card holds.")
--- a/gratipay/billing/payday.py
+++ b/gratipay/billing/payday.py
@@ -258,26 +258,36 @@ class Payday(object):
         log("Processing payment instructions.")
         cursor.run("UPDATE payday_payment_instructions SET is_funded=true;")
 
+
     @staticmethod
     def park_payment_instructions(cursor):
         """In the case of participants in whose case the amount to be charged to their cc's
         in order to meet all outstanding and current subscriptions does not reach the minimum
-        charge threshold, park this for the next week by adding the current subscription amount 
+        charge threshold, park this for the next week by adding the current subscription amount
         to giving_due
         """
-	cursor.run("""INSERT INTO participants_payments_uncharged 
-              SELECT participant 
-                FROM payday_payment_instructions
-            GROUP BY participant
-              HAVING SUM(amount + giving_due) < %(MINIMUM_CHARGE)s
-            """,dict(MINIMUM_CHARGE=MINIMUM_CHARGE))
+        cursor.run("""
 
-        cursor.run("""UPDATE payment_instructions 
-              SET giving_due = amount + giving_due
-            WHERE id IN (SELECT id 
-                FROM payday_payment_instructions ppi, participants_payments_uncharged ppu
-               WHERE ppi.participant = ppu.participant)
-                """)
+        INSERT INTO participants_payments_uncharged
+             SELECT participant
+               FROM payday_payment_instructions
+           GROUP BY participant
+             HAVING SUM(amount + giving_due) < %(MINIMUM_CHARGE)s
+
+        """, dict(MINIMUM_CHARGE=MINIMUM_CHARGE))
+
+        cursor.run("""
+
+        UPDATE payment_instructions
+           SET giving_due = amount + giving_due
+         WHERE id IN ( SELECT id
+                         FROM payday_payment_instructions ppi
+                            , participants_payments_uncharged ppu
+                        WHERE ppi.participant = ppu.participant
+                      )
+
+        """)
+
 
     @staticmethod
     def transfer_takes(cursor, ts_start):
--- a/gratipay/billing/payday.py
+++ b/gratipay/billing/payday.py
@@ -351,8 +351,16 @@ class Payday(object):
         cursor.run("""
             UPDATE payment_instructions pi
                SET due = pi2.due
-              FROM payday_payment_instructions pi2
+              FROM payment_instructions_due pi2
              WHERE pi.id = pi2.id
+        """)
+        # Reset older due values to zero
+        cursor.run("""
+            UPDATE payment_instructions pi
+               SET due = '0'
+             WHERE pi.id NOT IN (
+                 SELECT id
+                   FROM payment_instructions_due pi2)
         """)
 
         log("Updated the balances of %i participants." % len(participants))
--- a/gratipay/billing/payday.py
+++ b/gratipay/billing/payday.py
@@ -347,21 +347,6 @@ class Payday(object):
                 SELECT *, (SELECT id FROM paydays WHERE extract(year from ts_end) = 1970)
                   FROM payday_payments;
         """)
-        # Copy due value back to payment_instructions
-        cursor.run("""
-            UPDATE payment_instructions pi
-               SET due = pi2.due
-              FROM payment_instructions_due pi2
-             WHERE pi.id = pi2.id
-        """)
-        # Reset older due values to zero
-        cursor.run("""
-            UPDATE payment_instructions pi
-               SET due = '0'
-             WHERE pi.id NOT IN (
-                 SELECT id
-                   FROM payment_instructions_due pi2)
-        """)
 
         log("Updated the balances of %i participants." % len(participants))
 
--- a/gratipay/billing/payday.py
+++ b/gratipay/billing/payday.py
@@ -347,10 +347,10 @@ class Payday(object):
                 SELECT *, (SELECT id FROM paydays WHERE extract(year from ts_end) = 1970)
                   FROM payday_payments;
         """)
-        # Copy giving_due value back to payment_instructions
+        # Copy due value back to payment_instructions
         cursor.run("""
             UPDATE payment_instructions pi
-               SET giving_due = pi2.giving_due
+               SET due = pi2.due
               FROM payday_payment_instructions pi2
              WHERE pi.id = pi2.id
         """)
--- a/gratipay/billing/payday.py
+++ b/gratipay/billing/payday.py
@@ -262,36 +262,6 @@ class Payday(object):
 
 
     @staticmethod
-    def park_payment_instructions(cursor):
-        """In the case of participants in whose case the amount to be charged to their cc's
-        in order to meet all outstanding and current subscriptions does not reach the minimum
-        charge threshold, park this for the next week by adding the current subscription amount
-        to giving_due
-        """
-        cursor.run("""
-
-        INSERT INTO participants_payments_uncharged
-             SELECT participant
-               FROM payday_payment_instructions
-           GROUP BY participant
-             HAVING SUM(amount + giving_due) < %(MINIMUM_CHARGE)s
-
-        """, dict(MINIMUM_CHARGE=MINIMUM_CHARGE))
-
-        cursor.run("""
-
-        UPDATE payment_instructions
-           SET giving_due = amount + giving_due
-         WHERE id IN ( SELECT id
-                         FROM payday_payment_instructions ppi
-                            , participants_payments_uncharged ppu
-                        WHERE ppi.participant = ppu.participant
-                      )
-
-        """)
-
-
-    @staticmethod
     def transfer_takes(cursor, ts_start):
         return  # XXX Bring me back!
         cursor.run("""
--- a/gratipay/billing/payday.py
+++ b/gratipay/billing/payday.py
@@ -231,12 +231,14 @@ class Payday(object):
                     # not up to minimum charge level. cancel the hold
                     cancel_card_hold(holds.pop(p.id))
                     return
-            hold, error = create_card_hold(self.db, p, amount)
-            if error:
-                return 1
-            else:
-                holds[p.id] = hold
+            if amount >= MINIMUM_CHARGE:
+                hold, error = create_card_hold(self.db, p, amount)
+                if error:
+                    return 1
+                else:
+                    holds[p.id] = hold
         threaded_map(f, participants)
+
 
         # Update the values of card_hold_ok in our temporary table
         if not holds:
@@ -383,6 +385,7 @@ class Payday(object):
               FROM payday_payment_instructions pi2
              WHERE pi.id = pi2.id
         """)
+
         log("Updated the balances of %i participants." % len(participants))
 
 
--- a/gratipay/billing/payday.py
+++ b/gratipay/billing/payday.py
@@ -239,7 +239,6 @@ class Payday(object):
                     holds[p.id] = hold
         threaded_map(f, participants)
 
-
         # Update the values of card_hold_ok in our temporary table
         if not holds:
             return {}
--- a/sql/payday.sql
+++ b/sql/payday.sql
@@ -55,7 +55,7 @@ CREATE TABLE payday_payments_done AS
 
 DROP TABLE IF EXISTS payday_payment_instructions;
 CREATE TABLE payday_payment_instructions AS
-    SELECT s.id, participant, team, amount, giving_due
+    SELECT s.id, participant, team, amount, due
       FROM ( SELECT DISTINCT ON (participant, team) *
                FROM payment_instructions
               WHERE mtime < %(ts_start)s
@@ -77,8 +77,8 @@ CREATE INDEX ON payday_payment_instructions (team);
 ALTER TABLE payday_payment_instructions ADD COLUMN is_funded boolean;
 
 UPDATE payday_payment_instructions ppi
-   SET giving_due = s.giving_due
-  FROM (SELECT participant, team, SUM(giving_due) AS giving_due
+   SET due = s.due
+  FROM (SELECT participant, team, SUM(due) AS due
        FROM payment_instructions
        GROUP BY participant, team) s
  WHERE ppi.participant = s.participant
@@ -87,7 +87,7 @@ UPDATE payday_payment_instructions ppi
 ALTER TABLE payday_participants ADD COLUMN giving_today numeric(35,2);
 UPDATE payday_participants pp
    SET giving_today = COALESCE((
-           SELECT sum(amount + giving_due)
+           SELECT sum(amount + due)
              FROM payday_payment_instructions
             WHERE participant = pp.username
        ), 0);
@@ -134,10 +134,10 @@ RETURNS void AS $$
            SET balance = (balance + team_delta)
          WHERE slug = $2;
         UPDATE payday_payment_instructions
-           SET giving_due = 0
+           SET due = 0
          WHERE participant = $1
            AND team = $2
-           AND giving_due > 0;
+           AND due > 0;
         INSERT INTO payday_payments
                     (participant, team, amount, direction)
              VALUES ( ( SELECT p.username
@@ -154,7 +154,7 @@ RETURNS void AS $$
     END;
 $$ LANGUAGE plpgsql;
 
--- Add payments that were not met on to giving_due
+-- Add payments that were not met on to due
 
 CREATE OR REPLACE FUNCTION park(text, text, numeric)
 RETURNS void AS $$
@@ -162,7 +162,7 @@ RETURNS void AS $$
         IF ($3 = 0) THEN RETURN; END IF;
 
         UPDATE payday_payment_instructions
-           SET giving_due = $3
+           SET due = $3
          WHERE participant = $1
            AND team = $2;
     END;
@@ -180,11 +180,11 @@ CREATE OR REPLACE FUNCTION process_payment_instruction() RETURNS trigger AS $$
               FROM payday_participants p
              WHERE username = NEW.participant
         );
-        IF (NEW.amount + NEW.giving_due <= participant.new_balance OR participant.card_hold_ok) THEN
-            EXECUTE pay(NEW.participant, NEW.team, NEW.amount + NEW.giving_due, 'to-team');
+        IF (NEW.amount + NEW.due <= participant.new_balance OR participant.card_hold_ok) THEN
+            EXECUTE pay(NEW.participant, NEW.team, NEW.amount + NEW.due, 'to-team');
             RETURN NEW;
         ELSE
-            EXECUTE park(NEW.participant, NEW.team, NEW.amount + NEW.giving_due);
+            EXECUTE park(NEW.participant, NEW.team, NEW.amount + NEW.due);
             RETURN NULL;
         END IF;
         RETURN NULL;
--- a/sql/payday.sql
+++ b/sql/payday.sql
@@ -84,6 +84,13 @@ UPDATE payday_payment_instructions ppi
  WHERE ppi.participant = s.participant
    AND ppi.team = s.team;
 
+DROP TABLE IF EXISTS payment_instructions_due;
+CREATE TABLE payment_instructions_due AS
+    SELECT * FROM payday_payment_instructions;
+
+CREATE INDEX ON payment_instructions_due (participant);
+CREATE INDEX ON payment_instructions_due (team);
+
 ALTER TABLE payday_participants ADD COLUMN giving_today numeric(35,2);
 UPDATE payday_participants pp
    SET giving_today = COALESCE((
@@ -116,6 +123,7 @@ RETURNS void AS $$
     DECLARE
         participant_delta numeric;
         team_delta numeric;
+        payload json;
     BEGIN
         IF ($3 = 0) THEN RETURN; END IF;
 
@@ -133,11 +141,17 @@ RETURNS void AS $$
         UPDATE payday_teams
            SET balance = (balance + team_delta)
          WHERE slug = $2;
-        UPDATE payday_payment_instructions
+        UPDATE payment_instructions_due
            SET due = 0
          WHERE participant = $1
            AND team = $2
            AND due > 0;
+        IF ($4 = 'to-team') THEN
+            payload = '{"action":"pay","participant":"' || $1 || '", "team":"'
+                || $2 || '", "amount":' || $3 || '}';
+            INSERT INTO events(type, payload)
+                VALUES ('payday',payload);
+        END IF;
         INSERT INTO payday_payments
                     (participant, team, amount, direction)
              VALUES ( ( SELECT p.username
@@ -158,13 +172,20 @@ $$ LANGUAGE plpgsql;
 
 CREATE OR REPLACE FUNCTION park(text, text, numeric)
 RETURNS void AS $$
+    DECLARE payload json;
     BEGIN
         IF ($3 = 0) THEN RETURN; END IF;
 
-        UPDATE payday_payment_instructions
+        UPDATE payment_instructions_due
            SET due = $3
          WHERE participant = $1
            AND team = $2;
+
+        payload = '{"action":"due","participant":"' || $1 || '", "team":"'
+            || $2 || '", "due":' || $3 || '}';
+        INSERT INTO events(type, payload)
+            VALUES ('payday',payload);
+
     END;
 $$ LANGUAGE plpgsql;
 
@@ -195,7 +216,6 @@ CREATE TRIGGER process_payment_instruction BEFORE UPDATE OF is_funded ON payday_
     FOR EACH ROW
     WHEN (NEW.is_funded IS true AND OLD.is_funded IS NOT true)
     EXECUTE PROCEDURE process_payment_instruction();
-
 
 -- Create a trigger to process takes
 
--- a/sql/payday.sql
+++ b/sql/payday.sql
@@ -76,21 +76,6 @@ CREATE INDEX ON payday_payment_instructions (participant);
 CREATE INDEX ON payday_payment_instructions (team);
 ALTER TABLE payday_payment_instructions ADD COLUMN is_funded boolean;
 
-UPDATE payday_payment_instructions ppi
-   SET due = s.due
-  FROM (SELECT participant, team, SUM(due) AS due
-       FROM payment_instructions
-       GROUP BY participant, team) s
- WHERE ppi.participant = s.participant
-   AND ppi.team = s.team;
-
-DROP TABLE IF EXISTS payment_instructions_due;
-CREATE TABLE payment_instructions_due AS
-    SELECT * FROM payday_payment_instructions;
-
-CREATE INDEX ON payment_instructions_due (participant);
-CREATE INDEX ON payment_instructions_due (team);
-
 ALTER TABLE payday_participants ADD COLUMN giving_today numeric(35,2);
 UPDATE payday_participants pp
    SET giving_today = COALESCE((
@@ -141,7 +126,7 @@ RETURNS void AS $$
         UPDATE payday_teams
            SET balance = (balance + team_delta)
          WHERE slug = $2;
-        UPDATE payment_instructions_due
+        UPDATE current_payment_instructions
            SET due = 0
          WHERE participant = $1
            AND team = $2
@@ -176,7 +161,7 @@ RETURNS void AS $$
     BEGIN
         IF ($3 = 0) THEN RETURN; END IF;
 
-        UPDATE payment_instructions_due
+        UPDATE current_payment_instructions
            SET due = $3
          WHERE participant = $1
            AND team = $2;
--- a/sql/payday.sql
+++ b/sql/payday.sql
@@ -86,16 +86,16 @@ UPDATE payday_payment_instructions ppi
 
 DROP TABLE IF EXISTS participants_payments_uncharged;
 CREATE TABLE participants_payments_uncharged AS
-    SELECT participant 
+    SELECT id, giving_due 
       FROM payday_payment_instructions
      WHERE 1 = 2;
 
 ALTER TABLE payday_participants ADD COLUMN giving_today numeric(35,2);
-UPDATE payday_participants
+UPDATE payday_participants pp
    SET giving_today = COALESCE((
            SELECT sum(amount + giving_due)
              FROM payday_payment_instructions
-            WHERE participant = username
+            WHERE participant = pp.username
        ), 0);
 
 DROP TABLE IF EXISTS payday_takes;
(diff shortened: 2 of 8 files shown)


whit537:
!m @rorepo

When you're ready, go ahead and apply the "Review" lable.

gratipay/billing/payday.py
```diff
@@ -266,6 +271,26 @@ def process_payment_instructions(cursor):
...
         cursor.run("UPDATE payday_payment_instructions SET is_funded=true;")
 
+    @staticmethod
+    def park_payment_instructions(cursor):
+        """In the case of participants in whose case the amount to be charged to their cc's
+        in order to meet all outstanding and current subscriptions does not reach the minimum
+        charge threshold, park this for the next week by adding the current subscription amount 
+        to giving_due
+        """
+	cursor.run("""INSERT INTO participants_payments_uncharged
```
whit537: 4-space indents instead of tabs, please.

gratipay/billing/payday.py
```diff
@@ -308,7 +333,7 @@ def settle_card_holds(self, cursor, holds):
             SELECT *
               FROM payday_participants
              WHERE new_balance < 0
-        """)
+        """,dict(MINIMUM_CHARGE=MINIMUM_CHARGE))
```
whit537: Passing in the `dict` is extraneous, no?

whit537:
Reformatting for readability. :-)

----

Implemented the core part of the solution for issue "charge minimum in arrears, not in advance #3378" 
Overview of the solution:

1. Added a new column `giving_due` in `payment_instructions`. To be populated/incremented with the corresponding amount, each time the amount is not charged due to being under the minimum charge threshold
1. After creation of `payday_payment_instructions`, the `giving_due` value for each participant-team combination is updated with the sum of `giving_due` values for that combination, from `payment_instructions`. 
1. The `giving_today` value for each participant is calculated as `sum(amount + giving_due)` from all records for that participant in `payday_payment_instructions`
1. The `giving_today` value is checked against the `MINIMUM_CHARGE` threshold before creating card holds if at/above the threshold. Have a doubt here - whether it's the basic amount or the upcharged amount that is to be used in the check
1. In the case of participants who have been charged, all `giving_due` values in `payment_instructions` for those participants are reset to 0
1. In the case of participants who have not been charged, the `giving_due` value in the current record in `payment_instructions` is updated to `amount + giving_due`

Will await comments/suggestions before proceeding further. 

Further changes planned:

1. Add a new column `giving_due` in participants, to cache the total `giving_due` value for participants
1. Update the UI to include the above

gratipay/billing/payday.py
whit537: Removed in e087c60987c523b66902a196d4ed96ee511c1b34.

whit537:
> Have a doubt here - whether it's the basic amount or the upcharged amount that is to be used in the check

I think it should be the basic amount, because that will be more natural to people. "I give 50¢/wk, so I'll get charged every 20 weeks," rather than getting charged in the 19th week because the fee puts you over the top. I mean, that's pretty minor. It's implemented as checking the basic amount right now, iirc.

whit537:
Wtf?

screen shot 2015-08-13 at 4 01 04 pm

whit537:
We're going to need at least a few tests for this.

whit537:
I mean for this PR as a whole. I'm not seeing any tests for the new functionality.

sql/payday.sql
```diff
@@ -76,10 +76,24 @@ CREATE INDEX ON payday_payment_instructions (participant);
...
+       FROM payment_instructions 
+       GROUP BY participant, team) s
+ WHERE ppi.participant = s.participant
+   AND ppi.team = s.team;
+
+DROP TABLE IF EXISTS participants_payments_uncharged;
+CREATE TABLE participants_payments_uncharged AS
+    SELECT participant 
+      FROM payday_payment_instructions
+     WHERE 1 = 2;
```
whit537: What's going on with this construction? One does not equal two. :-)

rohitpaulk:
screen shot 2015-08-15 at 4 20 59 pm

@whit537 - Can you try re-enabling the repo at Travis?

rorepo:
Have corrected the formatting and the extraneous argument. 

The '1 = 2' thing is an old trick I've used quite a lot - when I wished to copy the structure (full or part) of some table, without the data. Might be overkill here for just one column, but for one thing, all other table creations in this script are in the form of selects and also, this usage means it will not break if the participant column is modified in payday_payment_instructions/payment_instructions.

Working on the tests now. Haven't used mock or vcr before, so took a while to figure these out.

whit537:
> Have corrected the formatting and the extraneous argument.

@rorepo Note that I pushed a couple commits towards this end as well: 22867fd32e8eca30b271fbc340df9fe75dd54372 & e087c60987c523b66902a196d4ed96ee511c1b34.

whit537:
> Can you try re-enabling the repo at Travis?

It appears to be re-enabled already. O.O

The failure on e087c60987c523b66902a196d4ed96ee511c1b34 was spurious ("GitHub rate limit exceeded"), so I restarted that build.

sql/payday.sql
whit537: https://github.com/gratipay/gratipay.com/pull/3675#issuecomment-131705592:

> The '1 = 2' thing is an old trick I've used quite a lot - when I wished to copy the structure (full or part) of some table, without the data. Might be overkill here for just one column, but for one thing, all other table creations in this script are in the form of selects and also, this usage means it will not break if the participant column is modified in payday_payment_instructions/payment_instructions.

whit537:
> Working on the tests now. Haven't used mock or vcr before, so took a while to figure these out.

Cool. You almost certainly don't need vcr to test this code, and probably not even mock.

rorepo:
@whit537 - I'm afraid I didn't get your comment about not needing vcr/mock. The way I figured it, I'd have to update test_billing_payday.py, which uses these. I'd have to add new tests as well as update existing ones, because the 'charge-in-arrears' functionality would make some of these invalid e.g. 'test_payday_moves_money', which was for a payment of $6 and which would now not be charged since this is now below the charge threshold.

whit537:
@rorepo I guess you're right. :-)

I was just going on a hunch, since you're actually touching code, don't mind me. Sorry! :-)

rorepo:
Looking at the tests exposed a few bugs as well as the fact that quite a few of the existing tests would have to change. 

The key points in tests:
* test_billing_payday.py: ```test_payday_moves_money``` needed to be replaced by two new methods - ```test_payday_moves_money_above_min_charge``` and ```test_payday_does_not_move_money_below_min_charge```, to test the behaviour when the amount to be charged is **>= MINIMUM_CHARGE** and **< MINIMUM_CHARGE**, respectively
* Other tests have needed related changes. Primarily, to increase amount to **>= MINIMUM_CHARGE**, to make sure the test works as intended

@whit537 I've removed the method ```park_payment_instructions``` from payday.py. This had actually been unused - I'd forgotten to remove it before pushing

whit537:
The commit log on this PR looks weird. Why are there two of some commits? Why is the 1880 version bump on here?

whit537:
I think maybe we need a rebase of this branch to clean up the commits.

whit537:
I pulled out the first commit, 13faa2a435f2e5331a8147533eaebc79ebf9899a, into a new PR (#3695). I'm getting bit by #3579 over on #3694, and also had this fix on #3568. Let's take care of this, already. :-)

whit537:
Rebasing on master got rid of the 1880 version bump. Now for the duplicate commits ...

rorepo:
Don't quite understand, @whit537. I did do a rebase before merging and I'd assume that's how the 1880 version came in.
Also, which are the duplicated commits you mean? If you mean the code formatting and cleanup, the latest one is what I did following some errors/warnings shown during the rebase about trailing spaces. Also removed an 'import time' I'd put in in between while running ```test_history.py```.

whit537:
@rorepo After rebasing on master, here are the commits I'm seeing:

screen shot 2015-08-19 at 1 59 02 pm

It *looks* like we have duplicate commits for:

 - Preliminary commit - charging in arrears
 - Schema changes moved to branch.sql
 - Clean up code formatting
 - Remove extraneous argument

whit537:
> Don't quite understand, @whit537.

Neither do I. :-) My git-fu is not that strong.

I think at this point we should rebase to squash the duplicate commits, and then make sure the overall diff is as expected, and then forge ahead. I'm going to rebase ... unless you object?

whit537:
`$ git rebase -i f887fdd5f8d146d4c2ff5849143fe38059363c3e`

... presents me with ...

```
pick 97ead4e Preliminary commit - charging in arrears                                                         
pick 5b58b70 Schema changes moved to branch.sql                                                               
pick 14045c6 Clean up code formatting                                                                         
pick d9980f5 Remove extraneous argument                                                                       
pick 816776f Preliminary commit - charging in arrears                                                         
pick af42c27 Schema changes moved to branch.sql                                                               
pick 005573e Clean up code formatting                                                                         
pick fe2c70f Remove extraneous argument                                                                       
pick 1b3a373 Refined and added tests                                                                          
pick 98f0f3e Code formatting and cleanup                                                                      
                                                                                                              
# Rebase f887fdd..98f0f3e onto f887fdd
```

rorepo:
You're welcome to proceed, @whit537. My git-fu is almost certainly weaker than yours, in any case :-) 
Saw the duplicates. Hadn't checked that far back earlier.

whit537:
> You're welcome to proceed, @whit537.

Okay, working on it ...

whit537:
Here's what I'm trying to get to work:

```
pick 97ead4e Preliminary commit - charging in arrears                                                         
fixup 816776f Preliminary commit - charging in arrears                                                        
pick 5b58b70 Schema changes moved to branch.sql                                                               
fixup af42c27 Schema changes moved to branch.sql                                                              
pick 14045c6 Clean up code formatting                                                                         
fixup 005573e Clean up code formatting                                                                        
pick d9980f5 Remove extraneous argument                                                                       
fixup fe2c70f Remove extraneous argument                                                                      
pick 1b3a373 Refined and added tests                                                                          
pick 98f0f3e Code formatting and cleanup                                                                      
                                                                                                              
# Rebase f887fdd..98f0f3e onto f887fdd
```

whit537:
The problem I'm running into is that the `park_payment_instructions` ends up in the final diff. The rebased `Refined and added tests` commit doesn't remove that method, for some reason. That makes me wonder what *else* is awry.

whit537:
Alright, new strategy:

```
pick 97ead4e Preliminary commit - charging in arrears                                                         
pick 5b58b70 Schema changes moved to branch.sql                                                               
pick 14045c6 Clean up code formatting                                                                         
pick d9980f5 Remove extraneous argument                                                                       
pick 1b3a373 Refined and added tests                                                                          
pick 98f0f3e Code formatting and cleanup                                                                      
                                                                                                              
# Rebase f887fdd..98f0f3e onto f887fdd
```

I compared the diffs for the duplicate commits, and decided that the second of each ought simply to be skipped:

 - Preliminary commit - charging in arrears: 816776f9dd7db8761f7caf15001e5dd5af1138ef makes three additions that are already included with 97ead4ea8cfab5fa2db7eb157c4d0e7b2c59be8c.
 - Schema changes moved to branch.sql: the one deletion in af42c27c7801362c2fce33ec841e801a89b97250 is included in 5b58b70fbe9e40abdabdd672f1d1dd91bdedc791.
 - Clean up code formatting: 005573edea63535729aaa17f126a6c3dffab12da is the same as 14045c61eadafb72cebef807dde3cd86f3216384 ... except! for a tell-tale second `park_payment_instructions` at the end of the diff.
 - Remove extraneous argument: fe2c70f01f2b7ffd7f796576bcdb9a856edf1ba4 is the same as d9980f52a51b6eaa99c05005c5b22ed7e6733a75.

rorepo:
Strange. I had this method come up as a conflict both when rebasing as well as when I tried to push to the branch (and was told to do a pull first). But thought I did finally get rid of it.

whit537:
Ah, nuts! The 5c23e5b489d7efc861dad2a7b93fbc07483bbdff commit *still* doesn't remove `park_payment_instructions`. :rage:

whit537:
That's the only difference between the diffs.

whit537:
https://github.com/gratipay/gratipay.com/commit/5c23e5b489d7efc861dad2a7b93fbc07483bbdff.diff
https://github.com/gratipay/gratipay.com/commit/1b3a373.diff

whit537:
The "Code formatting and cleanup" diff is good.

https://github.com/gratipay/gratipay.com/commit/aad163e7875c6143c4b6c939d30c50e8db578860.diff
https://github.com/gratipay/gratipay.com/commit/98f0f3e.diff

whit537:
How does the diff look to you, @rorepo? Are we back on track here?

rorepo:
Diff looked ok, but just in case, took the precaution of comparing the full file (payday.py). Matches my local copy. And no park_payment_instructions in evidence :-)

whit537:
> Matches my local copy. And no park_payment_instructions in evidence :-)

Awesome. :-)

Onward with review ...

rorepo:
Did the same with payday.sql. Afraid it's got this
```DROP TABLE IF EXISTS participants_payments_uncharged;
CREATE TABLE participants_payments_uncharged AS
    SELECT id, giving_due
      FROM payday_payment_instructions
     WHERE 1 = 2;```
which shouldn't be there (this table was meant for ```park_payment_instructions```)

**Comment by whit537** [comment]
@rorepo Okay. Can I let you go ahead and remove that?

**Comment by rorepo** [comment]
Remove that - on github itself, you mean?

**Comment by whit537** [comment]
@rorepo Yeah, I just meant make another commit here on this PR to sync up payday.sql with what it should be.

**Comment by rorepo** [comment]
@whit537 Done :-) Doubt if it would've been so easy if I'd tried pushing my copy again

**Comment by whit537** [comment]
Cool, thanks. :)

**Comment by rorepo** [comment]
Well, you did most of the hard work, @whit537 :-) I'm not comfortable enough with git yet, to handle this kind of a mess.

**Comment by whit537** [comment]
I still have no idea how we got into that mess, though. :-(

Okay!

I still want *some* kind of record of the actual deferred payment. It could be a log message during payday, or an entry in the events table, or a record in a new table. We need *something* to help with debugging when the day comes that we need to debug something around this.

Do we have adequate test coverage for the behavior of `giving_due` once a payment is satisfied?

Also, let's use `payment_due` or just simply `due` instead of `giving_due`, as the latter is with reference to the ~user's point of view, and we should use the viewpoint-agnostic term this far down in the app.

**Comment by rorepo** [comment]
The ```giving_due``` column in the active record in ```payment_instructions``` would always have the value of the deferred payment, but I agree it would be useful to have this information readily available in something like an event log for ready reference rather than having to query the table. 
Could I have an explanation for the events table? Perhaps a sample of the records?

Thinking of test coverage, it occurs to me that there are a few more that might be useful:

1.  ```giving_due``` value is reset to 0 after a successful payment, as you suggested
2.  Carry-over of ```giving_due``` value is taken into account when checking against the charge threshold i.e. ```amount + giving_due``` is indeed used in this
3.  Things work ok when the subscription value is changed, both upwards and downwards. Here, I have a question about how the change should be handled. I'm assuming that we do not change things in mid-week. Say, if the subscription is changed on Monday from $3 to $5, we keep any existing ```giving_due```, accrued at the rate of $3/week, unchanged and only apply the $5 to subsequent weeks.

I could change ```giving_due``` to ```due``` in the meantime

I have one other question - should I be uploading yml files if they've changed?

**Comment by whit537** [comment]
> The `giving_due` column in the active record in `payment_instructions` would always have the value of the deferred payment[.]

This picks up with the conversation from https://github.com/gratipay/gratipay.com/issues/3378#issuecomment-126612705:

> If we have multiple records in [payment_instructions] corresponding to each individual [payment instruction] (modification), seems to me we could use the giving_due column in [payment_instructions] itself and still have a capability to trace giving_due back to individual instances of the [payment instruction].

But we change the record in `payment_instructions` when the amount of the payment changes, not simply when the payment is fulfilled (i.e., actually charged).

1. payment instruction 0: $2.00 from ~alice to the Bakelite Team
1. Payday N: $2.00 due from ~alice to Bakelite, no charge
1. payment instruction 1: $3.00 from ~alice to Bakelite
1. payment instruction 2: $2.50 from ~alice to Bakelite
1. Payday N+1: $4.50 due from ~alice to Bakelite, no charge
1. Payday N+2: $7.00 due from ~alice to Bakelite, no charge
1. payment instruction 3: $1.00 from ~alice to Bakelite
1. Payday N+3: $8.00 due from ~alice to Bakelite, no charge
1. payment instruction 4: $4.00 from ~alice to Bakelite
1. Payday N+4: $12.00 due from ~alice to Bakelite, charge!

I'm looking for an easy way to produce this history:

1. Payday N: $2.00 due from ~alice to Bakelite because of payment instruction 0
1. Payday N+1: $4.50 due from ~alice to Bakelite because of payment instruction 2
1. Payday N+2: $7.00 due from ~alice to Bakelite because of payment instruction 2
1. Payday N+3: $8.00 due from ~alice to Bakelite because of payment instruction 3
1. Payday N+4: $12.00 due from ~alice to Bakelite because of payment instruction 4
1. Payday N+4: $0.00 due from ~alice to Bakelite because of charge X

Given the existing implementation on this PR, how would we produce that history?

**Comment by whit537** [comment]
> Could I have an explanation for the events table? Perhaps a sample of the records?

Grep for the `add_event` function for examples of usage.

**Comment by whit537** [comment]
> I have one other question - should I be uploading yml files if they've changed?

Yes, but in a separate commit from the test that uses the new yaml fixture, to avoid cluttering diffs.

**Comment by whit537** [comment]
> Here, I have a question about how the change should be handled. I'm assuming that we do not change things in mid-week. Say, if the subscription is changed on Monday from $3 to $5, we keep any existing giving_due, accrued at the rate of $3/week, unchanged and only apply the $5 to subsequent weeks.

Correct. We modify `due` exactly once per week: during payday. Each payday has a starting timestamp, and that's [the cutoff for which payment instruction we use](https://github.com/gratipay/gratipay.com/blob/1882/sql/payday.sql#L61) during that particular payday. A payment instruction can and sometimes will change many times during the week, but whatever a payment instruction is when payday starts, that's what we'll go with.

**Comment by rorepo** [comment]
Whatever's this conflict? Tried looking at the diff but didn't get it.

**Comment by whit537** [comment]
@rorepo The conflicts are due to #3698 landing on master. I'm rebasing now ...

**Comment by whit537** [comment]
Tentative rebase in https://github.com/gratipay/gratipay.com/compare/cia. Let's compare those commits to the ones on this PR. Does that look right? If so we can force-push that over `charge-in-arrears` and we'll be back on track here.

**Comment by whit537** [comment]
Alright! Rebased. :-)

Back on track, @rorepo?

**Comment by rorepo** [comment]
Yes, looks ok, @whit537

**Comment by rorepo** [comment]
@whit537, I implemented a new test, covering the scenario you were talking about - besides the history aspect, verifying these different permutations was necessary anyway. 

I'm afraid I had to make a design change. Kept running into an error like this:
> E       OperationalError: tuple to be updated was already modified by an operation triggered 
by the current command
E       HINT:  Consider using an AFTER trigger instead of a BEFORE trigger to propagate 
changes to other rows.

This appeared to be caused by the updates of **payday_payment_instructions** records in function ```process_payment_instruction``` in payday.sql, conflicting with the BEFORE trigger on this table. This error had come up earlier and it seemed to go away by itself (in other words, I don't know how), but it was persistent this time. I therefore introduced a new table **payment_instructions_due** for the 'pay' and 'park' updates of 'due'.

About the history bit - I'm afraid I couldn't make use of ```add_event``` itself, since payday.py does not actually have visibility of subscription level operations. Therefore I inserted records into **events** directly, from the 'pay' and 'park' points in payday.sql. I used *payday* as the event type.

At the moment, facing a problem with git push. I started with a git fetch origin and fresh checkout of origin/charge-in-arrears and made my changes on top of that. Checking manually on github, git diff, etc. suggested things were fine. Made my (2) commits, after which git status showed my local branch ahead of origin/charge-in-arrears by 2 commits. At this point, I did a rebase origin/master preparatory to pushing the changes and got the message 
> Your branch and 'origin/charge-in-arrears' have diverged,
and have 31 and 11 different commits each, respectively.

Will need some help understanding/fixing this before I push.

**Comment by whit537** [comment]
> At the moment, facing a problem with git push.

Hmmm ... my first hunch is that maybe you've got a local `charge-in-arrears` branch that is conflicting with `origin/charge-in-arrears`?

**Comment by rorepo** [comment]
Well, this is how I tried to ensure things were in sync:

1. I did have a local ```charge-in-arrears branch```, in which I'd been doing most of my trials
2. When I had a reasonable idea of the implementation, I took a look at github and it had some changes in the meantime, such as the introduction of 'TheEnterprise', 'picard', etc. Therefore, I renamed (moved) my existing local branch as ```charge-in-arrears-old```, did a git fetch origin and then 'git checkout charge-in-arrears' at which point I got a message like 'branch charge-in-arrears set up to track origin/charge-in-arrears. Checking the log as well as manually looking at the code, things seemed to be in sync, as expected.
3. I then (manually) transferred my changes to the new branch and made my commits. Git status at this point informed me that my branch was 2 commits ahead of ```origin/charge-in-arrears```, as expected
4. Did git rebase origin/master. It was at this point that I got the message about branches having diverged.

I tried deleting the old branch, but that didn't change anything.

**Comment by rorepo** [comment]
Been trying different permutations. Here's the history of the last one:
```git branch -D charge-in-arrears```
```git branch -D charge-in-arrears-old```
```git fetch origin```
```git rebase origin/master```
*Current branch master is up to date.*
```git checkout charge-in-arrears```
*Branch charge-in-arrears set up to track remote branch charge-in-arrears from origin.*
```git status```
*On branch charge-in-arrears*
*Your branch is up-to-date with 'origin/charge-in-arrears'.*
```git rebase origin/charge-in-arrears```
*Current branch charge-in-arrears is up to date.*
```git diff origin/charge-in-arrears```
no differences shown
```git diff origin/master```
shows differences
```git rebase origin/master```
*First, rewinding head to replay your work on top of it...*
*Applying: Preliminary commit - charging in arrears*
*Applying: Schema changes moved to branch.sql*
*Applying: Clean up code formatting*
*Applying: Remove extraneous argument*
*Applying: Refined and added tests*
*Applying: Code formatting and cleanup*
*Applying: Kill park_payment_instructions method*
*Applying: Trim trailing whitespace*
*Applying: Removed extraneous table script*
*Applying: Renamed payment_instructions 'giving_due' to 'due'*
*Applying: Fix whitespace typo*
```git status```
*On branch charge-in-arrears*
*Your branch and 'origin/charge-in-arrears' have diverged,*
*and have 29 and 11 different commits each, respectively.*
 * (use "git pull" to merge the remote branch into yours)*

**Comment by whit537** [comment]
@rorepo Okay, I'm able to reproduce your result. It's normal and expected. What has happened is that `master` has moved on since the last time we rebased this PR branch `charge-in-arrears` on it. If you scroll back through git history on a clean `charge-in-arrears` checkout then you should see that it is based on the 1884 version. After the rebase dance you describe (resulting in "29 and 11 different commits each, respectively") then you should see that `charge-in-arrears` is now has the 1887 version just before this PR's 11 commits.

That means you are doing it right! You are properly replaying this branch's commits onto the latest `master`. Now all that remains is to force-push your local, rebased `charge-in-arrears` branch back up to the GitHub remote.

Here's what a regular `git push` will look like:

```
$ git push
To git@github.com:gratipay/gratipay.com.git
 ! [rejected]        charge-in-arrears -> charge-in-arrears (non-fast-forward)
error: failed to push some refs to 'git@github.com:gratipay/gratipay.com.git'
hint: Updates were rejected because the tip of your current branch is behind
hint: its remote counterpart. Integrate the remote changes (e.g.
hint: 'git pull ...') before pushing again.
hint: See the 'Note about fast-forwards' in 'git push --help' for details.
$
```

You need to *force*-push since we want to alter the history of this PR to include the latest changes on `master`. Go ahead and do this after the `dance` you [described](https://github.com/gratipay/gratipay.com/pull/3675#issuecomment-134283403):

```
$ git push --force
```

Or just:

```
$ git push -f
```

Force-pushing PR branches is fine and normal. **BUT**, force-pushing `master` itself is Very Bad&trade; and you should never do it except in very special circumstances, so make sure you only ever force-push PR branches. :-) 

I'll go ahead and let you do the force-push of the rebased PR so you get a feel for it.

Once you've rebased on `master` then you should be able to push your two additional commits with no problem.

**Comment by rorepo** [comment]
Thanks for the clarification, @whit537. There's still something I'm doubtful about, though. Here are two extracts from history (I used git log --oneline --decorate --graph --all). It appears that the 11 commits brought in by my rebase at this point were already merged into master, under different commit ids (the first time is after version 1884). Is this ok?

branch history 1
branch history 2

Also, one question about the force push - do I do this right after the rebase and then only do my additional commits and push?

**Comment by whit537** [comment]
> Is this ok?

Hmm ... no, that doesn't sound okay. To clarify: those history snippets are from immediately after the [steps you described above](https://github.com/gratipay/gratipay.com/pull/3675#issuecomment-134283403)?

> Also, one question about the force push - do I do this right after the rebase and then only do my additional commits and push?

It actually doesn't matter. If your force-push includes the two additional commits then that is fine.

Here are some diagrams to explain what rebasing does. It replays individual commits starting at a new base commit, and *it results in new commits*. That's why you have to force-push to change the branch (remember that branch names are simply aliases for SHA1 commit hashes).

#### First

The `charge-in-arrears` branch is based off of the commit pointed to by `master`.

screen shot 2015-08-26 at 11 43 04 am

#### Second

The `master` branch is advanced to include two commits that are *not* also on `charge-in-arrears`.

screen shot 2015-08-26 at 11 43 35 am

#### Third

Rebasing `charge-in-arrears` on the new `master` (`d`) creates new commits, 

screen shot 2015-08-26 at 11 45 21 am

Because, following the rebase, one's local `charge-in-arrears` no longer shares history with the remote `charge-in-arrears`, a force-push is necessary to repoint the remote `charge-in-arrears` to `y`.

If, after the local rebase but before force-pushing, you had added a `z` commit locally, then a force-push would still have been necessary to alter the history of the remote `charge-in-arrears`, and the new `charge-in-arrears` would include `z`.

**Comment by whit537** [comment]
Being "on `charge-in-arrears`" means being in the history of the commit pointed to by the alias `charge-in-arrears`.

**Comment by rorepo** [comment]
I understand rebase. But haven't been able to figure out how those duplicates got in, or how to get them out - other than perhaps cherry picking (out) these while merging.

**Comment by whit537** [comment]
@rorepo I noticed d28f5daa7230de74fc9407d5453d943013394866 and 1a2f382f13351b61db944e897684e7f2bc74a573 while digesting your last comment—are those your two most recent commits? I.e., are we on-track on this PR as it stands?

**Comment by whit537** [comment]
@rorepo Okay, I think I know what's happening: up at https://github.com/gratipay/gratipay.com/pull/3675#issuecomment-133364613 I made a second branch, `cia`, as part of testing out the rebase I was doing. If I'm reading your second history extract correctly (formatted as it is, it's a bit hard to follow), we are looking at the commit on `cia` there, *not* on `master`.

Yeah, I see the same commit hashes in my own history using `--all`:

```
| * 0208bfd (upstream/pull/3706, origin/turn-communities-off) Turn communities off
|/  
| * 9762b75 (origin/cia, cia) Fix whitespace typo
| * 95732c4 Renamed payment_instructions 'giving_due' to 'due'
| * 5bce7c8 Removed extraneous table script
| * ab653dc Trim trailing whitespace
| * 64f0f57 Kill park_payment_instructions method
| * eaa5cc2 Code formatting and cleanup
| * 096d2b9 Refined and added tests
| * 21fff58 Remove extraneous argument
| * 63e2521 Clean up code formatting
| * 84d99c6 Schema changes moved to branch.sql
| * 3582630 Preliminary commit - charging in arrears
|/  
| * 38109c0 (str) Bring back a couple lines that were rebased out
```

**Comment by whit537** [comment]
I like `--all`, thanks for the tip. :-)

**Comment by whit537** [comment]
Actually, I like `--decorate` even better. :)

**Comment by rorepo** [comment]
Yes, @whit537, the two commits you mentioned are the final ones I made and other than for the thing about the duplicates, things are on track.

I was kind of tickled to find ```--decorate``` too. Almost understood what it showed :-)

**Comment by whit537** [comment]
> other than for the thing about the duplicates

Do you buy my explanation at https://github.com/gratipay/gratipay.com/pull/3675#issuecomment-135132355? :-)

**Comment by rorepo** [comment]
Buy, yes. But understood - only after I tried the graph without the ```--all``` :-) 

So now, looks like we're all set :-)

**Comment by whit537** [comment]
Huzzah! So let's see about this new code now ... :-)

**Comment by rorepo** [comment]
I'd been puzzled by ```current_payment_instructions``` - didn't see it being used anywhere. But then, forgot about it. 
Now, occurs to me that using this instead would also do away with the necessity for the first update. 

However, not sure I got this:
> let's change this test to exercise that.

**Comment by whit537** [comment]
> However, not sure I got this:

In other words, change queries like this:

```sql
            SELECT DISTINCT ON (participant, team) due
              FROM payment_instructions ppi
             WHERE ppi.participant = 'obama'
               AND ppi.team = 'TheEnterprise'
          ORDER BY participant, team, mtime DESC
```

into queries like this:

```sql
            SELECT due
              FROM current_payment_instructions cpi
             WHERE ppi.participant = 'obama'
               AND ppi.team = 'TheEnterprise'
```

Maybe even moving it into a method (untested):

```python
class Participant:
    [...]
    
    def get_due(self, team):
        return self.db.one("""
            SELECT due
              FROM current_payment_instructions cpi
             WHERE ppi.participant = %s
               AND ppi.team = %s
        """, (self.username, team.slug))
```

```

**Comment by rorepo** [comment]
Ok, got it :-)

**Comment by whit537** [comment]
> (untested)

By which I meant, "I have not yet tested or run the following code." :)

**Comment by whit537** [comment]
Today is payday! Maybe we can run with this?

File: gratipay/billing/payday.py [hunk 4]
```diff
@@ -354,6 +361,22 @@ def update_balances(cursor):
...
                   FROM payday_payments;
         """)
+        # Copy due value back to payment_instructions
+        cursor.run("""
+            UPDATE payment_instructions pi
+               SET due = pi2.due
+              FROM payment_instructions_due pi2
+             WHERE pi.id = pi2.id
+        """)
+        # Reset older due values to zero
```
whit537: I think this is extraneous. If we use the `current_payment_instructions` view in order to get the `due` amount, then we'll always get the current `due` amount. We'll never use values from old `payment_instructions` records, so it doesn't matter whether they're zero or some other value. Right?

sql/payday.sql
```diff
@@ -55,7 +55,7 @@ CREATE TABLE payday_payments_done AS
 
 DROP TABLE IF EXISTS payday_payment_instructions;
 CREATE TABLE payday_payment_instructions AS
-    SELECT participant, team, amount
+    SELECT s.id, participant, team, amount, due
       FROM ( SELECT DISTINCT ON (participant, team) *
```
whit537: This subquery can be replaced with a query against `current_payment_instructions`.

whit537:
I've added a `get_due` method.

I saw a single test failure on [this line](https://github.com/gratipay/gratipay.com/blob/1cfe67452ce240e1272863a2da043fb4a3ff6c7f/tests/py/test_billing_payday.py#L499) under that last commit (the actual was 0), but I haven't been able to reproduce it.

sql/payday.sql
```diff
@@ -76,12 +76,27 @@ CREATE INDEX ON payday_payment_instructions (participant);
 CREATE INDEX ON payday_payment_instructions (team);
 ALTER TABLE payday_payment_instructions ADD COLUMN is_funded boolean;
 
+UPDATE payday_payment_instructions ppi
+   SET due = s.due
+  FROM (SELECT participant, team, SUM(due) AS due
```
whit537: Why are we summing `due` here? My understanding was that `due` in `current_payment_instructions` would always be cumulative.

tests/py/test_billing_payday.py
```diff
@@ -29,8 +29,91 @@ def test_payday_moves_money(self):
...
+        Enterprise = self.make_team(is_approved=True)
+        self.obama.set_payment_instruction(Enterprise, '2.00')  # < MINIMUM_CHARGE
+
+        fch.return_value = {}
+        Payday.start().run()    # payday 0
+
+        assert self.obama.get_due('TheEnterprise') == D('2.00')
+
+        self.obama.set_payment_instruction(Enterprise, '3.00')  # < MINIMUM_CHARGE
+        self.obama.set_payment_instruction(Enterprise, '2.50')  # cumulatively still < MINIMUM_CHARGE
```
whit537: Cumulative with what? Not cumulative with the 3.00 just above, since this replaces that, right?

sql/payday.sql
rorepo: Well, I'm thinking of a situation like this:

1. Payday 0 - subscription amount $3, uncharged, resulting in due $3
2. Subscription amount changed to $4 i.e new record entered with amount = $4 (and due = 0)
3. Payday 1 - ```current_payment_instructions``` would pick up the latest record, with due = 0 and the existing due of $3 would be missed

In this situation, we'd need some way to transfer the existing due to the new record at step 2. Thought of using an AFTER INSERT trigger, but updating ```payday_payment_instructions``` seemed simpler.

tests/py/test_billing_payday.py
rorepo: Correct. Just the 2.50 at this point.

sql/payday.sql
whit537: Gotcha. I think we should update the `due` column every time a new `payment_instruction` is `INSERT`ed. We want to be able to surface that data in the UI, so we should keep it current, we shouldn't only ever update it during payday.

> Thought of using an AFTER INSERT trigger[.]

How about carrying the `due` column forward during `Participant.set_payment_instruction`?

sql/payday.sql
rorepo: The current subquery has this:
```WHERE mtime < %(ts_start)s```
while ```current_payment_instructions``` does not.
Can this be a problem?

Comment (whit537) [inline sql/payday.sql]: Yeah, we'll need to move that `WHERE` clause up into the parent query.

Comment (rorepo) [inline sql/payday.sql]: As I understand it, ```ts_start``` is defined at the start of payday. This would make it 
difficult to incorporate this into the ```current_payment_instructions``` view as a condition that would be meaningful at other times as well as at payday.

I'd also like to understand the situation where ```mtime``` in ```payment_instructions``` 
could/would be >= payday ```ts_start```. e.g. can a user specify a start time that is in the future,
for a subscription? Was unable to find any such provision in the dev environment.

Comment (rorepo) [inline sql/payday.sql]: I've made all changes except for this particular one, because I'm not sure about the filtering by ```ts_start```. As I see it, ```payment_instructions``` can only have a record with ```mtime``` > ```ts_start``` if a new subscription is created after the payday run has been started. And if this situation comes up, we do have a problem - the ```current_payment_instructions``` view will pick up this latest record, while ```payday_payment_instructions``` (as it is now) will not. If we do update the subquery, this could have different results depending on the exact point in payday when the new subscription is created.

Once again, I've got a 'diverged' situation after rebasing with master:
>Your branch and 'origin/charge-in-arrears' have diverged,
and have 196 and 14 different commits each, respectively.

However, this seems ok after checking the log - will just have to do a forced push like last time.

**Comment by whit537** [comment]
> Once again, I've got a 'diverged' situation after rebasing with master:

Yeah, that's normal. Everytime you rebase you're rewriting history, and you're going to have to force-push to get the remote (GitHub) to accept your rewrite.

**Comment by whit537** [comment]
@rorepo I rebased on master to resolve conflicts between this PR and .https://github.com/gratipay/gratipay.com/commit/aef017ab20b38fc092d2014422425aa293d764e7

**Comment by whit537** [comment]
That means you'll need to rebase your most recent local commits before you can push.

File: sql/payday.sql (hunk 5 above)
Comment (whit537) [inline sql/payday.sql]: > [T]he `current_payment_instructions` view will pick up this latest record[.]

Good point. Withdrawn. :-)

**Comment by whit537** [comment]
Alright, @rorepo, I think we just need to get your latest commits on here and then we're ready for final review! :-)

**Comment by rorepo** [comment]
Working on it, @whit537. There were a few conflicts to be sorted out. 

Ran into a new problem in testing: 
> 'column ncc_failing of relation paydays does not exist'. 

I see schema.sql now has a 'DROP COLUMN' for this and other columns. Payday.py still refers this column - possibly a merge issue when I rebased?. Tests run ok if I comment out the drop line for this column.

**Comment by whit537** [comment]
@rorepo Yeah, sorry for the curveball. :-/

`ncc_failing` is one of the columns we dropped from `paydays`, so yes it's probably a rebase issue. The only reference I see to `ncc_failing` on `master` is a test that is marked `xfail`.

**Comment by rorepo** [comment]
I see ```ncc_failing``` is directly referenced only in an ```xfail``` marked test. However, it is this bit of code in payday.py that is causing the problem for me:
>        # Record the number of failures
        cursor.one("""
            UPDATE paydays
               SET ncc_failing = %s
             WHERE ts_end='1970-01-01T00:00:00+00'::timestamptz
         RETURNING id
        """, (n_failures,), default=NoPayday)

This is in ```create_card_holds```. Could this be something that crept back in when I rebased?

**Comment by whit537** [comment]
> Could this be something that crept back in when I rebased?

Yeah, that got taken out in https://github.com/gratipay/gratipay.com/commit/aef017ab20b38fc092d2014422425aa293d764e7 on #3730. Try your rebase again?

**Comment by whit537** [comment]
I saw the same conflict when I rebased this branch on master.

**Comment by rorepo** [comment]
Rebasing didn't make any difference. 

I uncommented the 'DROP COLUMN' before pushing - so tests will fail for the moment.

**Comment by whit537** [comment]
Did I introduce that code into this branch when I did my latest rebase onto master?

**Comment by whit537** [comment]
Blech. Looks like it. :-( It's in the diff here.

**Comment by whit537** [comment]
Sorry, @rorepo. My git-fu needs leveling up. :)

**Comment by rorepo** [comment]
@whit537, it's beyond my git-fu at the moment, which is why I decided to push as is :-) 

I could try correcting on github if you could point me in the right direction.

**Comment by whit537** [comment]
@rorepo I'll re-rebase now that we have your latest commit on here ...

**Comment by rorepo** [comment]
I tried commenting out that code in payday.py in my local branch - tests ran ok. So looks like that (and one more line - see last bit of messages below) is indeed the only conflict. 
However, saw another problem. Got this when I ran the tests:
> gratipay/models/participant.py:1175: redefinition of unused 'update_giving' from line 916
gratipay/models/participant.py:1207: redefinition of unused 'update_due' from line 946
gratipay/models/participant.py:1244: redefinition of unused 'update_is_free_rider' from line 999
gratipay/models/participant.py:1260: redefinition of unused 'profile_url' from line 815
gratipay/models/participant.py:1267: redefinition of unused 'get_teams' from line 822
gratipay/models/participant.py:1278: redefinition of unused 'get_old_teams' from line 833
gratipay/models/participant.py:1294: redefinition of unused 'insert_into_communities' from line 849
gratipay/models/participant.py:1312: redefinition of unused 'change_username' from line 867
gratipay/billing/payday.py:240: local variable 'n_failures' is assigned to but never used

I checked the first few methods - they are defined twice now.

**Comment by whit537** [comment]
Okay, I effed up the rebase. I will try to clean it up here once I get payday wrapped (https://github.com/gratipay/inside.gratipay.com/issues/329). Sorry. :-(

**Comment by rorepo** [comment]
Good luck :-)

**Comment by whit537** [comment]
Thanks! :-)

**Comment by whit537** [comment]
@rorepo I looks like we have duplicate commits again ...

**Comment by whit537** [comment]
I ran this locally to get your latest commit without the two extra commits I was seeing on feeec33 (2493f711fbed2f758211858851935a8f6d6f0741 and 2493f711fbed2f758211858851935a8f6d6f0741):

```
$ git fetch
[...]
$ git cherry-pick feeec33
$ git push --force
```

whit537:
And now we're down to a pyflakes error! :-)

whit537:
Aaaaaaaand that's the error you were pointing out at https://github.com/gratipay/gratipay.com/pull/3675#issuecomment-137605106. :-(

whit537:
It's your latest commit, a8a6bc4bc8c564d2bb100b22d0078ae1bde21e83, that adds the duplicate methods to `Participant`.

whit537:
I've force-pushed a8a6bc4 off the branch, and now we're green again. What was the intention with that commit? The code doesn't seem to match what it says on the tin:

> **Updated to use current_payment_instructions view**
> current_payment_instructions view used in place of the new payment_instructions_due table which has now been removed

whit537:
feeec33079972f53715ba12d8d474eabc523bfd3 and a8a6bc4bc8c564d2bb100b22d0078ae1bde21e83 are identical.

whit537:
Once we land this, should we issue refunds for the remaining balances of people that we charged in advance? It would make #3539 easier.

whit537:
@rorepo I think we need you to redo your last commit.

rorepo:
I think where my merges go wrong is when things go to conflict resolution. Either I'm not reading/handling the conflicts right or git is deficient in picking up/highlighting things like duplicated code. But my understanding of git is improving, so hopefully this won't be a perennial problem.

A few thoughts on refunding remaining balances charged in advance:

1. We'd need a threshold for this as well, otherwise we'd end up trying to refund even below the min payout value, say.
2. There's also the aspect that with small balances which would be consumed in a couple of weeks, there may not be much point in doing the refund. So maybe we can link the threshold to this, say, 2 weeks into subscription amount.
3. In any case, I think we should try to eliminate the situation where we charge (in arrears) and end up refunding excess balance right away.
4. Remembering @rohitpaulk's comment a while back, that excess balances as a result of being charged in advance were a relatively low percentage of transaction volumes - what would be the percentage of such balances once we've switched to charging in arrears?

whit537:
Good work, @rorepo. :-)

!m @rorepo

rorepo:
:-) Glad there were no more problems.

What about the refund bit? Will it make sense to start looking at this?

whit537:
@rorepo Well, we'll see if there any more problems on Thursday! :-)

I've reticketed the refund as #3763.

!m @rorepo

PR #3678, opened 2015-08-12
rejected -> closed

whit537:
"Rejecting" teams is too harsh a term for the situation we want to optimize for, wherein we reach a decision collaboratively together with team owners about whether they and we are a good fit for each other. I think we should use the term "close" instead. That leaves ambiguous the circumstances surrounding the closing, whereas "rejected" only covers the harshest case, wherein Gratipay acts despite the owner's intention. A "closed" account could be closed by the owner's decision, by Gratipay's, or by both agreeing together.

### DIFF
--- a/www/%team/index.html.spt
+++ b/www/%team/index.html.spt
@@ -37,7 +37,7 @@ title = name = team.name
 <div class="col0">
     {% if team.is_approved in (None, False) %}
     {% if team.is_approved == None %}unreviewed{% endif %}
-    {% if team.is_approved == False %}rejected{% endif %} |
+    {% if team.is_approved == False %}closed{% endif %} |
     {{ team.homepage }}
     {% else %}
     <a href="{{ team.homepage }}">{{ team.homepage }}</a>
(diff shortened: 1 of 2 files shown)


whit537:
Ping @mattbk @kzisme @rohitpaulk et al.

www/%team/index.html.spt
```diff
@@ -37,7 +37,7 @@ title = name = team.name
 <div class="col0">
     {% if team.is_approved in (None, False) %}
     {% if team.is_approved == None %}unreviewed{% endif %}
-    {% if team.is_approved == False %}rejected{% endif %} |
+    {% if team.is_approved == False %}closed{% endif %} |
```
rohitpaulk: I'm not sure if it's used anywhere, but it would be cleaner to use the `status` property - https://github.com/gratipay/gratipay.com/blob/8e369153ba1d5f878fa2948653bc8e4a01a1d13d/gratipay/models/team.py#L83

whit537: Good find! I looked for that in the database schema, I didn't think to check for a computed property.

whit537: Fixed up in 2e9919bc6267c9f3d966078c8ee1a88a406dc9dc.

Issue #3602, opened 2015-07-06
Add project closing


whit537:
Reticketed from #3601.

##### Notify

- [x] https://gratipay.freshdesk.com/helpdesk/tickets/2750
- [x] https://gratipay.freshdesk.com/helpdesk/tickets/2645
- [x] https://gratipay.freshdesk.com/helpdesk/tickets/6453
- [x] https://gratipay.freshdesk.com/helpdesk/tickets/6575
- [x] https://github.com/gratipay/gratipay.com/issues/4286

mattbk:
Should teams whose applications have been denied be automatically removed at some point?

Should teams whose applications are under review be marked as such after creation and before being approved?

kaguillera:
@whit537 I am a bit confused about this ticket since https://github.com/gratipay/gratipay.com/pull/3601 was a PR to do this. If it is then shouldn't PR be closed? If not what else is required to close this ticket

mattbk:
> Should teams whose applications have been denied be automatically removed at some point?

We've been leaving them up since they are indicative of the types of teams that are approved. We can probably keep doing that even after Decouple, since we will retain "brand fit" as a review criterion.

JessaWitzel:
Ok. Am I right that there are no team log-ins, only contributor/owner log ins? If so, can only the owner close a team? Do we want a way to transition team ownership to another member of the team? Which makes me wonder, can a team have more than one owner?

mattbk:
> Am I right that there are no team log-ins, only contributor/owner log ins? 

Yes.

> If so, can only the owner close a team?

That's probably safest.

> Do we want a way to transition team ownership to another member of the team? 

That would be good to have, but that's probably worth doing in another PR. It's possible to change ownership by request, it just needs to be done manually in the database.

> Which makes me wonder, can a team have more than one owner?

Good question. What are the benefits to a team having more than one owner? Is it ability to edit the team? Should there be owners and also ~users who have team-editing privileges but not ability to close the team?

whit537:
@JessaWitzel Usually when I work on something like this I end up smashing around coding up a mess, just to wrap my head around the problem. Then I go back and produce several PRs methodically, starting from the inside out. In other words, the first PR probably modifies the database, the next one probably adds some Python APIs (and tests! and docs!). Then once there is Python API deployed, it's time for PRs that _use_ the API and expose that functionality in the UI.

[...]

PR #3680, opened 2015-08-13
actually, let's use rejected

whit537:
Sorry for the run-around on this, @mattbk @rohitpaulk et al. This PR backs out #3678. The reason is that I tried using the word "closed" ... somewhere (can't find it atm), and it sounded like doublespeak. Like, if we *did* reject someone but said the account was "closed" then that implies *they* closed their account—otherwise how are we going to distinguish that case? Let's use "rejected" for when we reject a team, and "closed" for when the owner closes the team, which could be before or after we review them. This PR reinstates "rejected", and we'll need #3602 for closing.

### DIFF
--- a/gratipay/models/team.py
+++ b/gratipay/models/team.py
@@ -82,7 +82,7 @@ class Team(Model):
     @property
     def status(self):
         return { None: 'unreviewed'
-               , False: 'closed'
+               , False: 'rejected'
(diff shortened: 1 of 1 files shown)


tshepang:
good decision; glad you acted on that instead of, you know, just letting it be

Issue #3654, opened 2015-07-30
Improve flow to attach PayPal when applying for a project


brylie:
Please add a link to instructions regarding how to add a bank account on the [team sign-up page](https://gratipay.com/new). This is a bit tricky to figure out.
##

whit537:
Sorry for the confusion, @brylie. What's worse is that we don't actually offer bank payouts anymore; the only option we support is PayPal (see #3567). Our error message is not just unhelpful. It's a lie. :-(

![screen shot 2015-07-30 at 8 06 24 am](https://cloud.githubusercontent.com/assets/134455/8982718/f1882142-3691-11e5-9a4e-7ef9d0a6df24.png)

You can attach a PayPal account [on your Gratipay settings page](https://gratipay.com/about/me/settings/).

brylie:
Thanks. PayPal is not a great option for me. I moved from the U.S. to Finland over a year ago, and PayPal will not let me connect my Finnish bank information. Perhaps some alternative funding sources would be advisable?

whit537:
@brylie It will take some time, but we eventually hope to [establish a proper anti-money laundering (AML) program](https://github.com/gratipay/inside.gratipay.com/issues/119), at which point we'll be able to approach vendors who would allow us to move money more directly (e.g. #417 #3366).

whit537:
+1 from https://github.com/gratipay/gratipay.com/issues/3859#issue-119580300

> @Gratipay To apply for a "team" (2.0???), I apparently need to link a PayPal account, but don't see the option to do so.

https://twitter.com/untitaker/status/670667868133019649

PR #3683, opened 2015-08-15
Remove mention of bank account on error 400

mattbk:
We only have payouts through PayPal right now, so in order to remove confusion we shouldn’t tell people to attach a bank account when they can’t.  https://github.com/gratipay/gratipay.com/issues/3654


whit537:
@mattbk The *.po translation files are updated automatically underneath [`release.sh`](https://github.com/gratipay/gratipay.com/blob/1874/release.sh). We shouldn't modify them directly here in this PR.

mattbk:
Okay, time to learn something new then.  
- Can I change the *.spt files, or does that mess with what messages the translation files are matched to?
- I should be able to undo the changes I made to the *.po files and keep it all in the same PR, right?

whit537:
> Can I change the *.spt files, or does that mess with what messages the translation files are matched to?

You should *only* change the *.spt files (not the *.po files). During [deployment](https://github.com/gratipay/gratipay.com/blob/1877/release.sh#L54-L62), we extract strings from the *.spt files themselves into `i18n/core.po`, upload that to Transifex, and then download completed 
*.po files from Transifex. This all happens in [these Makefile targets](https://github.com/gratipay/gratipay.com/blob/1877/Makefile#L94-L108). Bottom line: change *.spt files, not *.po files.

> I should be able to undo the changes I made to the *.po files and keep it all in the same PR, right?

Yes! Easiest would be to manually edit the *.po files, commit that change, and then push it up to the same branch (`error-400-bank-account`). How's your git-fu? Want/need more specific pointers than that? :)

mattbk:
I replaced the *.po files from master but left the *.spt files modified.  I think.

rohitpaulk:
!m @mattbk

whit537:
@mattbk Looks like we've got tests for those error messages. Wanna try updating those as well? :)

mattbk:
I think it worked?

whit537:
You win! :-)

whit537:
!m @mattbk

PR #3710, opened 2015-08-22
open the floodgates

whit537:
Advertise Team creation on the homepage! We have About pages now that explain it.

Issue #3704, opened 2015-08-21
spruce up the Team explorer


whit537:
Do #3703 first.

 - [ ] ~~Include images (#3701)~~
 - [ ] ~~put status in subnav~~
 - [ ] show all Teams to anon (right now anon can't browse or can't see unreviewed or something)
 - [ ] include cached values

whit537:
+1 from @timothyfcook at https://github.com/gratipay/gratipay.com/pull/3718#issuecomment-135768940: 

> you should feature the most interesting teams at the top of the list.

whit537:
> you should feature the most interesting teams at the top of the list.

I agree that the current sort order—Team creation time—is probably not the most interesting, especially since we don't list the creation time, so you can't even tell that it's sorted that way (it just seems random). But what *is* "most interesting"? Seems subjective. I think we should make the Team table sortable and filterable, so that people can follow their own interests in exploring Team data for themselves.

> can we not see total income of teams anymore?

Value caching is in a sorry state across the site. Now that this #3718 is landed, value caching is the next big thing I want to work on (per https://github.com/gratipay/inside.gratipay.com/issues/318#issuecomment-135585126).

mattbk:
A list of "trending" (most active) Teams would be another way to do it,  "Activity" would need to be defined.  See https://github.com/trending.

whit537:
Alright, spruced up in #3735, though plenty more to do:

 - sorting reticketed as #3736
 - status filtering reticketed as #3737
 - we already have images as #3701
 - paging reticketed as #3738
 - status icons reticketed as #3739

whit537:
Reviewing cached values for accuracy is #3729.

PR #3711, opened 2015-08-22
spruce up homepage

whit537:


whit537:
I'm not exactly sure where this PR came from or what I've done to it (I think I started a branch #3704 but ended up not using it and creating #3735 intead?), but the homepage is spruced up in #3735 ff. and the thing I turned this into is actually #3710. Closing.

Issue #3666, opened 2015-07-31
Although I receive money via Gratipay, I can't give it away


book:
Although I receive money via Gratipay, I can't give it away. Latest example: payday 165.

I started with a credit of $0.00, received more than enough to cover all the gifts I registered, but the amount was still sent to me in full (minus the PayPal fees) without making any of the donation I have listed.

I understand this is probably due to the way donations are resolved: my donation were taken from an empty account, and since I have no credit card listed, no money could be taken, therefore no donation was made in my name. Then the money given to my teams ended up on my account, and was paid to me.

I have no idea how payday computation is done at the moment, so I might be writing absurd and silly things below.

It should be possible to come up with an algorithm that would 1. allow me to give some of the money I receive through Gratipay to teams on Gratipay 2. minimize the total amount lost in fees.

By computing the graph of all people and teams (nodes) with the donations (edges) between them, one could come up with a "delta" for each person and team. The only issue would be people without a registered mean of payment, when the delta is negative. Some donations would not be honored, which would require another run of the graph computation, with the new data. I expect this would stabilize quickly.
##

book:
Looking at [payday.py](https://github.com/gratipay/gratipay.com/blob/master/gratipay/billing/payday.py), I see [this comment](https://github.com/gratipay/gratipay.com/blob/master/gratipay/billing/payday.py#L65-L68):

> On Payday, we want to use a participant's Gratipay balance to settle their tips due (pulling in more money via credit card as needed), but we only want to use their balance at the start of Payday. Balance changes should be atomic globally per-Payday.

So indeed, one has to have a positive balance at the start of Payday, otherwise money will be drawn from their credit card. I haven't looked at the code in depth, but I expect that if the balance at the end is positive, no money is actually withdrawn from their credit card. (Otherwise, the only real beneficiaries of the system are PayPal and the banks...)

That algorithm won't work for people who have not set up a payment mean, since they can't make a "virtual cash advance" to the system. One option is to clearly state that it's only possible to make gifts when a valid payment mean has been set up (issue solved with no code change!). However, if you want it to be possible to give from the stack of money received within the system, that probably means a process with multiples passes, which hopefully has a stable end state.

[...]

PR #3721, opened 2015-08-29
rationalize cached values

whit537:
Follow-on from #3599.


whit537:
Picking up from https://github.com/gratipay/gratipay.com/pull/3599#issuecomment-126562308, here is where we are and where we're going:

```
                sum           count        sharing bit
                ----------------------------------------------
current  ~user  giving        —            anonymous_giving
                receiving     npatrons     anonymous_receiving
                taking        —            anonymous_receiving

         Team   receiving     nsupporters  —
                payroll       nmembers     —       

target   ~user  giving        nrecipients  anonymous_giving
                taking        nteams       -

         Team   receiving     ngivers      —
                distributing  nmembers     —
```

The one change here from the original comment is that I've dropped `anonymous_taking`. There is no anonymous taking in Gratipay 2.0. Payroll is public, as is the numbers relative to Teams.

whit537:
Here's the diff:

 - [ ] remove `participants.npatrons`
 - [ ] remove `participants.receiving`
 - [ ] remove `participants.anonymous_receiving`
 - [ ] add `participants.nrecipients`
 - [ ] add `participants.nteams`
 - [ ] rename  `teams.nsupporters` to `teams.ngivers`
 - [ ] rename  `teams.payroll` to `teams.distributing`

For each:

 - [ ] ALTER TABLE
 - [ ] add/modify code and tests
 - [ ] UI

whit537:
@timothyfcook Also, the punched penny very nearly aligns with the leaf line, which is not vertical or otherwise related to the rest of the grid on the site (this new design has a 3-column grid, based on 192px columns). To me that symbolizes the way that Gratipay marches to the beat of a different drummer. We are aligned with *something*, but it's certainly not the mainstream. We are [building a new inside](https://github.com/gratipay/inside.gratipay.com/issues/318#issuecomment-135622366). According to the penny (~= Gratipay ~= each of us), it's the rest of the page that is off kilter. What, Gratipay asks, is one's frame of reference?

Also, that is a picture of the table that I am sitting at ***RIGHT NOW***. I bought this table at Capo's in New Brighton 15+ years ago, and refinished it with a friend. I bought it as a dining room table for my wife before we were married, while we were engaged. I'm looking at the knot ***RIGHT NOW***.

whit537:
FUCK. Wrong ticket again, sorry. :-) :full_moon_with_face:

whit537:
I think as part of this that we should move away from estimating dollar amounts (receiving, taking) based on the future, and instead base it on the past. Why? The past is more predictable than the future. :-)

This may well obviate `is_funded` (though does payday depend on that?).

I think we should cache last week's actual amount, and an average of the past four weeks. These two numbers will provide a fine answer to the question, "How much will my Team receive this week?"

whit537:
Basing `receiving` on the past instead of the future also lets us avoid the gymnastics we're currently doing to simulate payday every time someone changes their credit card or whatever. It's going to be simpler.

whit537:
I guess we can just update all of the numbers once at the end of each payday, rather than having to update them constantly throughout the week in response to various events (payment instruction changes, route changes).

whit537:
The downside is that one won't be able to watch drastic movement that happens inside a week. Especially [early on](https://medium.com/gratipay-blog/resentment-2c621dbd7541#53fb), we paid a lot of attention to intra-week movement. I think that Gratipay has matured, and we are ready to lengthen our attention span. Gratipay is intended to inculcate the long view. You'll get to see this week's amazing growth ... on Thursday.

whit537:
How about `taking` and `distributing`? Those depend somewhat on the income that a Team receives, but are largely in the control of Team members. Those numbers aren't as squidgy as `receiving`. Should we refer `taking` and `distributing` to the future instead of the past? Should we base those on Team members' `payroll_instructions` rather than on past `payments`?

rohitpaulk:
> though does payday depend on that?

No, it doesn't. :) 

I don't think going back on cached values is a good idea though... When I tip someone, I expect their receiving number to increase. Similarly if I was a team, I'd like to know an estimate of how much I'm going to receive. This stuff is **very** useful, @whit537.

whit537:
> When I tip someone, I expect their receiving number to increase.

Hmmm ... that's hard to argue with. :)

whit537:
I'm working through the tests here and realizing that we have a lot of skips and a lot of JSON endpoint tests that are obsoleted by 1.0 and the move away from tips. I'm finding that I want to cut out all that cruft before proceeding here.

whit537:
Starting over with an ordered approach: removing `anonymous_receiving`.

whit537:
I.e., taking each diff item in turn, one at a time.

whit537:
The remaining test failures get us into @book's questions on #3666.

whit537:
To keep this PR under control I think we should have the cached values reflect whatever the current behavior of payday is. If we want to change the behavior below #3666 then that'll be a separate PR.

whit537:
From https://github.com/gratipay/gratipay.com/issues/3399#issuecomment-101470445:

> It *looks* like throwing a `Team` abstraction into payday makes the algorithm *much* nicer. It seems like it will enable us to easily settle the graph completely each cycle, *and* have no escrow!

This is misleading. My original exploration assumes that `a` has a credit card attached. Without a card attached, we would still need to perform an iterative graph settling operation to enable people to pay entirely out of money taken from a Team in a given week.

whit537:
Here's a reset script to run after deployment:

https://gist.github.com/whit537/07ffabf7f110c60911d8

It's idempotent.

whit537:
I've been working on the UI. It seem like we should be able to use the rest of the blank space in the header (above the ~username/Team Name), but the tiny white on white that we've had is low-contrast, especially for how small it is. How about something like this?

screen shot 2015-08-30 at 3 51 03 am

screen shot 2015-08-30 at 3 54 44 am

whit537:
I've extended that pattern to the admin bits of the profile UI.

# Closed Accounts

### Admin

screen shot 2015-08-30 at 3 56 33 am

### Auth

screen shot 2015-08-30 at 3 58 32 am

### Anon

screen shot 2015-08-30 at 3 59 08 am

whit537:
# Absorbed Accounts

## Admin

screen shot 2015-08-30 at 4 02 48 am

## Auth

screen shot 2015-08-30 at 4 03 09 am

## Anon

screen shot 2015-08-30 at 4 03 34 am

whit537:
# Suspicious Accounts

Admin, auth, and anon:

screen shot 2015-08-30 at 4 04 59 am

screen shot 2015-08-30 at 4 05 37 am

screen shot 2015-08-30 at 4 06 04 am

whit537:
A few points:

 - We have clean ~user numbers again.
 - We have Team numbers for the first time ever! :-)
 - I'm trying out the word "shares" instead of "distributes."
 - I find myself wanting to distinguish between the amount that a Team owner takes, and the amount that the Team shares with everyone else—which is zero across the board right now since we don't have payroll.

whit537:
I'm gonna make this responsive and then land it.

whit537:
Is the javascript dynamically updating amounts properly?

whit537:
No! :-)

whit537:
I'm also checking the pricing page and the global cta ...

whit537:
Okay! Waiting for Travis ...

whit537:
:weary:

whit537:
Code and schema deployed, running the cache refresh script ...

whit537:
Done! :-)

screen shot 2015-08-30 at 5 12 01 am

mattbk:
!m @whit537!

whit537:
Thanks. :-)

mattbk:
I had a question about the receiving amount, but this explains it:

>I think we should cache last week's actual amount, and an average of the past four weeks. These two numbers will provide a fine answer to the question, "How much will my Team receive this week?"

My one supporter is no longer supporting, according to History.

whit537:
Here's what we ended up with ftr:

```
       sum           count             sharing bit
       ------------------------------------------------
~user  giving        ngiving_to        anonymous_giving
       taking        ntaking_from      —

Team   receiving     nreceiving_from   —
       distributing  ndistributing_to  —
```

Issue #2417, opened 2014-05-18
Payment made, but *not* deducted from user account?


pjf:
Opening this for [FreshDesk 459](http://gittip.freshdesk.com/helpdesk/tickets/459). We have a user who has reported that they've received funds from Gittip, but these funds were _not_ deducted from their Gittip account. I've confirmed that this has indeed happened.

I very this as a critical thing to investigate before the next payday run, as it may be that gittip is (or has been) leaking money, which is very, very bad.

Attn @patcon and @whit537 at the very least.

PR #3730, opened 2015-08-31
fix stats

### LINKED ISSUE #3446: clean up stats page
Reticketed from #3415.

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/15126562-clean-up-stats-page?utm_campaign=plugin&utm_content=tracker%2F85909&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F85909&utm_medium=issues&utm_source=github).
</bountysource-plugin>

whit537:
Closes #3446.


whit537:
The next step I'm working on is to add a `payday` column to the `transfers` table, so that we can recompute `volume` and `nactive` in a similar way both pre- and post-Gratipocalypse. However, I have discovered a single record in the transfers table where the amount is negative. We have a `positive` check constraint now which is catching the error. I'm not sure how the record ended up in there in the first place. It happened in May of last year. The ~user in question has a balance, and the amount is only 33¢. I'm inclined to correct the error using the user's balance.

whit537:
The user in question was involved in a MassPay snafu: https://github.com/gratipay/gratipay.com/issues/2417.

whit537:
Alright, so we have this negative-`amount` record in `transfers`, and the bottom line for this ticket is that that record is not going way. I propose to work around it by relaxing the positive check constraint on `amount`, performing the update, and then reinstating the check constraint.

P.S. The way we ended up with the negative transfer is complicated: our 1.0 payroll implementation computed the amount to transfer to the last person taking as the remainder of the team's balance, and as a result of the MassPay snafu in #2417, the team had a negative balance, so therefore the last person's take was negative!

whit537:
Positive constraint is in #2723, btw.

whit537:
Okay, got it. Looks like we have 563 non-payday transfers (take-over, final-gift), and 523,762 payday transfers.

whit537:
Actually, we don't need to reset the pre-Gratipocalypse values, though we may as well add the `payday` column to `transfers` since I've got that ready now.

whit537:
I'm seeing a volume spike in [159](https://github.com/gratipay/gratipay.com/issues/3566). What's that about?

screen shot 2015-08-31 at 12 14 53 pm

whit537:
> What's that about?

It was a user with multiple Teams consolidating their funds in one Team as part of getting paid out for the first time.

https://gratipay.freshdesk.com/helpdesk/tickets/2090

whit537:
Okay! We have accurate volume and active user charts again! :-)

screen shot 2015-08-31 at 12 23 31 pm

whit537:
Winding down here, ready for review if anyone is around. :)

whit537:
screen shot 2015-08-31 at 1 29 50 pm
screen shot 2015-08-31 at 1 29 59 pm

whit537:
Ready?

PR #3732, opened 2015-08-31
tinker with labels on detail tables

whit537:
Now it's readable as "weekly giving," etc.


whit537:
### Before

screen shot 2015-08-31 at 1 42 55 pm

### After

screen shot 2015-08-31 at 1 43 16 pm

PR #3744, opened 2015-09-02
make a 1.0 payout form

whit537:
Closes #3743. Currently this is stubbed out with an "email us" (https://github.com/gratipay/gratipay.com/issues/3539#issuecomment-136907133).

PR #3454, opened 2015-05-18
close the close loophole for payouts

whit537:
https://github.com/gratipay/gratipay.com/issues/3453#issuecomment-103073394
https://github.com/gratipay/gratipay.com/pull/3392#issuecomment-100435765

Issue #3453, opened 2015-05-18
reverse eight accidental payouts


whit537:
During last week's payday (#3415; the first under Gratipay 2.0), we [accidentally paid out](https://github.com/gratipay/gratipay.com/issues/3415#issuecomment-102181981) to eight people that we shouldn't have. It hurts, but I think we need to issue reversals of these payouts. :-(

Let's try to notify the users privately via email.

whit537:
Bah. We've also got people closing their accounts, and withdrawing that way (cf. https://github.com/gratipay/gratipay.com/pull/3392#issuecomment-100435765). :-(

whit537:
It feels to me like a bridge too far to reverse bank payouts on closed accounts.

whit537:
Two of the eight failed anyway.

whit537:
I've issued reversals for the six successful accidental payouts last week. They will four or five days to process. Now to notify users ...

whit537:
Upon reviewing history for the six users, I would actually expect two more of these payouts to fail (they've been failing in the recent past). That leaves four accidental successful payouts.

whit537:
(One of which was to me, btw. :-)

whit537:
Accidentally sent before I was finished (how do I turn on that "undo" feature in Gmail?), but thankfully only to @webmaven:

>Greetings!
>
You are one of four users to whom Gratipay accidentally made a payout last week.
>
We're in the midst of [a major transition](https://medium.com/gratipay-blog/gratipocalypse-42fd0ec0d9e8), during which we're changing our terms of service for receiving money on Gratipay. Our goal is to be able to bring you on under our new terms and send you your money, but we're not quite ready to do that yet, so unfortunately we've had to reverse last week's payout. :-(
>
chad

whit537:
From: support@gratipay.com
To: support@gratipay.com
Bcc: [4]
Subject: erroneous payout

>Greetings!
>
You are receiving this message because you are one of four users to whom Gratipay made an erroneous payout last week.
>
We're in the midst of [a major transition](https://medium.com/gratipay-blog/gratipocalypse-42fd0ec0d9e8), during which we're changing our terms of service for receiving money on Gratipay. Our goal is to be able to bring you on under our new terms and send you your money, but we're not quite ready to do that yet, so unfortunately we've had to issue a reversal for last week's payout. Sorry. :-(
>
[...]

PR #3747, opened 2015-09-02
compute 1.0 balances

whit537:
Needed for #3744.


whit537:
I want to write one SQL script that computes all 1.0 balances, and then run it.

whit537:
We need to compute 1.0 balance as a function of current balance. Actually, "1.0 balance" is really "*remaining* 1.0 balance," i.e., the amount received under 1.0 that is still in our escrow.

whit537:
My hunch is that the function could be some summation of `transfers` and `payments` and `exchanges` since `2015-05-07`.

whit537:
Your remaining 1.0 balance is your balance at the Gratipocalypse, minus anything you gave away (`transfers` and `payments`) or withdrew (`exchanges`) since then. The payins and receipts since the Gratipocalypse are irrelevant.

whit537:
Counting that way, we are at 2,173 participants and a total of $130,112.10.

whit537:
Alright, so ... ideally we would have not had any 1.0 leaks, and we would be paying out in full every week under Gratipay 2.0, so that it was guaranteed that for anyone with a balance, their balance *was* their Gratipay 1.0 balance.

However, we a) leaked 1.0 money, and b) haven't paid out everyone every week under Gratipay 2.0. What are the details with those cases?

whit537:
We leaked 1.0 money in two ways:

 - We paid some of it out (#3453). We recovered all of this money by reversing those transactions.
 - We transferred some of it to other participants as final gifts (#3454). We did not reverse these.
 - We transferred a tiny bit of it from two participants to a third as part of a take-over. We did not reverse these.

whit537:
The tiny bit we transferred from two participants to a third as part of a take-over is fine, because it's the same person, and so they'll have to apply for a payout in order to get the money anyway.

The amount we leaked as final gifts is effectively still 1.0 money, it's just that the participants that received it will have to apply for a payout to extract it instead of the participants who were holding that money on the day of the Gratipocalypse.

It's like hot potato. :-) :fire: :potable_water: 

The paid-out money, we recovered. So that's fine.

whit537:
The fourth way we leaked 1.0 money was with 2.0-style payments, and this is the most significant way. Of the 2,259 participants who received money under Gratipay 1.0, 215 (9.5%) have made voluntary payments under Gratipay 2.0. These payments decremented their Gratipay 1.0 balance.

whit537:
My next step here is to frame the above in terms of the columns in the `numbers_1_0` table.

whit537:
> These payments decremented their Gratipay 1.0 balance.

And, more importantly, they left the system, because the recipients were cleared for payouts.

whit537:
Reticketed immediately refunding failed payouts as #3754. We want it to be the case that the only money in escrow is 1.0 money, or 2.0 money that is below the payout threshold.

whit537:
Only 103 out of 2,209 (4.7%) have a current balance that is not equal to their remaining 1.0 balance. Only 241 (10.9%) have a balance that has moved at all since the Gratipocalypse. I think for the purposes of the 1.0-payout page (#3744) we can treat current balance as the 1.0 balance. Most people with a balance under $10 aren't going to apply for a payout, and/or are going to know that they're giving. We *should* refund the charged-in-advance money, though.

whit537:
I ended up merging this branch into #3744 as 1cc606558c7a7f10f755226d969bf48040fa724f.

PR #3766, opened 2015-09-10
pick off a few more emoji

whit537:
Left over from #3765.

PR #3775, opened 2015-09-11
fix regression in status icons on Team pages

whit537:
## Was

screen shot 2015-09-10 at 9 56 15 pm

## Now

screen shot 2015-09-10 at 9 56 19 pm


whit537:
Missed this in #3766. Ping @rohitpaulk et al. :)

PR #3790, opened 2015-09-17
update user review dashboard for Braintree

whit537:
See https://github.com/gratipay/inside.gratipay.com/issues/344#issuecomment-141263031.


whit537:
Gonna self-merge once Travis runs green.

Issue #3763, opened 2015-09-08
refund advances


whit537:
Reticketed from #3675.

rorepo:
@whit537, following on from my earlier questions, it's clear that we would no longer have the scenario of charging a user and then immediately having to refund the same user. And possibly, the extent of the 'small balance' situation I'd been thinking of will have diminished after a few weeks of charging in arrears. 

There's something else I'd like to clarify. Would refunds be made as payments back to the same credit card? Or to the paypal account of the user? If it's the latter, what do we do in the case of users who don't have paypal accounts? e.g. someone who is only giving and hasn't bothered to set up a paypal account?

whit537:
> Would refunds be made as payments back to the same credit card?

Yes. Balanced and Braintree have refund APIs that work even if the card was cancelled—though almost certainly not if the entire bank/card account was closed, a situation we'll almost certainly run into.

rorepo:
Having a spot of trouble. When running ```make test```, things getting stuck at **recreating public schema ...**. Doesn't seem to have anything to do with my code changes - trying with a fresh checkout doesn't help. Wonder if my machine's got corrupted.

Also, a couple of questions. Found a ```MINIMUM_CREDIT``` value in ```exchanges.py```. What is this meant for? And do Braintree levy a fee on credits as well?

rorepo:
I've moved beyond the 'corruption' (was my code after all :-(), but now facing a seemingly intractable problem. I think possibly something to do with my (lack of) understanding of ```threaded_map/thread pool``` usage. Trying to figure it out.

whit537:
@rorepo Possibly related to https://github.com/gratipay/gratipay.com/issues/3789?

whit537:
> I've moved beyond the 'corruption'

Cool. :)

> I think possibly something to do with my (lack of) understanding of threaded_map/thread pool usage. Trying to figure it out.

How can I help? :-)

> Found a `MINIMUM_CREDIT` value in `exchanges.py`. What is this meant for? 

This is vestigial from when we did bank payouts via Balanced. It can probably be removed now.

> And do Braintree levy a fee on credits as well?

Braintree doesn't provide credits (payouts). The payout route we offer is PayPal.

> it's clear that we would no longer have the scenario of charging a user and then immediately having to refund the same user.

Actually, if a payout fails, we *do* want to immediately refund all of the cards the money came in on. No?

rorepo:
[...]

PR #3802, opened 2015-10-02
Refund excess balances

rorepo:
This is for a preliminary review - have a few issues and questions. 

Got things working with my credentials, with ```tests/local.env``` and after adding the custom field participant_id in settings. 'Working' only in the sense that I get ```success``` and ```authorized```, though. When I tried checking on the dashboard, things are less clear - all transactions are shown as 'sale' and some say 'voided', 'submitted for settlement', 'gateway rejected', etc. 
The transaction summary page shows credit values as 0.

I've got the functionality worked out, but am facing the following issues:

1. A rather quirky problem. The new method ```test_refund_of_excess_balance``` that I added to ```test_billing_payday.py``` works as expected, but not along with the existing method ```test_payday_preserves_due_until_charged```. If both methods run, the credit part of ```test_refund_of_excess_balance``` fails with an authorisation error. I see a small possibility that this behaviour may be something to do with my local environment.
2. I managed to get things running with my own Braintree credentials, using ```tests/local.env```. I get 'success' and 'authorized' responses. However, things do not look ok when checking on the Braintree dashboard - credit in the transaction summary is always 0.

In addition, I have the following questions:

1. For the moment, I'm just looking for ```balances > 0``` to do refunds. I believe I will need to apply some other conditions e.g. to exclude 1.0 balances?
2. I've uploaded changed test fixtures as well, but since I'm now working with my own Braintree credentials, will these be relevant?

### DIFF
--- a/gratipay/billing/exchanges.py
+++ b/gratipay/billing/exchanges.py
@@ -83,6 +83,68 @@ def repr_exception(e):
         return repr(e)
 
 
+def create_card_credit_hold(db, participant, amount):
+    """Create a hold on the participant's credit card for a refund.
+
+    Amount should be the nominal amount.
+
+    """
+
+    typecheck(amount, Decimal)
+
+    username = participant.username
+
+    # Perform some last-minute checks.
+    # ================================
+
+    if participant.is_suspicious is not False:
+        raise NotWhitelisted      # Participant not trusted.
+
+    route = ExchangeRoute.from_network(participant, 'braintree-cc')
+    if not route:
+        return None, 'No credit card'
+
+    # Go to Braintree.
+    # ================
+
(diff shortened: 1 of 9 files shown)


whit537:
> I believe I will need to apply some other conditions e.g. to exclude 1.0 balances?

Yeah, this gets complicated. See https://github.com/gratipay/gratipay.com/issues/3763#issuecomment-144880568.

> I've uploaded changed test fixtures as well, but since I'm now working with my own Braintree credentials, will these be relevant?

My expectation has been that we'll handle #3763 and #3539 with one-off scripts. I'm not sure what general-purpose long-lived code changes will be necessary for those.

whit537:
Closing post-#3810.

!m @rorepo

Issue #3260, opened 2015-03-17
upgrade to Cedar-14


whit537:
>!   Cedar-10 will reach end-of-life on November 4th, 2015.
 !   Upgrade to Cedar-14 at your earliest convenience.
 !   For more information, check out the following Dev Center article:
 !   https://devcenter.heroku.com/articles/cedar-14-migration

whit537:
Getting that message during deployment.

whit537:
>On November 4th 2015, the Cedar-10 stack will reach its end-of-life. Cedar-14 is Heroku’s most modern stack and when we launched it one year ago, [Cedar-10 was deprecated](https://hello.heroku.com/e/36622/changelog-items-550/5y839g/481792037). After November 4th, the Cedar-10 stack will no longer receive security updates. You must migrate your app to Cedar-14 to ensure that it's running in a secure environment and we recommend that you do so as soon as possible.
>
Heroku is ending security support for Cedar-10 because the Ubuntu operating system that the stack is based on is [no longer maintained by Canonical](https://hello.heroku.com/e/36622/-end-of-life-on-april-30-2015-/5y839j/481792037). We take security very seriously, and we know you do too, so we want to help you migrate to a stack that is more secure.
>
The following apps which have received traffic recently, that you own or are collaborated on, currently run on Cedar-10:
>
shields-io, aspen-io, roobot-prod, img-shields-io, motivate-im, inside-gratipay-com, gratipay, gratipay-access-dashboard
>
To migrate an app, run the following commands:
> 
$ heroku stack:set cedar-14
$ git commit --allow-empty -m "Upgrading to Cedar-14"
$ git push heroku master
>
For details, see the [Cedar-14 migration guide on Dev Center](https://hello.heroku.com/e/36622/articles-cedar-14-migration/5y839l/481792037).
>
If an app has not been migrated by November 4th, Heroku will start a migration process for Cedar-10 apps, starting with free apps. Since there’s a slight chance your app won’t build correctly on Cedar-14 or that it performs in a subtly different way, you should not rely on the migration process. Instead, you should migrate apps yourself, to catch any errors or problems. If you have any questions, please let us know.
>
Happy coding,
>
Heroku

whit537:
That was https://gratipay.freshdesk.com/helpdesk/tickets/2939.

rohitpaulk:
Tried upgrading, got heroku's application error page soon after. Rolled back, checking for errors.

[...]

PR #3815, opened 2015-10-11
Use local pip packages on heroku

rohitpaulk:
https://github.com/gratipay/gratipay.com/issues/3260

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -23,8 +23,8 @@ env: requirements.txt requirements_tests.txt setup.py
 				--extra-search-dir=./vendor/ \
 				--always-copy \
 				./env/
-	$(pip) install --no-index -f ./vendor/ -r requirements.txt
-	$(pip) install --no-index -f ./vendor/ -r requirements_tests.txt
+	$(pip) install --no-index -r requirements.txt
+	$(pip) install --no-index -r requirements_tests.txt
 	$(pip) install -e ./
 	touch env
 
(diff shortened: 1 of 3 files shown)


rohitpaulk:
I don't know if there's a better way to handle this, but I've tested this on a new heroku app and it works.

whit537:
What's the motivation here? What indicates that we *aren't* using local packages already? The `-f ./vendor/` doesn't give us that?

whit537:
I guess this is an indication ... :-)

screen shot 2015-10-12 at 9 00 08 am

rohitpaulk:
Yep, I found that over at https://github.com/gratipay/gratipay.com/issues/3260#issuecomment-147205823

rohitpaulk:
> The -f ./vendor/ doesn't give us that?

It does, but we don't get to choose how heroku does `pip install`. I left a comment - https://github.com/gratipay/gratipay.com/issues/3260#issuecomment-147209248

> So heroku says -
> ```To add a local dependency in requirements.txt, specify the relative path to the directory containing setup.py:
> ./path/to/distribution```
> 
> ^ It's either that, or we maintain a custom buildback that instructs pip to fetch packages from `vendors`. First option sounds better to me, trying it out..

whit537:
Gotcha, I had missed #3260, sorry.

techtonik:
This is basically a revert of https://github.com/gratipay/gratipay.com/pull/3292 which was needed to install platform-dependent wheels.

Issue #3779, opened 2015-09-12
give more visibility into funding build-up


whit537:
Now that we're charging in arrears, we need to give both ~users and Teams visibility in their funding build-ups. Reticketed from #3378.
##

whit537:
Here's what I'm currently seeing as the build-up for nuvola-player, e.g.:

```sql
=> select sum(due), count(*) from current_payment_instructions where due > 0 and team='nuvola-player';
┌───────┬───────┐
│  sum  │ count │
├───────┼───────┤
│ 79.18 │    59 │
└───────┴───────┘
(1 row)
```

However, I'm not sure how exactly this interacts with the `is_funded` column on `current_payment_instructions`.

rohitpaulk:
There are two parts to this - the ~users UI, and Teams UI. The second part is blocked on https://github.com/gratipay/gratipay.com/issues/3705.

PR #3817, opened 2015-10-11
Expose dues in UI for ~users

rohitpaulk:
https://github.com/gratipay/gratipay.com/issues/3779

### DIFF
--- a/gratipay/models/participant.py
+++ b/gratipay/models/participant.py
@@ -921,6 +921,7 @@ class Participant(Model):
                 SELECT DISTINCT ON (pi.team)
                        pi.team  AS team_slug
                      , pi.amount
+                     , pi.due
                      , pi.ctime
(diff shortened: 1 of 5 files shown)


rohitpaulk:
Before: 
screen shot 2015-10-11 at 11 42 22 pm

After:
screen shot 2015-10-11 at 11 34 57 pm

The `due` column is only shown for someone who does have dues - I thought it'd be confusing for users who give more than $10 per week to always see $0.

techtonik:
What is `due`?

rohitpaulk:
@techtonik - Earlier we used to charge in advance (Charge $10 at the beginning for a $1/wk subscription, and slowly drain it over the subsequent weeks). Now, we're charging in arrears (Keep adding to the `dues` every week, and charge $10 when the dues reach $10).

Issue #3828, opened 2015-10-16
invalid route_id on record an exchange


whit537:
@rohitpaulk My hunch is that this is due to a recent PR of yours ... ring any bells?

![screen shot 2015-10-15 at 8 52 29 pm](https://cloud.githubusercontent.com/assets/134455/10531364/b4bdadf2-737e-11e5-94a3-25d567e627d6.png)

whit537:
Ah, hrm. I see that we now list the routes available on the "record an exchange" form. The issue here is that I'm trying to do a payout via Transferwise, which doesn't have proper route infrastructure yet.

rohitpaulk:
@whit537 - Can this be closed now?

whit537:
Yes. :)

PR #3829, opened 2015-10-16
Add a couple additional payment networks

whit537:
Transferwise is needed for https://github.com/gratipay/gratipay.com/issues/3828.
Both are needed for https://github.com/gratipay/gratipay.com/issues/2779.


whit537:
Ping @rohitpaulk @rorepo et al.

PR #3830, opened 2015-10-16
enable 1.0 payouts indefinitely

whit537:
To complement https://medium.com/gratipay-blog/39c07bd031b.

PR #3836, opened 2015-10-19
Extend 1.0 payout deadline 100 years

mattbk:
We published a blog post that says you can still withdraw 1.0 money, but this page has a deadline built into it.  As a stopgap, extended deadline to Oct 2, 2115.


whit537:
@mattbk Fuller solution is in #3830.

PR #3823, opened 2015-10-12
Fix usage of db/cursor in team.update_receiving

rohitpaulk:
https://github.com/gratipay/gratipay.com/issues/3811#issuecomment-147529226

PR #3842, opened 2015-10-21
remove branch.py

whit537:
This is from #3823.

@rohitpaulk Do we still need to run this before removing it?


rohitpaulk:
> Do we still need to run this before removing it?

Oops - forgot bout that. Yes we have to, doing it now.

rohitpaulk:
Done.

whit537:
!m @rohitpaulk

PR #3843, opened 2015-10-22
add email prefs for "News & Announcements"

whit537:
Addresses root problem behind https://github.com/gratipay/inside.gratipay.com/issues/380. Will result in an additional checkbox under `/~user/emails/`:

screen shot 2015-10-21 at 5 01 00 pm


whit537:
Should this be opt-in or opt-out? Opt-out, I think ... and! we should take care to manually opt out those who indicated as much underneath https://github.com/gratipay/inside.gratipay.com/issues/235.

whit537:
Interesting feedback on [FD3284](https://gratipay.freshdesk.com/helpdesk/tickets/3284):

> I interpreted both as extremely short, minor updates:
>
Gratipocalypse
 "We have an unfortunate announcement to make: after 152 weeks of continuous operation, we are for the first time today skipping Gratipay's weekly payment cycle. We are deeply sorry. :-( "
>
Gratipay, Year Three
 "Good news! We have returned to processing as Gratipay 2.0, and are rebuilding on a much more solid foundation ... just in time for our third birthday!"
>
I never realized they both linked to posts that contained such important news

Two options I see for addressing this:

 - Include the full text of the announcement in the email.
 - Add a button in the email with a call to action like "Read Full Announcement"

mattbk:
The Gratipay blog has a history of not being updated even if there is new information (which is fine*), so including the full text is probably not a bad idea, at least for the most important things.  If we follow the format we've been using recently of broad strokes followed by details, maybe just putting the broad strokes into the emails is enough.

This is one of the areas where tracking reads, clickthroughs and unsubscribes per email may help guide the number of messages sent.

*Some people constantly update their old blog posts, which I find annoying.  I'd rather see an editor's note at the top of blog posts that are now incorrect in some way, but leave the original text unchanged in most cases.

whit537:
> The Gratipay blog has a history of not being updated even if there is new information

"[#BalancedShutdown](https://medium.com/gratipay-blog/balancedshutdown-3a9fdb32e9c6)" and "[Gratipocalypse]()" have updates. I've added one to "[$130,000](https://medium.com/gratipay-blog/140-000-97d54e6d40ed)". What else should have an update?

whit537:
> If we follow the format we've been using recently of broad strokes followed by details, maybe just putting the broad strokes into the emails is enough.

Okay, let's keep this in mind when it comes time to send out our first "News & Announcements" email.

mattbk:
> "#BalancedShutdown" and "Gratipocalypse" have updates. I've added one to "$130,000". What else should have an update?

No, those make sense per my footnote.

whit537:
> No, those make sense per my footnote.

Okay, but what did you mean by, "The Gratipay blog has a history of not being updated even if there is new information"?

whit537:
More from [FD3284](https://gratipay.freshdesk.com/helpdesk/tickets/3284):

> A button for the 'full announcement' might work, maybe for these specific emails the short summary that was included could have been a bit clearer eg. "*Our payment processor is going out of business and finding a new one raises a few concerns that takes a lot of hard work. This might result in some of you will have to be excluded*" for the first one.

mattbk:
> Okay, but what did you mean by, "The Gratipay blog has a history of not being updated even if there is new information"?

Don't worry about it.  It's okay.

whit537:
:-)

mattbk:
How's the status on this?

whit537:
18+ months, closing. 😞

PR #3844, opened 2015-10-22
repair schema sql

whit537:
I discovered that we apparently never applied these last two `branch.sql`s(?!), and when I went to apply them I hit a syntax error with the first, and another error with the second (`ERROR:  ALTER TYPE ... ADD cannot run inside a transaction block`). I've run these modified commands against production.


whit537:
Ping @rohitpaulk et al.

PR #3860, opened 2015-12-01
Show dues in team pages

rorepo:
Changes to show ```due``` values, if any, on the teams list and individual team pages.

The value shown is from a new column ```due``` in table ```teams```, updated with the summary of values from table ```payment_instructions```.

### DIFF
--- a/gratipay/models/participant.py
+++ b/gratipay/models/participant.py
@@ -1013,15 +1013,17 @@ class Participant(Model):
                AND p.id != %(id)s
         """, dict(username=self.username,team=team,id=id))
 
-        # Update summary value in teams
+        # Update summary values in teams
         (cursor or self.db).run("""
             UPDATE teams t
-               SET due = pi.due
-              FROM (select team, sum(due) as due
+               SET due = pi.due, ndue = pi.ndue
+              FROM (select team, sum(due) as due, count(*) as ndue
               FROM payment_instructions
+             WHERE due > 0
           GROUP BY team) as pi
              WHERE t.slug = pi.team
-        """)
+               AND t.slug = %(team)s
+        """, dict(team=team))
 
     def _reset_due(self, team, except_for=-1, cursor=None):
         (cursor or self.db).run("""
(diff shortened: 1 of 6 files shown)


whit537:
CODE!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!! :heart_eyes:

whit537:
@rorepo Did you forget to commit `branch.sql`?

rorepo:
Oops! Done now :)

whit537:
!m @rorepo

whit537:
This PR inspired https://github.com/gratipay/gratipay.com/issues/3862. :-)

whit537:
I assume this is ready for review, ya?

whit537:
screen shot 2015-12-02 at 10 41 04 am

A few things:

 - Let's show `ndue` along with `due`. Looks goofy to have that cell blank.
 - On the Team page, let's show "Due" always, even if it's zero, as we do for the other values.
 - Do we need to backfill `due`? It's zero for everyone. Should it be?

whit537:
Also:

 - Let's tweak the vertical whitespace on the homepage tables so it fits better now that it has three lines.

rorepo:
Not sure about the backfill bit, but a thought here - this column is now being populated in ```participant``` ```_update_due```, called within ```set_payment_instruction``` - perhaps it would be a good idea to include this in the payday run as well.

Made all the other changes suggested.

rohitpaulk:
I'm :-1: on this. 'Dues' is not a value that we should be displaying across the leaderboard/general UI. I think it only makes sense to expose this on team specific pages (the receiving/history page, in particular). 

The cached value for 'receiving' and 'nreceiving_from' already factors in the 'dues' component.

whit537:
> 'Dues' is not a value that we should be displaying across the leaderboard/general UI.

Fair point. The number we *really* want to show on the leaderboards is `sharing`, but that's a long way off.

You okay to close this, @rorepo? Or do you want to make a case for displaying "due" in this way instead of the proposal on #3873?

rorepo:
Happy to defer to the old-timers :-)

Will close this. I take it I then delete the branch as well?

Issue #3705, opened 2015-08-21
make the receiving page public by default


whit537:
With #3873 we now have a Receiving page for Teams. However, we made it private to owners (and admins) because it's so rough. This ticket is about phasing it over to public by default. Steps:
- [ ] finish #3992—make the page something to be proud of
- [ ] add an option to make it public; private will be the default option for existing Teams
- [ ] make public the default for new Teams

---

**Original:** There's no way for anyone to see how much Teams are moving.
##

mattbk:
Same as https://github.com/gratipay/gratipay.com/issues/3612?

whit537:
Maybe? They were/are separate for ~users.

mattbk:
~users now have Giving (giving) and History (receiving).
Teams could have Distributing (among team members), History (receiving), and Giving (to other teams).

whit537:
> Giving (to other teams)

We don't have this right now, and I'm not sure we should. Let's discuss in a separate ticket, though.

whit537:
History reflects all past activity in the `payments` and `exchanges` tables (and `transfers`, the old one). Giving and Receiving and Distributing are snapshots from the `payment_instructions` and `payroll` tables.

History is past, Giving, etc. is present/future.

mattbk:
+1 from https://twitter.com/helloanselm/status/636484241560309760

whit537:
> There's no way for anyone to see how much Teams are moving.

This immediate concern was addressed in #3721, though we should keep this ticket open since we still do want separate Receiving and "[Sharing](https://github.com/gratipay/inside.gratipay.com/issues/325)" pages that show more detail.

whit537:
From @anselmh in #3757:

> Prior to the big change it was possible to see how many people participated per payout and also to see how many people gave how much money individually. I miss this option now. Is it gone forever, just coming back or what’s the status here?

rohitpaulk:
I've started work on the receiving pages for teams at https://github.com/gratipay/gratipay.com/pull/3873. To convey the charging in arrears concept effectively, we'll need a proper overhaul of the way we present information - 

I'm thinking of something like this - (taken from [Harvest](http://getharvest.com/))

<img width="574" alt="gratipaydemo" src="https://cloud.githubusercontent.com/assets/3893573/11762963/7917adf2-a11e-11e5-9ea2-f496545cc938.png">

We'll have to tackle https://github.com/gratipay/gratipay.com/issues/3851 first though.

[...]

Issue #3851, opened 2015-10-27
Improve schema for dues history


rohitpaulk:
Reticketed from https://github.com/gratipay/gratipay.com/commit/918cf40a8727680377103970d80ea55c8fbb582b#commitcomment-13966584
##

whit537:
I put this on Finish Teams Basics per https://github.com/gratipay/gratipay.com/pull/3873#issuecomment-165448321.

whit537:
Aaaaaaaand I'm taking it off again, because https://github.com/gratipay/gratipay.com/issues/3992#issuecomment-207852330 and https://github.com/gratipay/gratipay.com/issues/3705#issuecomment-207852318.

Issue #3818, opened 2015-10-11
avoid increasing due beyond payin minimum when cards fail


rohitpaulk:
If a ~user has a $5 tip set up to a team, and their card is failing - we won't charge them again until the card is updated, but we will be adding to their dues every week. If the user updates their card a year later, that's a $260 charge.

This should be clearly communicated to users - https://github.com/gratipay/gratipay.com/pull/3817 is step 1. Adding this to the documentation wherever relevant is step 2.

rohitpaulk:
I just realized that canceling a subscription won't erase the dues - fixed in https://github.com/gratipay/gratipay.com/pull/3819

whit537:
We should never increase dues beyond the minimum charge. If they reach the minimum charge and then their card fails, we should stop adding to due. If/when they fix their card we should charge the minimum at that point. We don't want $260 surprises!

mattbk:
Nice catch.

whit537:
@rohitpaulk Why do we have `Review` on this one?

whit537:
Done with #3876. !m @rohitpaulk :dancer:

PR #3873, opened 2015-12-12
Add team receiving page

rohitpaulk:
https://github.com/gratipay/gratipay.com/issues/3705, https://github.com/gratipay/gratipay.com/issues/3779


rohitpaulk:
Screenshots: 

screen shot 2015-12-12 at 8 28 45 pm

Payment distribution:

screen shot 2015-12-12 at 8 28 55 pm

rohitpaulk:
I'm made the team pages visible to only admins and the team's owner. My reasoning here is that the values we are exposing aren't polished enough to show to the general public. For example, the below graph would make sense to the owner of a team who is aware of charging in arrears - but to others, it'd just look like the number of payments reduced drastically over a week. 

screen shot 2015-12-12 at 9 48 19 pm

Here are a few pages generated from our DB (team names hidden): 

screen shot 2015-12-12 at 9 45 33 pm
screen shot 2015-12-12 at 9 45 04 pm

rohitpaulk:
The estimated next week's payment for Gratipay looks out of line though... 

screen shot 2015-12-12 at 9 45 56 pm

Probably a bug somewhere, I'm going to write a few simple test cases to see if something pops up

rohitpaulk:
The values of unfunded dues are pretty high for most teams, that's due to https://github.com/gratipay/gratipay.com/issues/3818

mattbk:
Nice!

rohitpaulk:
Rebased on master after https://github.com/gratipay/gratipay.com/pull/3787

rohitpaulk:
> The values of unfunded dues are pretty high for most teams, that's due to #3818

Fixed that in #3876, lets look if these values look better after those changes...

rohitpaulk:
Before: 

screen shot 2015-12-12 at 9 45 56 pm

After:

screen shot 2015-12-16 at 9 26 27 pm

rohitpaulk:
The estimated payment is definitely closer, but still pretty high. The last bar in "Dollars received per week" is ~$275.

rohitpaulk:
Everything looks right from my side, I guess we'll find out how good this 'estimated' value is tomorrow :)

rohitpaulk:
I'm going to run a script to generate the estimated values for all teams - And I'll write another one to compare them once we run Payday.

whit537:
@rohitpaulk My main concern after once read-through of this PR is that hiding the receiving amount sets a precedent. If we would rather have team receiving be public then we should make it that way from the outset.

rohitpaulk:
> If we would rather have team receiving be public then we should make it that way from the outset.

There's a reason behind this, mentioned at https://github.com/gratipay/gratipay.com/pull/3873#issuecomment-164165342

> I've made the team pages visible to only admins and the team's owner. My reasoning here is that the values we are exposing aren't polished enough to show to the general public. For example, the below graph would make sense to the owner of a team who is aware of charging in arrears - but to others, it'd just look like the number of payments reduced drastically over a week.

Once I've worked on #3851, I'll work on https://github.com/gratipay/gratipay.com/issues/3705#issuecomment-164167705 and then this will be ready to be shown to the public. 

(said the same over at https://github.com/gratipay/gratipay.com/issues/3705#issuecomment-164168024)

rohitpaulk:
I'm not confident that this is a good enough user experience for the public - but it certainly is for a team owner. And we know from our own experience that huge PRs don't make it till production, hence the attempt to make incremental improvements :)

whit537:
> I'm not confident that this is a good enough user experience for the public - but it certainly is for a team owner.

Right, but it doesn't seem right to roll this out as a private page and then *automatically* make it a public page at some point in the future. If we roll it out as private that'll become the default by default, and we'll have to add a switch to opt in to sharing it publicly once that's ready. I don't think we want that, because openness should be the default for Teams on Gratipay. I'm disinclined to even offer private receiving as an option, actually.

> And we know from our own experience that huge PRs don't make it to production,

Well, that's true. :disappointed: 

> hence the attempt to make incremental improvements :)

In the past, of course, we've simply "released early and often" even if the UX is half-baked (at best). If we want to start raising the bar on what we're willing to release, then one pattern we could try is to make PRs against PRs—PR inception! <:-) That keeps our batch size small for the primary review queue so we can process it efficiently, while giving us a chance to aggregate work units for release based on user-centric concerns. See https://github.com/saxifrage/cityasacampus/pull/364 for an example.

The three options I'm seeing are:

 - Deploy this as-is. Team receiving pages are private by default.
 - Make Team receiving pages public, and then deploy. They're confusing for the general public.
 - Block this on #3851 and https://github.com/gratipay/gratipay.com/issues/3705#issuecomment-164167705, using PRs against this PR as necessary.

How do you think we should proceed, @rohitpaulk?

rohitpaulk:
I'm against option 3, PRs against PRs don't really solve the core problem, which is PRs going stale. Everything works fine as long as we're moving fast - but if someone decides to create the second PR (on top of one) 3 months later, it's still going to be difficult to rebase and update the base PR. What PRs against PRs does solve, is huge PRs that are difficult to review. Since we make our commits oh-so-atomic, I don't think that's a big problem for us :)

Although I think that option 1 is the way to go, both option 1 and 2 are fine with me. Thoughts, @mattbk @webmaven @techtonik @rorepo @kzisme?

whit537:
> I'm against option 3, PRs against PRs don't really solve the core problem, which is PRs going stale. Everything works fine as long as we're moving fast - but if someone decides to create the second PR (on top of one) 3 months later, it's still going to be difficult to rebase and update the base PR.

Agreed.

> both option 1 and 2 are fine with me. 

I'm fine with option 2 if no-one else objects. To soften the blow we could add a note to the chart like we used to have about "minimum charge upped to $10 in week 8" or whatever. There may still be a `note` CSS class hanging around for this occasion.

This PR is good work and we should get it out there! :-)

mattbk:
I see option 1 as a "beta" to allow team owners to give feedback on how accurate it seems ("truthiness," not actual math). If it's deployed privately with a statement that it will be made public in the future (when we decide it's out of beta), it should end up being opt-out, rather than opt-in.

whit537:
> it should end up being opt-out, rather than opt-in.

Why opt-*anything*? Options are confusing.

> I see option 1 as a "beta" to allow team owners to give feedback on how accurate it seems ("truthiness," not actual math).

I could maybe see that if we turned it off entirely for a while to rework it based on feedback before releasing publicly.

whit537:
> If we roll it out as private that'll become the default by default, and we'll have to add a switch to opt in to sharing it publicly once that's ready.

@mattbk @rohitpaulk I haven't seen you directly address this. Do you admit this consequence of option 1 yet accept it? Or do you not admit this consequence? Or ... ?

rohitpaulk:
> I haven't seen you directly address this

I think this depends on how we present it... As long as we make it clear (by adding a note, as @mattbk suggested) that we intend to make it public in the near future, we won't be deceiving anyone. If we face significant pushback on the same, it's probably a sign that we should reconsider providing anonymous receiving options.

> I see option 1 as a "beta" to allow team owners to give feedback on how accurate it seems ("truthiness," not actual math). If it's deployed privately with a statement that it will be made public in the future (when we decide it's out of beta), it should end up being opt-out, rather than opt-in.

> Why opt-anything? Options are confusing.

Shouldn't we consider deciding this based on the feedback that we receive from team owners? Rather than forcing this upon them - why not open this to them and see what they have to say?

rohitpaulk:
This is blocked on https://github.com/gratipay/gratipay.com/pull/3876

rohitpaulk:
Rebased on master at @ec2c732d822cb446dd8c43540c5a7c3a9ea8522a. Previous head was @673475357000a7f0bf8e5fd4a6fff58843734a0f

rohitpaulk:
#3980 and #3981 are out, this is no longer blocked.

rohitpaulk:
Rebased on master.

rohitpaulk:
Did a bit of browsing through on a recent backup, the charts look okay to me.

whit537:
Alright, this is super-rough but [we know that](https://github.com/gratipay/gratipay.com/pull/3873#issuecomment-164165342). I'm merging in order to [keep our velocity up](https://github.com/gratipay/gratipay.com/pull/3873#issuecomment-165449001). I've reticketed #3992 for further iteration, still on Finish Team Basics.

!m @rohitpaulk

whit537:
(P.S. Also this is read-only so it's low-risk.)

PR #3785, opened 2015-09-14
stop Python-interpolating payday.sql

whit537:
Using Python string interpolation on the `payday.sql` file introduces complexity that leads to bugs. The one I hit was using `%` for interpolation in a Postgres `RAISES` clause: it would've had to be escaped to make it through the Python interpolation.

I cherry-picked this from 364d066bb4e7f69b7a2febc31230fdb8e57b05a6 in #3653.

See reference in https://github.com/gratipay/gratipay.com/issues/3780#issuecomment-139909299.

Issue #3506, opened 2015-06-01
require email before charging users


whit537:
Primary driver is consumer protection: we want to notify people about Gratipay charges to reduce surprises.
##

whit537:
> Additionally, our sponsoring bank suggested collecting and saving cardholder information, such as phone numbers, in the future for cases where documentation may be necessary. These situations may occur if you have to dispute a chargeback and provide evidence for your supporting case. 

https://github.com/gratipay/gratipay.com/issues/3531#issuecomment-112256884

whit537:
+1 https://gratipay.freshdesk.com/helpdesk/tickets/3292

whit537:
+1 https://github.com/gratipay/inside.gratipay.com/issues/477#issuecomment-173634468

mattbk:
Requiring email would help in cases like https://github.com/gratipay/gratipay.com/pull/4221, where the ToS is changing and we should notify people.

PR #3876, opened 2015-12-12
Cap unfunded dues

rohitpaulk:
https://github.com/gratipay/gratipay.com/issues/3818


rohitpaulk:
Now to change the existing values...

rohitpaulk:
Not ready to merge. I'm seeing weird data values in our DB, got to investigate more... 

```sql
=> SELECT amount, due, is_funded FROM current_payment_instructions WHERE due > 10 AND amount > 0 AND is_funded;
 amount |  due   | is_funded
--------+--------+-----------
   1.00 |  14.00 | t
  10.00 | 140.00 | t
   5.00 |  70.00 | t
   1.00 |  11.00 | t
   1.00 |  14.00 | t
   1.00 |  14.00 | t
   1.00 |  14.00 | t
   1.00 |  14.00 | t
   5.00 |  70.00 | t
   1.00 |  14.00 | t
   1.00 |  14.00 | t
   1.00 |  14.00 | t
   1.00 |  14.00 | t
   2.25 |  31.50 | t
   8.00 |  80.00 | t
   2.00 |  28.00 | t
   1.00 |  14.00 | t
   1.00 |  14.00 | t
   5.00 |  70.00 | t
   1.00 |  14.00 | t
   1.00 |  11.00 | t
   1.00 |  12.00 | t
   0.75 |  10.50 | t
   1.00 |  14.00 | t
(24 rows)
```

How do funded payments get to a value greater than the minimum charge???

rohitpaulk:
^ cc: @rorepo

rohitpaulk:
Looks like none of these ~users have CCs attached... Do we not make `is_funded` as false when a CC is removed?

rohitpaulk:
Aha. These are people who have *never* attached a CC. The tips were funded because of the balance. We used to refresh these values after every Payday. 

I guess running `Participant.update_giving` on all users should correct this problem. Going to try that on a backup...

gratipay/billing/payday.py
```diff
@@ -27,7 +27,7 @@
 
 
 with open('sql/payday.sql') as f:
-    PAYDAY = f.read()
+    PAYDAY = f.read() % {'minimum_charge': MINIMUM_CHARGE}
```
whit537: Ick. We moved away from extra interpolation in ... b045fb1f0edb4af19e289239ff691b8131d0f918 (#3785). Can we avoid reintroducing that here? I guess that would mean putting `minimum_charge` in the database somewhere.

sql/branch.sql
```diff
@@ -0,0 +1,4 @@
+BEGIN;
+    UPDATE payment_instructions SET due = 0 WHERE amount = 0 AND due != 0;
+    UPDATE payment_instructions SET due = floor(9.81/(amount)) * amount WHERE due > 9.81;
```
whit537: Help me out with the 9.81 ... I don't recognize it as the minimum amount minus credit card fees, or something like that. Where does it come from?

sql/branch.py
```diff
@@ -0,0 +1,16 @@
+from gratipay import wireup
+
+db = wireup.db(wireup.env())
+
+participants = db.all("""
+    SELECT p.*::participants
+      FROM participants p
+      JOIN payment_instructions pi ON pi.participant = p.username -- Only include those with tips
+""")
+total = len(participants)
```
whit537:
```sql
=> select count(*) from participants p JOIN payment_instructions pi ON pi.participant = p.username;
┌───────┐
│ count │
├───────┤
│  2921 │
└───────┘
(1 row)
```

I would expect this script to run for a few minutes, anyway.

gratipay/billing/payday.py
rohitpaulk: Aha, I'll see if I can get this done in a cleaner fashion.

sql/branch.sql
rohitpaulk: Isn't that the minimum charge?

rohitpaulk: /me checks...

rohitpaulk: https://github.com/gratipay/gratipay.com/blob/master/gratipay/billing/exchanges.py#L20

Oh, 9.41. I'll make the change soon

sql/branch.py
rohitpaulk: Yes, and I think that's fine - the operations are idempotent so crash/rerun shouldn't be a problem.

sql/branch.sql
whit537: P.S. If you replace `master` in that URL with the [current version number](https://gratipay.com/version.txt) then you get a permalink—master is likely to drift over time. You can also hit "y" as a GitHub keyboard shortcut but our version number is neater than the git hash. ;)

whit537:
Initial punchlist:

 - [ ] avoid Python-interpolating SQL—https://github.com/gratipay/gratipay.com/pull/3876#discussion_r48239765
 - [ ] update the minimum charge amount—https://github.com/gratipay/gratipay.com/pull/3876#discussion_r48240586

I still need to slow down and read this code to understand what we're doing here ...

sql/branch.sql
whit537: Done in 3ecb129b0ed811dc19377c9836cc448dd669b928.

gratipay/billing/payday.py
whit537: How about the approach in 249379279a931a3ece8d29e2e9ec255665c997ad, @rohitpaulk?

whit537:
I'm not sure what's up with this Travis failure. I'm pretty sure it's present on master, too ...

whit537:
Hoping to read and grok this PR tomorrow.

whit537:
Build *is* failing on master: https://travis-ci.org/gratipay/gratipay.com/builds/98361516. Reticketed as #3885.

whit537:
I've kicked Travis post-#3891.

whit537:
Looks like there is still some breakage ...

gratipay/billing/payday.py
rohitpaulk: Looks good to me :)

whit537: Cool. A further improvement could be to call it `payday_settings`, and create it as a temporary table as part of payday.

rohitpaulk: I think having it as a permanent table makes sense - If we start to write heavy SQL outside of payday, it could be useful.

rohitpaulk:
@whit537 - What's left here?

whit537:
Not much ... looks like I pushed some commits that didn't end up on this branch?

rohitpaulk:
Ohhhhhh

rohitpaulk:
I might've forced pushed :/ Do you have the branch on your local? If so, please push it to a different branch and I'll merge the two manually

whit537:
No worries. My commits are still in GitHub: https://github.com/gratipay/gratipay.com/commits/249379279a931a3ece8d29e2e9ec255665c997ad.

whit537:
;-)

```
$ gf
remote: Counting objects: 57, done.
remote: Total 57 (delta 36), reused 36 (delta 36), pack-reused 21
Unpacking objects: 100% (57/57), done.
From github.com:gratipay/gratipay.com
 + 2493792...12057d5 cap-unfunded-dues -> origin/cap-unfunded-dues  (forced update)
```

whit537:
Alright, brought those back. We'll see if Travis is okay with them.

rohitpaulk:
Ah, this is happening because the DB is wiped out after every test, and the `settings` table is wiped out too.

whit537:
> and the `settings` table is wiped out too.

I thought I tried to address that ...

whit537:
> I thought I tried to address that ...

https://github.com/gratipay/gratipay.com/commit/5f75997fca0f559f6644866ce96fb7c65d4bbfd1#diff-b61f891149698bae209a4723f9165be5R118

gratipay/testing/__init__.py
```diff
@@ -116,6 +116,10 @@ def clear_tables(self):
                 tablenames.insert(0, tablename)
         self.db.run("ALTER SEQUENCE participants_id_seq RESTART WITH 1")
 
+        # TODO: Find a better way to do this
```
whit537: Factor out a `gratipay.wireup.settings` and call that here?

rohitpaulk:
@whit537 - How about https://github.com/gratipay/gratipay.com/commit/5efe6f6bde2de4e8e202087f57227c53278219b8?

whit537:
wfm

whit537:
And works for Travis ...

whit537:
I still need to fundamentally understand what this PR is doing, one sec ...

whit537:
Right, so, we want to avoid increasing dues beyond the minimum charge in the case of a credit card failure.

gratipay/wireup.py
```diff
@@ -59,6 +59,11 @@ def db(env):
 
     return db
 
+def settings(db):
+    from gratipay.billing.exchanges import MINIMUM_CHARGE
+    db.run('DELETE FROM settings')
```
whit537: Seems a little funky to have multiple web app processes potentially running this code at the same time. At the very least we should put this in a transaction so we don't try to delete from settings when there *is* no settings.

sql/branch.sql
```diff
@@ -0,0 +1,6 @@
+BEGIN;
+    UPDATE payment_instructions SET due = 0 WHERE amount = 0 AND due != 0;
+    UPDATE payment_instructions SET due = floor(9.41/(amount)) * amount WHERE due > 9.41;
+    CREATE TABLE settings (minimum_charge numeric(35,2) DEFAULT NULL);
+    INSERT INTO settings DEFAULT VALUES;
```
whit537: I guess this `INSERT` is redundant now that `gratipay.wireup.settings` deletes and inserts.

gratipay/wireup.py
rohitpaulk: > Seems a little funky to have multiple web app processes potentially running this code at the same time.

Nice catch - we could end up with two inserts, and all our `db.one` calls will fail.

sql/branch.sql
whit537: Done in 6722e5e7d8861588b312b707a3371434c372025d.

gratipay/wireup.py
whit537: Done in 6722e5e7d8861588b312b707a3371434c372025d.

sql/payday.sql
```diff
@@ -186,13 +186,14 @@ CREATE OR REPLACE FUNCTION process_payment_instruction() RETURNS trigger AS $$
               FROM payday_participants p
              WHERE username = NEW.participant
         );
+
         IF (NEW.amount + NEW.due <= participant.new_balance OR participant.card_hold_ok) THEN
             EXECUTE pay(NEW.participant, NEW.team, NEW.amount + NEW.due, 'to-team');
             RETURN NEW;
-        ELSE
+        ELSIF NEW.amount + NEW.due <= (SELECT minimum_charge FROM settings) THEN
```
whit537: Is 9.41 the value we want here? Why not 10.00?

rohitpaulk:
https://github.com/gratipay/gratipay.com/commit/6722e5e7d8861588b312b707a3371434c372025d lgtm

sql/payday.sql
whit537: It looks like the way we [use](https://github.com/gratipay/gratipay.com/blob/6722e5e7d8861588b312b707a3371434c372025d/gratipay/billing/exchanges.py#L169) `MINIMUM_CHARGE`, we should actually make this `<` instead of `<=`. Let's get a test or two for the edge cases here ...

rohitpaulk: This is at a point where we haven't introduced fees into the equation, and our minimum charge is 9.41 (actual money in gratipay) + 0.59 (CC processor fees)

whit537: Alright, so I realize this is kinda late in the game, aaaaaaand this was [my idea](https://github.com/gratipay/gratipay.com/issues/3818#issuecomment-150629335), but isn't it kinda weird to use `MINIMUM_CHARGE` as a *maximum* now?

rohitpaulk: Yep, I was thinking about that too - and I now see one approach that'll give us the same end result, but might make more sense. How about not increase dues for any participant without [`has_credit_card`](https://github.com/gratipay/gratipay.com/blob/master/sql/payday.sql#L17) set to `true`?

whit537: > We should never increase dues beyond the minimum charge. If they reach the minimum charge and then their card fails, we should stop adding to due. If/when they fix their card we should charge the minimum at that point. We don't want $260 surprises!

I mean, I guess it makes sense. The reason we are parking dues is to build up to the minimum charge, and once that threshold is reached, we charge the user. In the ideal failure case (modulo #3506), they get an email notification about the failure, fix their card, and get back on track the next week.

rohitpaulk: What happens if a user has a tip set up, but doesn't have a credit card to back it? Should we accumulate dues?

rohitpaulk: > We don't want $260 surprises!
> but isn't it kinda weird to use MINIMUM_CHARGE as a maximum now?
This PR solves the $260 surprise problem, but as you said - the semantics don't seem right..

whit537: > How about not increasing dues for any participant without `has_credit_card` set to true?

That could work. Participants with a balance but no credit card don't accumulate dues anyway, right? As long as they had sufficient balance their payment would still go through.

whit537: > What happens if a user has a tip set up, but doesn't have a credit card to back it? Should we accumulate dues?

I think not.

whit537:
Interesting results looking at the minimum charge rounding in recent paydays. It's a single user the past couple months.

screen shot 2016-01-27 at 2 32 13 pm

sql/payday.sql
whit537: > How about not increasing dues for any participant without `has_credit_card` set to `true`?

That would mean we don't need `minimum_charge` in the database, so we can back out the `settings` table.

whit537:
> How about not increasing dues for any participant without `has_credit_card` set to `true`?

9ccb02a23fd20e5478b8abfb17021d52f29f0d9a, ya?

sql/branch.sql
```diff
@@ -0,0 +1,4 @@
+BEGIN;
+    UPDATE payment_instructions SET due = 0 WHERE amount = 0 AND due != 0;
+    UPDATE payment_instructions SET due = floor(9.41/(amount)) * amount WHERE due > 9.41;
```
whit537: Alright, so what's up with these updates. Why are these the right repairs for our situation?

whit537: I count 579 current payment instructions with due >= 9.41. I guess these are the folks that never had a card attached?

whit537: Seems like we could reset those dues all to zero.

whit537: And I think we can prune the second update now that we're not going from `MINIMUM_CHARGE`.

whit537: Does this cover it?

```sql
UPDATE payment_instructions SET due=0 WHERE due > 9.41;
```

whit537: Done in 2454f8f6b2d22056d747e78659ef55ea2120f9bc.

sql/branch.sql
```diff
@@ -0,0 +1,3 @@
+BEGIN;
+    UPDATE payment_instructions SET due = 0 WHERE due > 9.41;
```
whit537: We could do something like this instead:

```sql
BEGIN;
    CREATE TEMP TABLE _participants AS
        SELECT id
             , username
             , ( SELECT count(*)
                   FROM current_exchange_routes r
                  WHERE r.participant = p.id
                    AND network = 'braintree-cc'
                    AND error = ''
               ) > 0 AS has_credit_card
          FROM participants p;

    UPDATE payment_instructions SET due = 0 WHERE due >= 9.41 AND participant IN (
        SELECT username FROM _participants WHERE NOT has_credit_card
    );

    DROP TABLE _participants;
END;
```

whit537: I'm outta time to work on this today, though. I will try to pick up again tomorrow, @rohitpaulk.

whit537:
@rohitpaulk I think we're almost ready to land this, we just need to make sure we're repairing old dues correctly: 187fbd27e155f5b401cabd7dd3f89ec5178b4bfe and https://github.com/gratipay/gratipay.com/pull/3876#discussion_r51040424 and https://github.com/gratipay/gratipay.com/pull/3876#discussion_r51043776.

whit537:
Rebased on master at af05120. Previous head was 718a1937327741ecda5b9f732fb2264c30d6fbc7.

rohitpaulk:
Rebased on master again.

whit537:
Rebased on master at ec2c732d822cb446dd8c43540c5a7c3a9ea8522a. Previous head was ad82f6073adfa6832400e1c2560a162b7b60c521.

rohitpaulk:
The commit log on this is pretty confusing - makes it hard to review. I'm going to squash this down to 2-3 meaningful commits.

rohitpaulk:
I'm seeing three major changes here, the rest is just noise.

1. Prevent dues from exceeding minimum_charge in payday
2. Add SQL script to repair old values that are above minimum charge
3. Add script to recalculate stale cached values

rohitpaulk:
I'm going to cherry pick **1** into a separate PR - and retain **2** and **3** as part of this PR.

rohitpaulk:
Split into https://github.com/gratipay/gratipay.com/pull/3981 and https://github.com/gratipay/gratipay.com/pull/3980

sql/branch.sql
whit537: Actually ... 187fbd27e155f5b401cabd7dd3f89ec5178b4bfe, ya?

PR #3896, opened 2016-01-19
Change emails foreign key - 2

aandis:
This is the next step after #3893. Pr includes a migration to fill the new `emails.participant_id` column with existing data and modifies email reads to then read from the new column instead of `emails.participant`.

### DIFF
--- a/gratipay/models/participant.py
+++ b/gratipay/models/participant.py
@@ -347,7 +347,7 @@ class Participant(Model):
                    AND is_member IS true
             );
 
-            DELETE FROM emails WHERE participant = %(username)s;
+            DELETE FROM emails WHERE participant_id = %(participant_id)s;
             DELETE FROM statements WHERE participant=%(participant_id)s;
 
             UPDATE participants
@@ -380,10 +380,11 @@ class Participant(Model):
 
         # Check that this address isn't already verified
         owner = self.db.one("""
-            SELECT participant
-              FROM emails
-             WHERE address = %(email)s
-               AND verified IS true
+            SELECT p.username
+              FROM emails e INNER JOIN participants p
+                ON e.participant_id = p.id
+             WHERE e.address = %(email)s
+               AND e.verified IS true
         """, locals())
         if owner:
             if owner == self.username:
@@ -416,11 +417,11 @@ class Participant(Model):
             nonce = self.db.one("""
                 UPDATE emails
                    SET verification_start=%s
-                 WHERE participant=%s
+                 WHERE participant_id=%s
                    AND address=%s
                    AND verified IS NULL
              RETURNING nonce
-            """, (verification_start, self.username, email))
+            """, (verification_start, self.id, email))
             if not nonce:
                 return self.add_email(email)
 
@@ -470,10 +471,10 @@ class Participant(Model):
             self.db.run("""
                 UPDATE emails
                    SET verified=true, verification_end=now(), nonce=NULL
-                 WHERE participant=%s
+                 WHERE participant_id=%s
                    AND address=%s
                    AND verified IS NULL
-            """, (self.username, email))
+            """, (self.id, email))
         except IntegrityError:
             return emails.VERIFICATION_STYMIED
 
@@ -485,17 +486,17 @@ class Participant(Model):
         return self.db.one("""
             SELECT *
               FROM emails
-             WHERE participant=%s
+             WHERE participant_id=%s
                AND address=%s
-        """, (self.username, email))
+        """, (self.id, email))
 
     def get_emails(self):
         return self.db.all("""
             SELECT *
               FROM emails
-             WHERE participant=%s
+             WHERE participant_id=%s
           ORDER BY id
-        """, (self.username,))
+        """, (self.id,))
 
     def get_verified_email_addresses(self):
         return [email.address for email in self.get_emails() if email.verified]
@@ -505,8 +506,8 @@ class Participant(Model):
             raise CannotRemovePrimaryEmail()
         with self.db.get_cursor() as c:
             add_event(c, 'participant', dict(id=self.id, action='remove', values=dict(email=address)))
-            c.run("DELETE FROM emails WHERE participant=%s AND address=%s",
-                  (self.username, address))
+            c.run("DELETE FROM emails WHERE participant_id=%s AND address=%s",
+                  (self.id, address))
 
     def send_email(self, spt_name, **context):
         context['participant'] = self
@@ -1383,16 +1384,16 @@ class Participant(Model):
             WITH emails_to_keep AS (
                      SELECT DISTINCT ON (address) id
                        FROM emails
-                      WHERE participant IN (%(dead)s, %(live)s)
+                      WHERE participant_id IN (%(dead)s, %(live)s)
                    ORDER BY address, verification_end, verification_start DESC
                  )
             DELETE FROM emails
-             WHERE participant IN (%(dead)s, %(live)s)
+             WHERE participant_id IN (%(dead)s, %(live)s)
                AND id NOT IN (SELECT id FROM emails_to_keep);
 
             UPDATE emails
-               SET participant = %(live)s
-             WHERE participant = %(dead)s;
+               SET participant_id = %(live)s
+             WHERE participant_id = %(dead)s;
 
         """
 
@@ -1525,7 +1526,7 @@ class Participant(Model):
                 # Take over email addresses.
                 # ==========================
 
-                cursor.run(MERGE_EMAIL_ADDRESSES, dict(live=x, dead=y))
+                cursor.run(MERGE_EMAIL_ADDRESSES, dict(live=self.id, dead=other.id))
 
                 # Disconnect any remaining elsewhere account.
                 # ===========================================
--- a/sql/branch.sql
+++ b/sql/branch.sql
@@ -1,0 +1,5 @@
+UPDATE emails AS e
+SET participant_id = p.id
+FROM participants AS p
+WHERE p.username = e.participant
+AND e.participant_id IS NULL;
--- a/sql/branch.sql
+++ b/sql/branch.sql
@@ -3,3 +3,5 @@ SET participant_id = p.id
 FROM participants AS p
 WHERE p.username = e.participant
 AND e.participant_id IS NULL;
+
+ALTER TABLE emails ALTER COLUMN participant_id SET NOT NULL;
--- a/tests/py/test_email.py
+++ b/tests/py/test_email.py
@@ -102,8 +102,8 @@ class TestEmail(EmailHarness):
         self.db.run("""
             UPDATE emails
                SET verification_start = (now() - INTERVAL '25 hours')
-             WHERE participant = 'alice'
-        """)
+             WHERE participant_id = %s;
+        """, (self.alice.id,))
         nonce = self.alice.get_email(address).nonce
         r = self.alice.verify_email(address, nonce)
         assert r == emails.VERIFICATION_EXPIRED


aandis:
I tried to modify all reads which use `emails.participant` to use `emails.participant_id`. But a review is needed to make sure I didn't miss anything.

gratipay/models/participant.py
```diff
@@ -380,10 +380,11 @@ def add_email(self, email, resend_threshold='3 minutes'):
...
         # Check that this address isn't already verified
         owner = self.db.one("""
-            SELECT participant
-              FROM emails
-             WHERE address = %(email)s
-               AND verified IS true
+            SELECT p.username
+              FROM emails e INNER JOIN participants p
+                ON e.participant_id = p.id
+              WHERE e.address = %(email)s
```
rohitpaulk: change_emails_foreign_key_-_2_by_aandis_ _pull_request__3896_ _gratipay_gratipay_com

:angry:

rohitpaulk:
> But a review is needed to make sure I didn't miss anything.

I hope our tests are exhaustive enough to catch this...

Try removing the address column in your local, and see if the tests pass? If they do, then I think we're good to go... 

 If we do encounter errors in production, we'll patch them and add tests so that they never happen again.

gratipay/models/participant.py
aandis: ok so that's what the uneven indentation at the start of the line is about! I was scratching my head with that. Man, now you're just messing with a java guy. :smirk:

aandis: Regardless, I'll admit that this does look cleaner. Fixing.

aandis:
Running test.

gratipay/models/participant.py
rohitpaulk: https://botbot.me/freenode/gratipay/2015-03-19/?msg=34537414&page=1 ;)

gratipay/models/participant.py
```diff
@@ -416,11 +417,11 @@ def add_email(self, email, resend_threshold='3 minutes'):
...
                    SET verification_start=%s
-                 WHERE participant=%s
+                 WHERE participant_id=%s
                    AND address=%s
                    AND verified IS NULL
              RETURNING nonce
-            """, (verification_start, self.username, email))
+            """, (verification_start, self.id, email))
             if not nonce:
                 return self.add_email(email)
```
aandis: @rohitpaulk tests are failing. :disappointed: All the email tests get suck in this recursive call. Do you have any idea why? I have dropped the `participant` column and modified email writes to only include `participant_id`

gratipay/models/participant.py
aandis: Nice! Should answer [this](http://stackoverflow.com/questions/5243596/python-sql-query-string-formatting) :smile:

gratipay/models/participant.py
rohitpaulk: Can you share a diff of the changes you made vs the code in this PR?

aandis:
``` diff
diff --git a/gratipay/models/participant.py b/gratipay/models/participant.py
index 242ebff..e64c508 100644
--- a/gratipay/models/participant.py
+++ b/gratipay/models/participant.py
@@ -410,9 +410,9 @@ class Participant(Model):
                 add_event(c, 'participant', dict(id=self.id, action='add', values=dict(email=email)))
                 c.run("""
                     INSERT INTO emails
-                                (address, nonce, verification_start, participant, participant_id)
-                         VALUES (%s, %s, %s, %s, %s)
-                """, (email, nonce, verification_start, self.username, self.id))
+                                (address, nonce, verification_start, participant_id)
+                         VALUES (%s, %s, %s, %s)
+                """, (email, nonce, verification_start, self.id))
         except IntegrityError:
             nonce = self.db.one("""
                 UPDATE emails
```

rohitpaulk:
```diff
diff --git a/gratipay/models/participant.py b/gratipay/models/participant.py
index 242ebff..e64c508 100644
--- a/gratipay/models/participant.py
+++ b/gratipay/models/participant.py
@@ -410,9 +410,9 @@ class Participant(Model):
                 add_event(c, 'participant', dict(id=self.id, action='add', values=dict(email=email)))
                 c.run("""
                     INSERT INTO emails
-                                (address, nonce, verification_start, participant, participant_id)
-                         VALUES (%s, %s, %s, %s, %s)
-                """, (email, nonce, verification_start, self.username, self.id))
+                                (address, nonce, verification_start, participant_id)
+                         VALUES (%s, %s, %s, %s)
+                """, (email, nonce, verification_start, self.id))
         except IntegrityError:
             nonce = self.db.one("""
                 UPDATE emails
diff --git a/sql/branch.sql b/sql/branch.sql
index 4a74cea..f316636 100644
--- a/sql/branch.sql
+++ b/sql/branch.sql
@@ -2,4 +2,6 @@ UPDATE emails AS e
 SET participant_id = p.id
 FROM participants AS p
 WHERE p.username = e.participant
-AND e.participant_id IS NULL;
\ No newline at end of file
+AND e.participant_id IS NULL;
+
+ALTER TABLE emails DROP COLUMN participant;
```

That's weird... I applied the above diff and all tests are passing for me..

rohitpaulk: What's your test output? Which test is getting stuck in the loop?

rohitpaulk: I hope that I've managed to convince you of the importance of tests (though it took a long time) :P

aandis: Haha. This is convincing enough. Most email test are failing. Below is one,

```
====================================================================== FAILURES ======================================================================
__________________________________________________ TestClosing.test_cpi_clears_personal_information __________________________________________________

args = (<test_close.TestClosing testMethod=test_cpi_clears_personal_information>, <MagicMock name=u'_mailer' id='46937897223696'>), keywargs = {}
extra_args = [<MagicMock name=u'_mailer' id='46937897223696'>], entered_patchers = [<mock._patch object at 0x2ab0944ed990>]
exc_info = (<type 'exceptions.RuntimeError'>, RuntimeError('maximum recursion depth exceeded while encoding a JSON object',), <traceback object at 0x2ab0949db950>)
patching = <mock._patch object at 0x2ab0944ed990>, arg = <MagicMock name=u'_mailer' id='46937897223696'>

    @wraps(func)
    def patched(*args, **keywargs):
        # don't use a with here (backwards compatability with Python 2.4)
        extra_args = []
        entered_patchers = []
    
        # can't use try...except...finally because of Python 2.4
        # compatibility
        exc_info = tuple()
        try:
            try:
                for patching in patched.patchings:
                    arg = patching.__enter__()
                    entered_patchers.append(patching)
                    if patching.attribute_name is not None:
                        keywargs.update(arg)
                    elif patching.new is DEFAULT:
                        extra_args.append(arg)
    
                args += tuple(extra_args)
>               return func(*args, **keywargs)

env/lib/python2.7/site-packages/mock.py:1201: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
tests/py/test_close.py:147: in test_cpi_clears_personal_information
    alice.add_email('alice@example.net')
gratipay/models/participant.py:426: in add_email
    return self.add_email(email)
gratipay/models/participant.py:426: in add_email
    return self.add_email(email)
gratipay/models/participant.py:426: in add_email
    return self.add_email(email)
gratipay/models/participant.py:426: in add_email
    return self.add_email(email)
gratipay/models/participant.py:426: in add_email
    return self.add_email(email)
gratipay/models/participant.py:426: in add_email
    return self.add_email(email)
gratipay/models/participant.py:426: in add_email
    return self.add_email(email)

```

aandis: Basically, anything which calls `add_email` gets stuck.

rohitpaulk: Hmm. The Integrity error is being hit, which means a record exists. If a record exists though, we shouldn't be hitting the recursive loop... 

Given that the loop exited right there, we should be able to inspect the DB and see what's happening.. Open psql and try? Let's make sure we have the same setup though... here's mine

```bash
$ psql gratipay-test

gratipay-test=# \d emails
                                        Table "public.emails"
       Column       |           Type           |                      Modifiers
--------------------+--------------------------+-----------------------------------------------------
 id                 | integer                  | not null default nextval('emails_id_seq'::regclass)
 address            | text                     | not null
 verified           | boolean                  |
 nonce              | text                     |
 verification_start | timestamp with time zone | not null default now()
 verification_end   | timestamp with time zone |
 participant_id     | bigint                   |
Indexes:
    "emails_pkey" PRIMARY KEY, btree (id)
    "emails_address_verified_key" UNIQUE CONSTRAINT, btree (address, verified)
    "emails_participant_id_address_key" UNIQUE CONSTRAINT, btree (participant_id, address)
Check constraints:
    "verified_cant_be_false" CHECK (verified IS NOT FALSE)
Foreign-key constraints:
    "emails_participant_id_fkey" FOREIGN KEY (participant_id) REFERENCES participants(id) ON UPDATE RESTRICT ON DELETE RESTRICT

gratipay-test=#
```

aandis: ok I have no idea what the problem was but I couldn't even run the site locally at `localhost:8537`. `make run` kept exiting. So I restarted my PC and now everything works! All tests pass. :dancer:

rohitpaulk: Aha. So this is ready to go, eh?

aandis: Aha! I see. I applied your diff before running test. `participant` was still existing in `gratipay-test`. I had no idea `branch.sql` is run by the test suite. Figured it'd be like rails where the test database is created from the current schema.

aandis: > Aha. So this is ready to go, eh?

Yes! :dancer:

aandis:
@rohitpaulk on a side not, why is that recursive call there anyway? If record failed to be created once because of a reason like this, it'll surely keep failing.

rohitpaulk:
Okay we need one last thing - we want to constrain this `participant` column to `NOT NULL`. I guess that we should be able to do that in this PR itself, right?

aandis:
> we want to constrain this participant column to NOT NULL

`participant_id` you mean. Okay on it.

rohitpaulk:
> on a side not, why is that recursive call there anyway?

The recursive call handles a specific race condition, I think.

What if an email address existed - and was unverified before the `try\except` block, but by the time we hit the `except`, it was verified?

participant_py_ _gratipay_com

I do get that the code can be kinda extremely hard to read though...

```python
"""
           This is called when
            1) Adding a new email address
            2) Resending the verification email for an unverified email address

            Returns the number of emails sent.
"""
```

This might benefit from being split into two separate functions (and extracting common stuff out into different functions), seems to me like we're trying to stuff too much logic in here.

rohitpaulk:
Okay, this looks good to me. I'll try deploying this tomorrow. If anyone gets to this before me, keep in mind that we should run the migration script **before** deploying code.

aandis:
:+1:

aandis:
@rohitpaulk Did you get into any issues while running the migration for copying values in production?

rohitpaulk:
> Did you get into any issues while running the migration for copying values in production?

Nope, smooth sailing :boat:

techtonik:
@rohitpaulk awesome graphical analysis. I guess that the bugs like this is the primary use case that is solved by Flux architecture - https://facebook.github.io/flux/docs/overview.html

PR #3904, opened 2016-01-23
Speed up tip migration. Fixes #3894

### LINKED ISSUE #3894: Speed up tip migration script
Reticketed from https://github.com/gratipay/inside.gratipay.com/issues/468#issuecomment-171707621

When we wrote this script, we thought that we'd only be using it for a month or two. Now that it's here to stay - time to work on it a bit :) 

The script takes so long because we're pulling in *all* approved teams. We could place a check in the script so that we only pull in those teams that satisfy the following criteria - 

1) Team owner must have 0+ tips from Gratipay 1.0
2) Team must have zero `payment_instructions` that have a `ctime` lesser than the team's `ctime` (i.e. migrated tips). We use the `ctime` attribute to differentiate `payment_instructions` that were created as a result of migrated tips vs created in Gratipay 2.0

https://github.com/gratipay/gratipay.com/blob/master/bin/migrate-tips.py#L6-L10

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/29958036-speed-up-tip-migration-script?utm_campaign=plugin&utm_content=tracker%2F85909&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F85909&utm_medium=issues&utm_source=github).
</bountysource-plugin>

aandis:
@rohitpaulk , @whit537 tests are failing as of now. If this looks good, I'll go ahead and fix them. ;)

https://github.com/gratipay/gratipay.com/issues/3894

### DIFF
--- a/bin/migrate-tips.py
+++ b/bin/migrate-tips.py
@@ -1,20 +1,26 @@
 from gratipay.wireup import db, env
-from gratipay.models.team import Team, AlreadyMigrated
+from gratipay.models.team import AlreadyMigrated
 
 db = db(env())
 
-slugs = db.all("""
-    SELECT slug
-      FROM teams
-     WHERE is_approved IS TRUE
+teams = db.all("""
+    SELECT distinct ON (t.slug) t.*::teams
+      FROM teams t
+      JOIN tips ON t.owner = tips.tippee    -- Only fetch teams whose owner have tips.
+     WHERE t.is_approved IS TRUE            -- Only fetch approved teams.
+       AND NOT EXISTS (                     -- Make sure not already migrated.
+            SELECT 1
+              FROM payment_instructions pi
+             WHERE t.slug = pi.team
+               AND pi.ctime < t.ctime
+	     )
 """)
 
-for slug in slugs:
-    team = Team.from_slug(slug)
+for team in teams:
     try:
         ntips = team.migrate_tips()
-        print("Migrated {} tip(s) for '{}'".format(ntips, slug))
+        print("Migrated {} tip(s) for '{}'".format(ntips, team.slug))
     except AlreadyMigrated:
-        print("'%s' already migrated." % slug)
+        print("'%s' already migrated." % team.slug)
 
 print("Done.")
--- a/bin/migrate-tips.py
+++ b/bin/migrate-tips.py
@@ -6,14 +6,15 @@ db = db(env())
 teams = db.all("""
     SELECT distinct ON (t.slug) t.*::teams
       FROM teams t
-      JOIN tips ON t.owner = tips.tippee    -- Only fetch teams whose owner have tips.
-     WHERE t.is_approved IS TRUE            -- Only fetch approved teams.
-       AND NOT EXISTS (                     -- Make sure not already migrated.
+      JOIN tips ON t.owner = tips.tippee    -- Only fetch teams whose owners had tips under Gratipay 1.0
+     WHERE t.is_approved IS TRUE            -- Only fetch approved teams
+       AND NOT EXISTS (                     -- Make sure tips haven't been migrated for any teams with same owner
             SELECT 1
               FROM payment_instructions pi
-             WHERE t.slug = pi.team
-               AND pi.ctime < t.ctime
-	     )
+              JOIN teams t2 ON t2.slug = pi.team
+             WHERE t2.owner = t.owner
+               AND pi.ctime < t2.ctime
+       )
 """)
 
 for team in teams:
--- a/gratipay/models/team.py
+++ b/gratipay/models/team.py
@@ -162,8 +162,7 @@ class Team(Model):
             SELECT pi.*
               FROM payment_instructions pi
               JOIN teams t ON t.slug = pi.team
-              JOIN participants p ON t.owner = p.username
-             WHERE p.username = %s
+             WHERE t.owner = %s
                AND pi.ctime < t.ctime
         """, (self.owner, ))
 


gratipay/models/team.py
```diff
@@ -158,19 +158,6 @@ def to_dict(self):
         }
 
     def migrate_tips(self):
-        payment_instructions = self.db.all("""
```
rohitpaulk: I think the basic filtering done above should be good enough - removing this check might do more harm than good. 

We generally have 5-10 new teams every week, and ~150 teams in total. Adding the first check alone brings down the number of DB calls from 300 (150 * 2) to 20 (10 * 2). Removing this second check would bring that number down to 10 - but wouldn't give us the flexibility to blindly call `Team.migrate_tips` over the console. Also, the `migrate-tips.py` script isn't covered by our test suite, but `Team.migrate_tips` is. 

Thoughts?

bin/migrate-tips.py
```diff
@@ -4,9 +4,15 @@
 db = db(env())
 
 slugs = db.all("""
-    SELECT slug
-      FROM teams
-     WHERE is_approved IS TRUE
+    SELECT distinct(t.slug)
+      FROM teams t
+      JOIN tips ON t.owner = tips.tippee
```
rohitpaulk: It'd be great if we could add comments here to explain what we're filtering based on. For example, 

screen shot 2016-01-24 at 12 53 36 pm

gratipay/models/team.py
```diff
@@ -158,19 +158,6 @@ def to_dict(self):
...
-            SELECT pi.*
-              FROM payment_instructions pi
-              JOIN teams t ON t.slug = pi.team
-              JOIN participants p ON t.owner = p.username
-             WHERE p.username = %s
-               AND pi.ctime < t.ctime
-        """, (self.owner, ))
-
```
aandis: @rohitpaulk in case we decide to keep this, I think the `JOIN participants p ON t.owner = p.username WHERE p.username = %s` is unnecessary. A simple `where t.owner = %s` should suffice.

rohitpaulk: Yep, :+1:

bin/migrate-tips.py
```diff
@@ -3,14 +3,20 @@
...
-    SELECT slug
-      FROM teams
-     WHERE is_approved IS TRUE
+teams = db.all("""
+    SELECT distinct ON (t.slug) t.*::teams
+      FROM teams t
+      JOIN tips ON t.owner = tips.tippee	-- Only fetch teams whose owner have tips.
+     WHERE t.is_approved IS TRUE           	-- Only fetch approved teams.
+       AND NOT EXISTS (                  	-- Make sure not already migrated.
+            SELECT 1
```
aandis: Dammit why doesn't this align. Looks fine in the editor. :angry:

aandis:
@rohitpaulk changes made.

aandis:
Let me know if I should undo 0707b592a2068a463724603ab65a6b3d052640f8

rohitpaulk:
> Let me know if I should undo 0707b59

I think it'd make sense to undo it.

bin/migrate-tips.py
aandis: @rohitpaulk do you have any idea why the comment looks funny here but looks fine in the editor?
screenshot from 2016-01-24 15 25 09

bin/migrate-tips.py
```diff
@@ -3,14 +3,20 @@
...
+               AND pi.ctime < t.ctime
+	     )
 """)
 
-for slug in slugs:
-    team = Team.from_slug(slug)
+for team in teams:
     try:
         ntips = team.migrate_tips()
         print("Migrated {} tip(s) for '{}'".format(ntips, slug))
```
rohitpaulk: Replace `slug` with `team.slug`

aandis: oops. missed that.

bin/migrate-tips.py
rohitpaulk: Most likely spaces vs tabs..

rohitpaulk: Yep, confirmed. 

screen shot 2016-01-24 at 3 31 37 pm

This is sublime text - See the difference? The `.`s are spaces, and `-`s are tabs

aandis: Changed it to only spaces. Should work now.

aandis:
@rohitpaulk done! Since `migrate-tips.py` isn't covered by the test suite, can you somehow run this and test that the new query works as expected? If it does, I think we are good to go.

gratipay/models/team.py
rohitpaulk: Oh I'm sorry - I was wrong here. The `JOIN` statement accounts for subscriptions for all teams a user owns. 

If a user creates Team A, we migrate tips and then he creates Team B - we don't want to migrate tips again.

@aandis - Can you fix this and also add a test to ensure that this doesn't happen again?

rohitpaulk:
I've been running this on my local, yes - That's how I discovered https://github.com/gratipay/gratipay.com/pull/3904#discussion_r50633149 :)

Are you not able to run this script yourself? Is fake data completely broken?

rohitpaulk:
Once I have fake data in the DB, I use `./env/bin/honcho run -e defaults.env ./env/bin/python ./bin/migrate-tips.py` to run the script

rohitpaulk:
I've added a comment at https://github.com/gratipay/gratipay.com/pull/3904#discussion_r50633536. Other than that, this is good to go

gratipay/models/team.py
rohitpaulk: Ah sorry again. Ignore the above

aandis:
Thing is, my fake data doesn't have any teams which satisfy the two `where` conditions simultaneously. So it just exits with `Done.` Hence I can't be 100% sure it does what we expect it to do. I'll have to create all those rows manually first to test. :tired_face:
If you are able to run and see expected outputs, we can merge this.

rohitpaulk:
Works fine for me, yes - Travis seems to be complaining about something, can you fix that?

aandis:
[Builds](https://travis-ci.org/gratipay/gratipay.com/builds/104414859) are passing for this pr. What is wrong?

aandis:
Oh `import` error I guess.

rohitpaulk:
> Builds are passing for this pr.

They're still :red_circle:, we want :green_apple:!

aandis:
There's your green :apple: :stuck_out_tongue:

rohitpaulk:
@aandis - I've added a commit, go ahead and merge if it looks good to you

rohitpaulk:
http://sqlfiddle.com/#!15/20a13/5 

@aandis - Is that accessible to you?

aandis:
Yes, Checking it out.

rohitpaulk:
> teams.owner is unique too because it references participant.username. So a slug can only belong to one owner.

`teams.owner` need not be unique - by `referencing participant.username`, the only constraint we're putting on the column is that the value has to have a valid record in the participants table. It's possible for one owner to have multiple teams. Example: https://gratipay.com/~bbatsov/

aandis:
Yep. Got it. #3894 described restriction on `team` hence the misunderstanding.

aandis:
I'll merge this then. This was fun. :-)

aandis:
Ok merged. Do I change the version now? What is the criteria for changing version?

rohitpaulk:
> Do I change the version now? What is the criteria for changing version?

That happens automatically as part of the [release process](https://github.com/gratipay/gratipay.com/blob/master/release.sh) when we ship code to Heroku :)

aandis:
Cool! :+1:

Issue #3460, opened 2015-05-20
subscriptions.json


whit537:
We killed the tips.json API somewhere beneath #3399. We need to bring back a parallel API for `subscriptions`.

kaguillera:
This should be blocked by the fact that a team page is needed before we add a subscription this widget to the team page (that does not currently exist). 
Going to work on the Team profile page now

rohitpaulk:
!m @kaguillera

whit537:
I found a pretty nice way (IMO) to write literate APIs, but it got buried in a dead-end PR (see https://github.com/gratipay/gratipay.com/pull/3786#issuecomment-141134715 and, e.g., [`www/api/v2/index.spt`](https://github.com/gratipay/gratipay.com/pull/3786/files#diff-ca4234e057355f5f591002bdb5738a58R1)). I have it in mind to rescue that work some day. Check it out if you're interested—but don't get too hung up on it if you're not. ;-)

rohitpaulk:
Woah, that's some nice stuff :)

whit537:
Thanks! :-) I thought it was a really interesting way to use simplates to integrate the API and documentation with the site itself.

aandis:
Thinking of working on this next. Can someone point me in the right direction? :)
What are `subscriptions`? I don't see any relation in `schema.sql`. And what should `subscriptions.json` do?

rohitpaulk:
@aandis - `subscriptions` is the user-facing term for `payment_instructions` in our schema. We used to offer an API endpoint for users to view and modify their tips (authenticated with their API keys). When we moved from tips to subscriptions (Gratipay 1.0 to Gratipay 2.0) - in the interest of velocity we didn't update the endpoint, thought we'd get back to it later someday (TODAY!). 

So I poked up the the oldest tag (version) we have - https://github.com/gratipay/gratipay.com/tree/1840

I'd start by looking at `%username/tips.json` in the documentation under  /https://github.com/gratipay/gratipay.com/tree/1840#api, and its [corresponding implementation](https://github.com/gratipay/gratipay.com/blob/1840/www/~/%25username/tip.json.spt).

You could try checking out that version of the codebase, firing up the server and actually hitting the API to understand what it does - and maybe suggest improvements too :)

rohitpaulk:
[...]

Issue #412, opened 2012-12-07
use cents instead of decimal


chadwhitacre:
I'm using Decimal everywhere for dollar values but it turns out that best practice is to deal with cents. That's how the Balanced API is set up (Stripe, too). It makes it much easier cause then we can use plain ints instead of more complicated datatypes, in both Python and Postgres.

PR #3909, opened 2016-01-30
Subscriptions api

aandis:
Still WIP.

Added `GET` to show all subscriptions of a user along with filtering on team slug.
References #3460

### DIFF
--- a/README.md
+++ b/README.md
@@ -504,9 +504,20 @@ an object giving a point-in-time snapshot of Gratipay. The
           - `http://www.openstreetmap.org/user/%openstreetmap_username`
 
 **/`%username`/payment-instructions.json**
-([source](https://github.com/gratipay/gratipay.com/www/~/%username/payment-instructions.json.spt))&mdash;*private*&mdash;Responds to `GET` with an array of objects representing your current payment instructions. A payment instruction is created when a ~user instructs Gratipay to make voluntary payments to a Team. Pass a `team_slug` with `GET` to fetch payment instruction only for that particular team. `POST` an array of objects containing `team_slug` and `amount` to bulk upsert payment instructions(make sure to set `Content-Type` to `application/json`). The `amount` must be encoded as a string rather than a number. In case the upsert is not successful for any object, there will be an `error` attribute in the response explaining the error along with the `team_slug` to identify the object for which the error occured.
+([source](https://github.com/gratipay/gratipay.com/www/~/%username/payment-instructions.json.spt))&mdash;*private*&mdash;Responds
+to `GET` with an array of objects representing your current payment
+instructions. A payment instruction is created when a ~user instructs Gratipay
+to make voluntary payments to a Team. Pass a `team_slug` with `GET` to fetch
+payment instruction only for that particular team. `POST` an array of objects
+containing `team_slug` and `amount` to bulk upsert payment instructions (make
+sure to set `Content-Type` to `application/json`). The `amount` must be encoded
+as a string rather than a number. In case the upsert is not successful for any
+object, there will be an `error` attribute in the response explaining the error
+along with the `team_slug` to identify the object for which the error occured.
 
-This endpoint requires authentication. Look up your user ID and API key on your [account page](https://gratipay.com/about/me/settings/) and pass them using basic auth.
+This endpoint requires authentication. Look up your user ID and API key on your
+[account page](https://gratipay.com/about/me/settings/) and pass them using
+basic auth.
 
 E.g.:
 Request
--- a/README.md
+++ b/README.md
@@ -503,6 +503,36 @@ an object giving a point-in-time snapshot of Gratipay. The
           - `undefined` (key not present)&mdash;no OpenStreetMap account connected
           - `http://www.openstreetmap.org/user/%openstreetmap_username`
 
+**/`%username`/payment-instructions.json**
+([source](https://github.com/gratipay/gratipay.com/www/~/%username/payment-instructions.json.spt))&mdash;*private*&mdash;Responds to `GET` with an array of objects representing your current payment instructions. A payment instruction is created when a ~user instructs Gratipay to make voluntary payments to a Team. Pass a `team_slug` with `GET` to fetch payment instruction only for that particular team. `POST` an array of objects containing `team_slug` and `amount` to bulk upsert payment instructions(make sure to set `Content-Type` to `application/json`). The `amount` must be encoded as a string rather than a number. In case the upsert is not successful for any object, there will be an `error` attribute in the response explaining the error along with the `team_slug` to identify the object for which the error occured.
+
+This endpoint requires authentication. Look up your user ID and API key on your [account page](https://gratipay.com/about/me/settings/) and pass them using basic auth.
+
+E.g.:
+Request
+
+```
+curl -L https://gratipay.com/username/payment-instructions.json \
+     -u $userid:$api_key \
+     -X POST \
+     -d '[{"amount": "1.00", "team_slug": "foobar"}]' \
+     -H "Content-Type: application/json"
+```
+
+Response
+
+```
+[
+    {
+        "amount": "1.00",
+        "ctime": "2016-01-30T12:38:00.182230+00:00",
+        "due": "0.00",
+        "mtime": "2016-02-06T14:37:28.532508+00:00",
+        "team_name": "Foobar team",
+        "team_slug": "foobar"
+    }
+]
+```
 
 API Implementations
 -------------------
--- a/tests/py/test_payment_instructions_json.py
+++ b/tests/py/test_payment_instructions_json.py
@@ -1,0 +1,199 @@
+import json
+
+from gratipay.testing import Harness
+
+class TestPaymentInstructionApi(Harness):
+
+    def test_get_with_team_filter(self):
+        "Test that GET with team_slug passed returns correct response."
+
+        alice = self.make_participant("alice", claimed_time='now')
+        Enterprise = self.make_team("The Enterprise", is_approved=True)
+        Trident = self.make_team("The Trident", is_approved=True)
+
+        alice.set_payment_instruction(Enterprise, '10.0')
+        alice.set_payment_instruction(Trident, '12.0')
+
+        data = json.loads(self.client.GET(
+             "~/alice/payment-instructions.json?team_slug=" + Enterprise.slug
+            , auth_as='alice').body)
+
+        assert data['team_slug'] == Enterprise.slug
+        assert data['team_name'] == Enterprise.name
+        assert data['amount'] == '10.00'
+        assert 'ctime' in data
+        assert 'mtime' in data
+        assert 'due' in data
+
+    def test_get_with_team_filter_raises_error_on_invalid_team_slug(self):
+        self.make_participant("alice", claimed_time = 'now')
+
+        response = self.client.GxT(
+             "~/alice/payment-instructions.json?team_slug=no-team"
+            , auth_as='alice')
+
+        assert response.code == 400
+
+    # pi => payment instruction.
+
+    def test_get_with_team_filter_returns_default_if_no_pi(self):
+        self.make_participant("alice", claimed_time='now')
+        Enterprise = self.make_team("The Enterprise", is_approved=True)
+
+        data = json.loads(self.client.GET(
+             "~/alice/payment-instructions.json?team_slug=" + Enterprise.slug
+            , auth_as='alice').body)
+
+        assert data['team_slug'] == Enterprise.slug
+        assert data['team_name'] == Enterprise.name
+        assert data['amount'] == '0.00'
+        assert data['due'] == '0.00'
+        assert data['ctime'] == None
+        assert data['mtime'] == None
+
+    def test_simple_get(self):
+        "Test that GET without any parameters returns correct response."
+
+        alice = self.make_participant("alice", claimed_time='now')
+        Enterprise = self.make_team("The Enterprise", is_approved=True)
+        Trident = self.make_team("The Trident", is_approved=True)
+
+        data = json.loads(self.client.GET(
+             "~/alice/payment-instructions.json", auth_as='alice').body)
+
+        assert len(data) == 0 # Empty 'array' should be returned.
+
+        alice.set_payment_instruction(Enterprise, '10.0')
+        alice.set_payment_instruction(Trident, '12.0')
+
+        data = json.loads(self.client.GET(
+             "~/alice/payment-instructions.json", auth_as='alice').body)
+
+        assert len(data) == 2
+
+        assert data[0]['team_slug'] == Trident.slug # response is ordered by amount desc
+        assert data[0]['amount'] == '12.00'
+        assert data[0]['team_name'] == Trident.name
+
+        assert data[1]['team_slug'] == Enterprise.slug
+        assert data[1]['amount'] == '10.00'
+        assert data[1]['team_name'] == Enterprise.name
+
+        for d in data:
+            assert 'due' in d
+            assert 'ctime' in d
+            assert 'mtime' in d
+ 
+    def test_post(self):
+        "Test that POST to this endpoint works correctly."
+
+        self.make_participant("alice", claimed_time='now')
+        Enterprise = self.make_team("The Enterprise", is_approved=True)
+        Trident = self.make_team("The Trident", is_approved=True)
+
+        request_body = [
+            { 'amount': "1.50", 'team_slug': Enterprise.slug },
+            { 'amount': "39.50", 'team_slug': Trident.slug }
+        ]
+
+        response = self.client.POST( "~/alice/payment-instructions.json"
+                                    , body=json.dumps(request_body)
+                                    , content_type='application/json'
+                                    , auth_as='alice')
+
+        assert response.code == 200
+
+        # Make sure apt response returned.
+        data = json.loads(response.body)
+
+        assert len(data) == 2
+
+        assert data[0]['team_slug'] == Enterprise.slug
+        assert data[0]['team_name'] == Enterprise.name
+        assert data[0]['amount'] == '1.50'
+
+        assert data[1]['team_slug'] == Trident.slug
+        assert data[1]['team_name'] == Trident.name
+        assert data[1]['amount'] == '39.50'
+
+        for d in data:
+            assert 'due' in d
+            assert 'ctime' in d
+            assert 'mtime' in d
+
+        # Make sure actually written to database.
+        data = json.loads(self.client.GET(
+             "~/alice/payment-instructions.json", auth_as='alice').body)
+
+        assert data[0]['team_slug'] == Trident.slug
+        assert data[0]['amount'] == '39.50'
+
+        assert data[1]['team_slug'] == Enterprise.slug
+        assert data[1]['amount'] == '1.50'
+
+    def test_post_with_no_team_slug_key_returns_error(self):
+        self.make_participant("alice", claimed_time='now')
+
+        response = self.client.POST( "~/alice/payment-instructions.json"
+                                    , body=json.dumps([{ 'amount': "1.50" }])
+                                    , content_type='application/json'
+                                    , auth_as='alice')
+
+        assert response.code == 200 # Since batch processing.
+        assert 'error' in json.loads(response.body)[0]
+
+        data = json.loads(self.client.GET(
+             "~/alice/payment-instructions.json", auth_as='alice').body)
+        assert len(data) == 0
+
+    def test_post_with_no_amount_key_returns_error(self):
+        self.make_participant("alice", claimed_time='now')
+        Enterprise = self.make_team("The Enterprise", is_approved=True)
+
+        response = self.client.POST( "~/alice/payment-instructions.json"
+                                    , body=json.dumps(
+                                        [{ 'team_slug': Enterprise.slug }])
+                                    , content_type='application/json'
+                                    , auth_as='alice')
+
+        assert response.code == 200
+        assert 'error' in json.loads(response.body)[0]
+
+        data = json.loads(self.client.GET(
+             "~/alice/payment-instructions.json", auth_as='alice').body)
+        assert len(data) == 0
+
+    def test_adding_pi_for_invalid_team_returns_error(self):
+        self.make_participant("alice", claimed_time='now')
+
+        request_body = [{ 'team_slug': 'no-slug', 'amount': '39.50' }]
+
+        response = self.client.POST( "~/alice/payment-instructions.json"
+                                    , body=json.dumps(request_body)
+                                    , content_type='application/json'
+                                    , auth_as='alice')
+
+        assert response.code == 200
+        assert 'error' in json.loads(response.body)[0]
+
+        data = json.loads(self.client.GET(
+             "~/alice/payment-instructions.json", auth_as='alice').body)
+        assert len(data) == 0
+
+    def test_adding_pi_for_unapproved_team_returns_error(self):
+        self.make_participant("alice", claimed_time='now')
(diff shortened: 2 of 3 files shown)


aandis:
From https://github.com/gratipay/gratipay.com/issues/3460#issuecomment-176405556

> One thing we had on the old tips.json was the ability to bulk update a bunch of tips in one POST. The Khan Academy setup used that, e.g.

[payment_instruction.json](https://github.com/gratipay/gratipay.com/blob/master/www/%25team/payment-instruction.json.spt) doesn't support batch subscription creation and it also exposes some extra information not relevant to this endpoint. So, I thought I'd leave it alone and ditch the redirection thing and add all functionality directly to this.

aandis:
So if that sounds good @rohitpaulk , @whit537, I'll go ahead and `POST` to this

aandis:
Would appreciate a quick code review as well. :)

www/~/%username/subscriptions.json.spt
```diff
@@ -0,0 +1,41 @@
...
+from aspen import Response
+from gratipay.exceptions import NoTeam
+
+[------------------------------------------------------------------------]
+
+if user.ANON:
+    raise Response(403, _("Please sign in first."))
+
+else:
+    participant = user.participant
```
rohitpaulk: Use [`get_participant`](https://github.com/gratipay/gratipay.com/blob/master/gratipay/utils/__init__.py#L56) here to make sure the user exists, and that the authenticated user has the necessary permissions to access the endpoint. The current way this is implemented, looks like hitting `www/~/aandis/subscriptions.json.spt` is not going to give me (logged in as `rohitpaulk`) a 403/404, it is going to return my own subscriptions.

www/~/%username/subscriptions.json.spt
```diff
@@ -0,0 +1,41 @@
...
+
+    if request.method == 'GET':
+        slug = request.qs.get("team_slug", "")
+
+        if slug: # Get subscription info only for this team.
+            try:
+                subscription = participant.get_payment_instruction(slug)
+            except NoTeam:
+                raise Response(400, _("Invalid team name."))
+            out = {
```
rohitpaulk: We should maintain consistency in the format of what we return for the batch endpoint vs filtered endpoint. If we're returning 6 keys per item in the batch endpoint, we should be returning 6 keys per item in the filtered endpoint too... I think the best way to do this is to call `get_giving_for_profile` anyway, and filter the results in Python.

www/~/%username/subscriptions.json.spt
```diff
@@ -0,0 +1,41 @@
...
+                     "mtime" : subscription['mtime']
+                  }
+        else: # Get subscription info for all teams.
+            subscriptions, totals = participant.get_giving_for_profile()
+
+            out = []
+            for s in subscriptions:
+                out.append({ "team_slug": s.team_slug,
+                             "team_name": s.team_name,
+                                "amount": str(s.amount),
```
rohitpaulk: I'm curious here - is there a reason that you chose to expose `amount` and `due` as strings? JSON does support number types, you know :stuck_out_tongue:

www/~/%username/subscriptions.json.spt
aandis: Oh yeah. I was yet to change that.

www/~/%username/subscriptions.json.spt
aandis: Hmm yeah. Makes sense.

www/~/%username/subscriptions.json.spt
aandis: It was all over the code base. Eg., [this](https://github.com/gratipay/gratipay.com/blob/master/www/%25team/payment-instruction.json.spt) , [this](https://github.com/gratipay/gratipay.com/blob/1840/www/~/%25username/tip.json.spt) and [this](https://github.com/gratipay/gratipay.com/blob/1831/www/~/%25username/tips.json.spt). I was just being consistent. :stuck_out_tongue_closed_eyes: Figured it'd be some python thing.

Changaco: Floating-point math leads to rounding errors, so Gratipay's API sends back money amounts as strings instead of floats to discourage misuse.

rohitpaulk: Aha :)

aandis:
@rohitpaulk `POST` still doesn't yet work for me. Do you have any idea why? See https://github.com/gratipay/gratipay.com/issues/3460#issuecomment-176381669

www/~/%username/subscriptions.json.spt
whit537: Related history: at one point we considered switching to cents (#412).

aandis:
Added changes discussed. Yet to find a way around https://github.com/gratipay/gratipay.com/issues/3460#issuecomment-176381669

aandis:
Alright. Fixed https://github.com/gratipay/gratipay.com/issues/3460#issuecomment-176381669. :dancer:  Ended up modifying [this](https://github.com/gratipay/aspen.py/blob/db93d839529d98e10fe1444188bd7ece661e49d7/aspen/website.py#L79) after digging for a couple of hours. Looks [correct](https://www.w3.org/Protocols/rfc2616/rfc2616-sec10.html#sec10.3.8) to me. Not really sure if this will happen on prod too. If it does, we should fix this in aspen. 
Anyway, I'll continue with `POST` later today! :)

aandis:
`POST` added. :dancer: Review needed. 

btw Fixing https://github.com/gratipay/gratipay.com/issues/3460#issuecomment-176381669 made github and other provider's sign up redirects to stop working. I am sensing it's an [ssl](http://stackoverflow.com/questions/12548341/how-to-work-around-post-being-changed-to-get-on-302-redirect) thing. But it's way easier for development compared to setting up ssl. :wink:

techtonik:
What users are subscribed to?

aandis:
Giving to teams @techtonik. See https://github.com/gratipay/gratipay.com/issues/3460 for details. :)

whit537:
> `POST` added. :dancer:

!m @aandis :)

whit537:
@aandis First thing I notice is that we'll need some tests for this endpoint.

aandis:
Yep. That's next. Figured I should first review then write tests. :)

techtonik:
That sounds strange. When I subscribe to something (especially for something), I expect to receive something. If I have a lot of subscribers, I automatically feel obliged to do something in exchange.

If we are still for "giving with no strings attached" then maybe there is a better term? Facebook has "liking", newspapers and lists have "subscribing", and when we send money the closest term is "supporting", but if we want more involvement and more positive value on the receiver's end, how do we call it? "links" to the causes, so we "link" people with "projects" and use "routes" to send them gratitude in the emotionless form of money that can be used to do something good for them. =)

whit537:
@aandis Gotcha.

whit537:
@techtonik You raise a good point: over in https://github.com/gratipay/inside.gratipay.com/issues/117 we decided to go with "payment instructions" instead of "subscriptions" (https://github.com/gratipay/inside.gratipay.com/issues/117#issuecomment-133030222 notwithstanding). Presumably #3460 predates that decision.

@aandis Let's go ahead and switch from "subscription" to "payment instructions" here. That'll line up with the database table and `set_payment_instruction` method name, as well.

whit537:
> You raise a good point

(Though we are no longer "for 'giving with no strings attached,'" post-https://github.com/gratipay/inside.gratipay.com/issues/180 and pre-https://github.com/gratipay/inside.gratipay.com/issues/423 .)

www/~/%username/subscriptions.json.spt
```diff
@@ -0,0 +1,100 @@
...
+participant = get_participant(state, restrict=True)
+
+if request.method == 'GET':
+
+    # Fetch all subscriptions for this user.
+    subscriptions, totals = participant.get_giving_for_profile()
+
+    slug = request.qs.get("team_slug", "")
+
+    if slug: # Filter subscription info only for this team from the list.
```
whit537: How is this intended to related to the existing [`%team/payment-instruction.json.spt`](https://github.com/gratipay/gratipay.com/blob/1939/www/%25team/payment-instruction.json.spt) API? Looks like it at least partially duplicates the functionality over there. Why?

www/~/%username/subscriptions.json.spt
```diff
@@ -0,0 +1,100 @@
...
+            one = format_error(s['team_slug'], "No amount.")
+        else:
+            team = Team.from_slug(s['team_slug'])
+            if team:
+                try:
+                    sub = participant.set_payment_instruction(
+                        s['team_slug'], parse_decimal(s['amount'])
+                    )
+                except Exception, exc:
+                    one = format_error(team.slug, exc.__class__.__name__)
```
whit537: Do we really have to do our own error handling here? What happens if we depend on higher-level error handling?

www/~/%username/subscriptions.json.spt
```diff
@@ -0,0 +1,100 @@
...
+[-----------------------------------------------------------------------------]
+
+def format_subscription(s):
+    return {
+        "team_name": s.team_name,
+        "team_slug": s.team_slug,
+           "amount": str(s.amount),
+              "due": str(s.due),
+            "ctime": s.ctime,
+            "mtime": s.mtime
```
whit537: Interesting! We've gone back and forth over [leading-comma](https://gist.github.com/isaacs/357981) style [before](https://botbot.me/freenode/gratipay/2014-01-13/?msg=9719489&page=4), but this style is new to me. Our SQL [formatting](https://github.com/gratipay/gratipay.com/pull/3896#discussion_r50124769) has inspired you, I take it? :-)

www/~/%username/subscriptions.json.spt
aandis: Only thing similar between them as of now is `GET` with team filter. But [team/payment-instruction.json.spt](https://github.com/gratipay/gratipay.com/blob/1939/www/%25team/payment-instruction.json.spt) returns a different response which isn't really relevant to this api. This endpoint offers an additional `GET` without a team filter to show all subscriptions of a user.  Per https://github.com/gratipay/gratipay.com/pull/3909#discussion_r51345897, we should maintain consistency in the api responses. Hence everything is implemented right here.

www/~/%username/subscriptions.json.spt
aandis: If you mean the `try except` block, it's necessary because we are processing multiple subscription creation here. We may end up successfully creating some and failing to create others. So it's important to return an apt error response for each new failed subscription request. Just like [tips.json](https://github.com/gratipay/gratipay.com/blob/1831/www/~/%25username/tips.json.spt)

www/~/%username/subscriptions.json.spt
aandis: Yesh! It's pretty neat! :smiley:

aandis:
> Let's go ahead and switch from "subscription" to "payment instruction"

Alright, will do. :)

aandis:
Tests added. :dancer:

aandis:
Also renamed `subscriptions` to `payment_instructions`. I hope this doesn't get too confusing though. Because we already have [%25team/payment-instruction.json.spt](https://github.com/gratipay/gratipay.com/blob/1939/www/%25team/payment-instruction.json.spt). imo we should rename that to `subscriptions` because `teams` have subscriptions (hence they [receive](https://github.com/gratipay/gratipay.com/pull/3909#issuecomment-178179474)) and `payment-instructions` is an appropriate name for this endpoint.

whit537:
> Also renamed `subscriptions` to `payment_instructions`.

Actually, it looks like it's renamed to [`payment_instruction`](https://github.com/gratipay/gratipay.com/blob/eebf8f854c7ae5ca1daa22ca529f3248e33f7934/www/~/%25username/payment-instruction.json.spt) (singular). Eh?

> `teams` have subscriptions

On https://github.com/gratipay/inside.gratipay.com/issues/117 we decided that teams *don't* have `subscriptions`, they have `payment instructions` (see https://github.com/gratipay/gratipay.com/pull/3652 for the table name change).

> (hence they [receive](https://github.com/gratipay/gratipay.com/pull/3909#issuecomment-178179474))

"['Receiving' refers to payments from the Team's point of view.](https://github.com/gratipay/inside.gratipay.com/issues/117#issuecomment-126369596)"—and "payment instruction" is what we call the direction we have from a `~user` to move money from them to a `Team`.

whit537:
> I hope this doesn't get too confusing though. 

Yeah, let's try to get clear on what we're talking about here. :-)

Let's back up for a second: note/recall that under Gratipay 1.0, any user could give directly to any other user. Therefore, the pattern of the old endpoints was:

 1. [`%username/tip.json`](https://github.com/gratipay/gratipay.com/blob/1831/www/~/%25username/tip.json.spt)—**singular**—represented the authenticated user's tip to `%username`
 1. [`%username/tips.json`](https://github.com/gratipay/gratipay.com/blob/1831/www/~/%25username/tips.json.spt)—**plural**—represented all of `%username`'s tips to other users 

With [Gratipay 2.0](https://medium.com/gratipay-blog/gratipay-2-0-2453d3c53077), we promoted Teams to first-class entities, and now `~user`s can only give to `Team`s. If we were to preserve the above semantics, then we would expect:

 1. [`%team/payment-instruction.json.spt`](https://github.com/gratipay/gratipay.com/blob/1939/www/%25team/payment-instruction.json.spt)—**singular, under `%team`**—represents the authenticated user's payment instruction for `%team`
 1. `~/%username/payment-instructions.json.spt`—**plural, under `~/%username`**—would represent *all* of `%username`'s payment instructions to Teams

**But! ...**

whit537:
I think what I hear you proposing, @aandis, is that we *change* the semantics ... which might make sense.

> we already have [%25team/payment-instruction.json.spt](https://github.com/gratipay/gratipay.com/blob/1939/www/%25team/payment-instruction.json.spt). imo we should rename that to `subscriptions`

If I'm hearing you right, this would actually be a new *receiving* endpoint for Teams, which would be related to @rohitpaulk's work over on #3873 (where I believe it's actually covered by `charts.json.spt`). Does that sound right?

> and `payment-instructions` is an appropriate name for this endpoint.

It seems to me like you're subconsciously trying to treat `~/%username/payment-instructions.json.spt` as *combining* the two endpoints previously distinguished as `tip.json` and `tips.json` under Gratipay 1.0. For example, the [`if slug` block](https://github.com/gratipay/gratipay.com/blob/eebf8f854c7ae5ca1daa22ca529f3248e33f7934/www/~/%25username/payment-instruction.json.spt#L31-L55) in `~/%username/payment-instruction.json.spt` duplicates the `GET` functionality already [present](https://github.com/gratipay/gratipay.com/blob/1939/www/%25team/payment-instruction.json.spt#L35) in `%team/payment-instruction.json.spt`, and of course it was possible to update a single tip using either endpoint under 1.0.

@aandis Is that a fair interpretation?

aandis:
>> Also renamed subscriptions to payment_instructions.

> Actually, it looks like it's renamed to payment_instruction (singular). Eh?

Ah sorry. That was a typo in the comment. Per https://github.com/gratipay/gratipay.com/pull/3909#issuecomment-178181473, I renamed it to `payment-instruction`. What I essentially meant was, `users` give out `payment_instructions` and `teams` *receive* money as a result of those `payment_instructions` (on a weekly basis). This, in a way means, `teams` are *subscribed* to those `payment_instructions`. Hence the name `payment_instruction.json` would make sense when hitting the %username end point(because we'd only want to know a user's `payment_instructions` and/or create new ones) and `subscription.json` would make sense when hitting the %team endpoint(to see what `payment_instructions` are they subscribed to, hence receiving money from). 

But.. 

> 1. %team/payment-instruction.json.spt—singular, under %team—represents the authenticated user's payment instruction for %team
2. ~/%username/payment-instructions.json.spt—plural, under ~/%username—would represent all of %username's payment instructions to Teams

This sounds good enough to me as well. Though the two closely related names could confuse newcomers and are a little hard to converse with ;) . But as long as we are following a convention and it's properly documented, I guess it should be okay.

> It seems to me like you're subconsciously trying to treat ~/%username/payment-instructions.json.spt as combining the two endpoints

I don't intend to combine the two endpoints because `payment-instruction.json.spt`(singular) is only used internally to serve specific data for a particular [view](https://github.com/gratipay/gratipay.com/blob/master/templates/your-payment.html) and possibly to more [views](https://github.com/gratipay/gratipay.com/pull/3873#issuecomment-164159904) in the future. Whereas, this new end point would serve as an api for other people to use to build their own apps and it returns data only specific to this endpoint.

aandis:
tl; dr

> 1. %team/payment-instruction.json.spt—singular, under %team—represents the authenticated user's payment instruction for %team
2. ~/%username/payment-instructions.json.spt—plural, under ~/%username—would represent all of %username's payment instructions to Teams

I'm okay with this if you are. :)

whit537:
> I'm okay with this if you are.

Okay, in that case, do we need the `if slug` block under `GET`? :)

whit537:
> only used internally

@aandis I think it's best to have one API that both we and third parties use. Maybe we should:

 1. Drop the singular `%team/payment-instruction.json.spt`
 1. Change our views that use the singular endpoint to use the plural `~/%username/payment-instructions.json.spt` instead.

Eh?

aandis:
> Okay, in that case, do we need the if slug block under GET?

It's an added feature which makes sense with this endpoint. But with `%team/payment-instruction.json.spt` also doing similar stuff, I can see why you would want it removed. I'll get rid of it if you'd like.

> I think it's best to have one API that both we and third parties use.

Ideally yes. I haven't yet explored in depth which page utilizes the data served by `%team/payment-instruction.json.spt` and how, so I can't really say for sure how the transition would work out. In any case, I think this deserves a ticket of it's own. :)

whit537:
> I'll get rid of it if you'd like.

I guess I flip-flopped there: I think it makes sense to *keep* the `if slug` behavior on the new plural endpoint, and *remove* the singular endpoint. The removal can happen either in this PR or in a separate one.

tests/py/test_payment_instruction_api.py
```diff
@@ -0,0 +1,199 @@
...
+from gratipay.testing import Harness
+
+class TestPaymentInstructionApi(Harness):
+
+    def test_get_with_team_filter(self):
+        "Test that GET with team_slug passed returns correct response."
+
+        alice = self.make_participant("alice", claimed_time='now')
+        gratipay = self.make_team("gratipay", is_approved=True)
+        gittip = self.make_team("gittip", is_approved=True)
```
whit537: A blast from the past! :-)

In other parts of the test suite we use `Enterprise` and `Trident` as example Teams.

whit537: > In other parts of the test suite we use `Enterprise` and `Trident` as example Teams.

Aaaaand let's do that here as well. :)

tests/py/test_payment_instruction_api.py
```diff
@@ -0,0 +1,199 @@
...
+
+        alice.set_payment_instruction(gratipay, '10.0')
+        alice.set_payment_instruction(gittip, '12.0')
+
+        data = json.loads(self.client.GET(
+             "~/alice/payment-instruction.json?team_slug=" + gratipay.slug
```
whit537: This should be plural: `payment-instructions`.

whit537:
#### Punchlist

 - [ ] rename endpoint to plural: `payment-instructions.json`—https://github.com/gratipay/gratipay.com/pull/3909#discussion_r52055526
 - [ ] don't use `gittip` in test code; `Enterprise` and `Trident` are the suggested test Teams—https://github.com/gratipay/gratipay.com/pull/3909#discussion_r52055367
 - [ ] since this is intended to be public API, let's update the API section of the README

aandis:
> The removal can happen either in this PR or in a separate one.

I think it's better to deploy the new endpoint first and have a separate ticket/pr for the transition.

whit537:
> I think it's better to deploy the new endpoint first and have a separate ticket/pr for the transition.

Works for me. !m @aandis

tests/py/test_payment_instruction_api.py
aandis: Haha ok. :)

aandis: How about [Defiant](https://en.wikipedia.org/wiki/USS_Defiant) and [Voyager](https://en.wikipedia.org/wiki/USS_Voyager_(Star_Trek)) too? :stuck_out_tongue:

tests/py/test_payment_instruction_api.py
aandis: Yeah. I am yet to do the renaming.

tests/py/test_payment_instruction_api.py
whit537: Now you're talkin'! :star: :tractor:

aandis:
Alright, done with https://github.com/gratipay/gratipay.com/pull/3909#issuecomment-180495892 - 1 and 2. I'll update the read me tomorrow.

aandis:
Wanted to write a test for https://github.com/gratipay/gratipay.com/pull/3909/files#diff-703cef3cc133ab1b5302e8bb02db5cb1R96 but came across https://github.com/gratipay/aspen.py/issues/514. Any particular reason just doing the following not correct?
screenshot from 2016-02-06 01 15 00

Tests I wrote after doing the above passed (not included in this commit obviously).

whit537:
@aandis Yeah that's most of it. We'd need some tests to land a PR in Aspen, and Aspen is undergoing [big changes](https://github.com/gratipay/aspen.py/issues/526) right now so you can decide if you want to spend time on that. :) The workaround would be to do `hxt('PATCH', ...)`.

aandis:
Cool. I'll look into it. Thanks :)

aandis:
Readme updated. :dancer:

whit537:
@aandis Nice! Can you rebase this PR branch onto the latest master? Seems we have conflicts.

aandis:
:+1:

whit537:
README looks fine. I hard-wrapped the new paragraphs and added a missing space in 63c74dd41762bfba12e963c6ca885717a559f74d.

I think this is ready to go!

Except that ... @rohitpaulk Didn't you say somewhere that you want to move to `/api/v1/` style endpoints? Was that a blocker for this PR?

rohitpaulk:
> Was that a blocker for this PR?

Not a blocker

whit537:
!m @aandis

aandis:
@whit537 If you have deployed this, can you confirm that `POST` to this endpoint works? I think I am still seeing signs of https://github.com/gratipay/gratipay.com/issues/3460#issuecomment-176381669

whit537:
@aandis It's working for me (and I found a good example of #3876 ;).

screen shot 2016-02-09 at 3 32 58 pm

screen shot 2016-02-09 at 3 32 42 pm

whit537:
(The second payment instruction is the one that changes in those screenshots—I guess including 0.00 tips is okay, so we can use this for the "Canceled" tab on the Giving page if we want. The first one to sudo room demonstrates the runaway dues.)

aandis:
Heh okay! I see now. All this time I had been hitting `/~/aandis/payment-instructions.json` instead of `/~aandis/payment-instructions.json`. The former redirects and becomes a `GET`.

aandis:
Learnt about http redirect methods in depth while overcoming this though. So all good in the end I guess. ;)

aandis:
The doc instructs to hit `/aandis/payment-instructions.json` which also doesn't work. Should change that.

aandis:
Ideally though, all these routes *should* work with `POST ` too because a `GET ` to any of them will give you identical response.

whit537:
> All this time I had been hitting `/~/aandis/payment-instructions.json`

Oh no! :scream_cat: Glad you figured it out. :)

> The doc instructs to hit /aandis/payment-instructions.json which also doesn't work. Should change that.

#3916, ya?

Issue #3684, opened 2015-08-17
Use Splinter for web tests


techtonik:
Currently I don't know who knows how our web test work, because if is based on JavaScript stack. The Splinter allows to write the same stuff in Pythonic way.

https://splinter.readthedocs.org/en/latest/index.html

whit537:
I started looking into this, and it's an interesting option. We have very few jstests currently. The main piece of work would be refactoring the Python test infrastructure to provide a browser instance, and, in particular, wiring up authentication (the in-process testing harness and the jstests both provide examples of this). Porting over the five tests we currently have would be trivial, and if we did so we could have a single test suite, which would greatly facilitate in-browser tests. Good idea! :-)

!m @techtonik

whit537:
Having the js test infrastructure in Python would make it easier for our Python-centric team to fix problems like #2953. Other jstest issues: #3069 #3493.

whit537:
@deivid-rodriguez Any interest in tackling this? :)

deivid-rodriguez:
Yes! I don't know how long it will take to acquire the knowledge to tackle this, but I will definitely keep this on my radar and give it a try!

whit537:
PR in #4079.

whit537:
🎉

PR #3155, opened 2015-02-05
Clean up assets before exiting

Changaco:
Fixes #3024.

Issue #2953, opened 2014-11-19
jstests fail when Gratipay is already running


whit537:
Reticketing from https://github.com/gratipay/gratipay.com/pull/2947#issuecomment-63671656 per [IRC](https://botbot.me/freenode/gratipay/2014-11-19/?msg=25812029&page=2).
##

deivid-rodriguez:
I fixed this yesterday by adding the `stop` _task_ as a _prerrequisite_ for the `jstests` _task_ (not sure what the `make` terminology is here, I'm using [`rake`](https://github.com/ruby/rake) terminology, but I can add a diff if what I mean is not clear!). Unfortunately it broke the travis build. If this is considered a good approach to fixing this, I could dedicate some more time to figure out why the build broke.

whit537:
@deivid-rodriguez As you've started to discover on #3911, our js test suite is not very mature yet. I see #3684 as a longer-term way to make js tests more integral to the project, and I believe it might fix this present ticket as a side-effect. Do you have any interest in pursuing that approach instead?

P.S. the `make` terminology would be "target" instead of "task" (in our case, a "[phony target](http://www.gnu.org/software/make/manual/make.html#Phony-Targets)" since it doesn't point to an actual file on the filesystem), and "dependency". "Prerequisite" or "dependency" are fine.

deivid-rodriguez:
Oh yes, I know remember those terms from my C days! :)

Yes, it looks like approaching #3684 is better than doing extra work that _might_ get overwritten by it! I'll have a look at it!

whit537:
Awesome, thanks! :-)

whit537:
I'm pretty sure our infrastructure has changed enough that his is now moot.

PR #3911, opened 2016-02-04
First time setup fixes

deivid-rodriguez:
Hi friends!

This PR is just an excuse to say hi! I've loved gratipay since a long time ago and really enjoy your way of building things. And I want to offer some help! These are the changes I made to get the app up and running and the tests passing. I must say [I cheated a bit](https://github.com/gratipay/gratipay.com/commit/ae5d9d291616477bd22931159a5457c01f2faae3)...

I also wanted to ask for some beginner pending tasks for someone who knows nothing about coding in python, but wants to learn. :)


whit537:
Yay! Thanks and welcome, @deivid-rodriguez! I've made an onboarding ticket for you at https://github.com/gratipay/inside.gratipay.com/issues/489. Feel free to use that to ask any random questions you have about Gratipay and how to contribute. :)

deivid-rodriguez:
Thanks @whit537 ! I'll ask there any questions I may have!

whit537:
Re: 89e037d9dbd4959a7f5dd7004256d8eef38533c7

> I was getting an error in the tests in test_authed_hompage.js because my default browser language is Spanish, so the texts don't match.

Ah, interesting! Makes sense. :)

> I started trying to fix it, but at some point I read the test and wondered whether these tests are actually testing any specific js functionality. Are they?

Barely. We do much more Python testing than JavaScript. We have a ticket to adopt a different js testing framework that would hopefully make it easier for us to write js tests: https://github.com/gratipay/gratipay.com/issues/3684. I think I'm okay with dropping the js tests for now, though I'd feel better if we ported them to #3684 instead of dropping them entirely. Opinions, @rohitpaulk @aandis et al.?

whit537:
Re: 975597039374c8015720c2f2c268726bf603f708

> I think this files should be gitignored

Perhaps. On the other hand, they're supposed to be removed by the test suite when it completes successfully. If they're still around it can cause problems—like if the test suite crashes, the files are still around, and then you hit http://localhost:8537/ and wonder why the heck your CSS/JS changes aren't showing up. It's because they're masked by the generated files from the test suite. I think @clone1018 got bit by this once?

If the files are *not* git ignored then it's easier to notice them so you can clean up manually.

whit537:
The rest of your commits look great. :)

deivid-rodriguez:
> Perhaps. On the other hand, they're supposed to be removed by the test suite when it completes successfully. If they're still around it can cause problems—like if the test suite crashes, the files are still around, and then you hit http://localhost:8537/ and wonder why the heck your CSS/JS changes aren't showing up. It's because they're masked by the generated files from the test suite. I think @clone1018 got bit by this once?

> If the files are not git ignored then it's easier to notice them so you can clean up manually.

Makes sense, I didn't know they were supposed to be removed. The test suite was crashing initially for me, because I got bitten [by this](https://github.com/gratipay/gratipay.com/issues/2953). While working on figuring out what was the issue, I noticed those files and thought they were always supposed to be there, so I gitignored them. So I guess an improvement instead would be to get those removed even when the test suite errors. In any case, let me remove the commit! :wink:

whit537:
> So I guess an improvement instead would be to get those removed even when the test suite errors.

:+1:, though to be honest I'm not sure why we have them in the first place. Our dynamic CSS and JS files are behind a CDN in production, and I suspect that pregenerating them at all is an over-optimization (adding complexity, you see ;).

deivid-rodriguez:
Is it [here](https://github.com/gratipay/gratipay.com/commit/9852a478b258ac9b46857d881f7f6ab8ac97bd21) where those optimizations landed? Was there another reason?

whit537:
No, further back ...

whit537:
Here's where the [files get written](https://github.com/gratipay/gratipay.com/blob/1939/gratipay/wireup.py#L268-L290) ...

whit537:
> Is it here where those optimizations landed?

#2838

> removed even when the test suite errors

See #3155 for a previous attempt at this.

whit537:
>  Was there another reason?

Looks like we [didn't really answer](https://github.com/gratipay/gratipay.com/pull/2838#discussion_r19441386) the "why" question. The `compile_assets` function was part of a larger PR.

whit537:
> Don't ignore assets/gratipay.{css,js} after all

https://github.com/gratipay/gratipay.com/commit/8ed11d9a6bcccfbd085cda9f9a808bce98d2bf4d ;-)

whit537:
> why ignore the compiled assets in git ? then you can't see in `git status` that you have stray files

[Confirmed](https://botbot.me/freenode/gratipay/2014-11-03/?msg=24762538&page=3).

deivid-rodriguez:
Thanks @whit537, lots of helful links and comments!

What I'll do is removing 89e037d and then go work on #3684 since it will probably help with most of the other issues mentioned here. Sounds good?

whit537:
> Sounds good?

Sounds good! :-)

whit537:
Thanks for the cleanup, @deivid-rodriguez! :-)

Changaco:
> Looks like we didn't really answer the "why" question. The `compile_assets` function was part of a larger PR.

I don't really remember, but looking back at that PR it seems to me that compiling was necessary. How do you properly cache by hash if you don't have a compiled file to hash?

> Our dynamic CSS and JS files are behind a CDN in production, and I suspect that pregenerating them at all is an over-optimization (adding complexity, you see ;).

CDNs aren't infallible, and neither is the Gratipay app, a bug in either one could result in more requests for assets.

whit537:
> How do you properly cache by hash if you don't have a compiled file to hash?

Yeah, but we don't necessarily need to *store* the compiled version, since we can assume that a recomputed version will be identical (because we trust that neither the inputs nor the computation will have varied). It might take some gymnastics to do this though, depending on how the caching and the hashing are implemented—and we have bigger :fish: to fry. :-)

Changaco:
> since we can assume that a recomputed version will be identical

No, you can't, you can't even check that the inputs haven't changed because you don't know what the inputs are.

I think you underestimate the elegance of the assets compiler, it's less than 40 lines of code, it just works, and for any new `.spt` file under `/assets` too without any configuration required. :-)

Anyway, we both have bigger fish to fry, as you said.

techtonik:
My wish for asset compiler is to be able to compile them as a separate self-sufficient step before running application. This is needed for transition from one CSS compiler to another. It would be nice to have something like Django's `manage.py` in the root that can do:

    gratipay serve
    gratipay compile

whit537:
> I think you underestimate the elegance of the assets compiler

:-)

!m @Changaco 

> My wish for asset compiler is to be able to compile them as a separate self-sufficient step before running application.

That doesn't sound right to me, but we can discuss in another ticket.

PR #3807, opened 2015-10-07
backfill exchanges

whit537:
For #2779. This requires a little schema update: we want to link exchanges to transaction in foreign systems (Balanced, PayPal, etc.). Though with cash I guess we won't have a transaction to link against. Hmm ...

Issue #3822, opened 2015-10-12
avoid null route, ref, and status in new exchanges


whit537:
See #3810 for why cross-linking transactions is a good idea. We were doing this for Balanced by the end, but I don't think we're doing this for _any_ payment networks now.
##

rohitpaulk:
We had the 'reverse' of this in balanced, iirc. There was a meta column named 'exchange_id' on every transaction in Balanced.

whit537:
@rohitpaulk Yes! We preferred that on #3807:

https://github.com/gratipay/gratipay.com/blob/e236680d9daf21afa5e5c078561a3858a37b8c11/backfill-exchanges.py#L124-L129

whit537:
Really we should refactor payment routes along the lines of accounts elsewhere, to have a more consistent interface across all of the routes we support, to include method and transaction linking, and status changes via postbacks.

whit537:
We haven't had a `NULL` `route` (at least) since #3810. See https://github.com/gratipay/gratipay.com/pull/3975#issuecomment-262084837.

whit537:
Folding this into #4442.

Issue #680, opened 2013-02-24
implement an immutable user id


chadwhitacre:
Right now we conflate username and user_id together into participant_id. In order to simplify things like fraud detection (#679) and possibly oauth (#455) we should implement an immutable user_id and rename participant_id to username (#287).

PR #3912, opened 2016-02-04
backfill remaining exchanges

### LINKED ISSUE #2779: backfill the `status`, `route`, and `ref` columns of the `exchanges` table

whit537:
Picks up where #3807 left off. Hopefully we can close #2779, which is blocking https://github.com/gratipay/finances/issues/3.


whit537:
| type of exchange | number | % of total |
|-------|----------:|----------:|
| fine | 49,947 | 62.3 |
| null ref | 30,276 | 37.7 |
| null route | 24,116 | 30.0 |
| null status | 7,930 | 9.9 |
| **total** | 80,233 | 100.0 |

rohitpaulk:
> null status

Ooh, what are we going to do about those? Mark them all as succeeded?

whit537:
@rohitpaulk I guess so? @kaguillera and I were thinking we should approach this month by month alongside https://github.com/gratipay/finances/issues/3, rather than trying to get the backfilling all done before the catching up. For June, 2012 we only have 66 exchanges, and I think we can assume the all succeeded?

whit537:
@kaguillera points out that payday logfiles include username and payment network ("charging foo on Stripe"), which should be helpful here.

whit537:
Script (and `branch.sql`) in 98edf6298fd96067f633ba6f46cb7755ad2c5653. Watch for input logs to accumulate in https://github.com/gratipay/logs/tree/master/misc/3912.

whit537:
@rohitpaulk @aandis et al. If this script looks good then @kaguillera and I can start loading up data with it once we are able to produce the input CSVs.

sql/branch.sql
```diff
@@ -0,0 +1,4 @@
+ALTER TYPE payment_net ADD VALUE 'samurai';
+ALTER TYPE payment_net ADD VALUE 'stripe';
+
+ALTER TABLE exchanges ADD UNIQUE (ref);
```
rohitpaulk: Shouldn't this unique constraint be for a combination of (`network` + `ref`)?

rohitpaulk: Oh wait - `network` is in the exchange routes table. Hmm, how would this work out? Isn't it possible that the `ref` from a `braintree` and that from a `balanced` payment collide?

whit537: Yeah, fair enough. Fixed in 8bec4766e00bd003f40b69f9e4c90788fd470a56.

whit537:
P.S. I'm not expecting this PR to be merged. I'm expecting to use this PR to track code and comments for backfilling the database (#2779).

whit537:
Blech. We don't have `network` in `exchanges`, only in `exchange_routes`. It'll take a bit work to enforce this uniqueness constraint.

whit537:
And I guess we'll have to merge this PR or another one like it, to apply the schema changes.

whit537:
@kaguillera and I are working on a script to match up Stripe transfer details with exchanges in our database. The problem we're up against is ... usernames as primary key (#835). :blush:

whit537:
The way we log username changes, we don't record the old username, only the new one. This makes it impossible to trace back to what username a participant had at a particular point in time. We do have old database snapshots we could look at, but we didn't add user id until late enough that and old snapshot wouldn't help us map old username to current user id anyway.

whit537:
Let this be a lesson, kids! :flushed:

whit537:
Immutable user id implemented in https://github.com/gratipay/gratipay.com/issues/680.

whit537:
Event logging added in #2006. Username change events were backfilled at that point using the username *at that time* but with the original claimed time (+ 0.01 s).

whit537:
Tech debt, anyone? :)

whit537:
:wine_glass: :cherries:

whit537:
But we can still use the old database backups to get the exchange id and username-at-the-time from `exchanges`.

whit537:
Our offline backups only go back to August of 2013.

whit537:
So I guess we manually reconstruct before then?

whit537:
Making progress on this, I'm pretty close to matching all Stripe transactions with exchanges in our database. Now it looks like I have to back up a step because I'm seeing duplicate entries in the Stripe input files.

whit537:
Here's the script I used for https://github.com/gratipay/logs/commit/f0ab1ba9f3fce55b1a7b0b7d4e8e393a8aa89872:

```python
#!/usr/bin/env python2 -u

import os


def process_month(year, month):
    seen = set()
    lines = list()
    filepath = '3912/{}/{}/_stripe-payments.csv'.format(year, month)
    for line in open(filepath, 'r'):
        if line not in seen:
            lines.append(line)
            seen.add(line)
    open(filepath, 'w+').writelines(lines)


def main():
    for year in os.listdir('3912'):
        if not year.isdigit(): continue
        for month in os.listdir('3912/' + year):
            if not month.isdigit(): continue
            process_month(year, month)


if __name__ == '__main__':
    main()
```

And here's the variant I used for https://github.com/gratipay/logs/commit/1ea21621d84b1fb1c8292c2d0fcc420cbcfdc0dd:

```diff
--- find-dupes.py	2016-03-09 13:39:39.000000000 -0500
+++ find-dupes.py.new	2016-03-09 13:39:31.000000000 -0500
@@ -9,9 +9,10 @@
     lines = list()
     filepath = '3912/{}/{}/_stripe-payments.csv'.format(year, month)
     for line in open(filepath, 'r'):
-        if line not in seen:
+        stripped = line.strip()
+        if stripped not in seen:
             lines.append(line)
-            seen.add(line)
+            seen.add(stripped)
     open(filepath, 'w+').writelines(lines)
 ```

**Comment by whit537** [comment]
Okay! The `match-stripe.py` script in b4e9c3ad57253b4f6765681589e99c29edab3eec seems to be working! We're able to find matches for all of the payments that we're finding in CSVs downloaded from Stripe.

That means we're about ready to dial back out and run the [`backfill.py`](https://github.com/gratipay/gratipay.com/blob/b4e9c3ad57253b4f6765681589e99c29edab3eec/backfill.py) script for 2012-07 (our first month with Stripe data).

**Comment by whit537** [comment]
@kaguillera are thinking that we'll backfill exchanges in the database month by month in lockstep with https://github.com/gratipay/finances/issues/3, because we'll be able to learn and adapt more easily as we discover bugs and edge cases and whatnot.

**Comment by whit537** [comment]
We didn't record Stripe failures *at all* in `exchanges`. Let's go ahead and add those in as well. That's not strictly necessary for https://github.com/gratipay/finances/issues/3, but this ticket is broader than that.

**Comment by whit537** [comment]
##### How to Backfill Exchanges

For each month in https://github.com/gratipay/finances/issues/3:

1. generate input CSVs for all vendors that we started using in the given month; for each vendor:
 1. export CSVs for all data for the given—these are our **raw input CSVs**
 1. write one-off scripts to split up raw input CSVs into one CSV per month in the `3912` [directory](https://github.com/gratipay/logs/tree/backfill/misc/3912) in the `logs` repo (`backfill` branch; prefix with `_` to be ignored by `backfill.py`)—these are the **input CSVs**
1. write a `match-{foo}.py` script to generate `3912/YYYY/MM/foo` files (no extension! `{foo}` matches a payment network in the `ENUM` in the database)—these are our **match CSVs**
1. run `backfill.py` to backfill all `exchanges` and `exchange_routes` needed for the given month based on the match CSVs
1. [reconcile the books](https://github.com/gratipay/finances#how-this-repo-is-organized) for the given month

**Comment by whit537** [comment]
Are we assuming that the ~8k exchanges where status is null were successful?

**Comment by whit537** [comment]
September, 2014 is our last `null` status.

August, 2014 is our first `failed` status.

**Comment by whit537** [comment]
FTR our very first few manual payouts are recorded on https://github.com/gratipay/gratipay.com/issues/53.

**Comment by whit537** [comment]
Okay! Stripe is looking pretty good, and `2012-07` is looking pretty good. Now it looks like in `2012-07` we have four Balanced charges that fell through the cracks at https://github.com/gratipay/gratipay.com/pull/3807#issuecomment-146414880, and also our second payout (https://github.com/gratipay/gratipay.com/issues/53#issuecomment-6910203). At least we don't have to do PayPal until ... `2013-03`, it looks like? PyCon 2013! :-)

So the next step here is to go through Step 1 of https://github.com/gratipay/gratipay.com/pull/3912#issuecomment-194485146 for Balanced.

**Comment by whit537** [comment]
I've been doing some work on getting the Balance transactions into the pipeline here. We matched many of these during #3807, but not all of them. We should double-check the ones we matched, and match the rest. Since there's so much data here and it's in a somewhat complex JSON format, it ended up being easier to flatten it into a SQL table (via `sqlite3`) rather than writing to CSV as with Stripe. Below is the `transactions-json2sql.py` script I came up with. The `INSERT`s are getting slower and slower. Hopefully it finishes before eternity. :-) Queries seem to be zippy enough, though, so I'm hoping I won't have to run this script again.

```python
#!/usr/bin/env python3

import os
import json
import sqlite3


cur = sqlite3.connect('transactions.dat').cursor()
cur.execute('DROP TABLE IF EXISTS transactions')
cur.execute('CREATE TABLE transactions (id text, kind text)')


def store(tid, d, prefix=''):
    simple = []
    for key, val in d.items():
        if type(val) is dict:
            store(tid, val, prefix=key+'_')
        else:
            simple.append(key)

    if not simple:
        return

    simple = [prefix + key for key in simple]
    d = {prefix + key: d[key] for key in d}

    # ensure columns are there
    cur.execute("pragma table_info('transactions')");
    known = [col[1] for col in cur.fetchall()]
    for unknown in set(simple) - set(known):
        sqltype = {str: 'text', None.__class__: 'text', int: 'int'}[type(d[unknown])]
        print('adding column: {} {}'.format(unknown, sqltype))
        cur.execute('ALTER TABLE transactions ADD COLUMN "{}" {}'.format(unknown, sqltype))

    # insert data
    SET = ', '.join(['"{}"=?'.format(k) for k in simple])
    vals = tuple([d[k] for k in simple] + [tid])
    UPDATE = 'UPDATE transactions SET {} WHERE id=?'.format(SET)
    cur.execute(UPDATE, vals)
    cur.connection.commit()


for filename in reversed(sorted(os.listdir('transactions'))):
    data = json.load(open(os.path.join('transactions', filename)))
    print(filename, data.keys())
    for key in data:
        if key in ('links', 'meta'): continue
        for transaction in data[key]:
            tid = transaction.pop('id')
            del transaction['links']
            cur.execute('INSERT INTO transactions (id, kind) VALUES (?, ?)', (tid, key))
            store(tid, transaction)
```

whit537:
About 92% of the way through the JSON files:

```
sqlite> select count(*) from transactions;
109835
sqlite> pragma table_info('transactions');
0|id|text|0||0
1|kind|text|0||0
2|meta_state|text|0||0
3|meta_participant_id|text|0||0
4|transaction_number|text|0||0
5|status|text|0||0
6|created_at|text|0||0
7|updated_at|text|0||0
8|failure_reason_code|text|0||0
9|currency|text|0||0
10|voided_at|text|0||0
11|href|text|0||0
12|amount|int|0||0
13|description|text|0||0
14|expires_at|text|0||0
15|failure_reason|text|0||0
16|meta_exchange_id|text|0||0
17|appears_on_statement_as|text|0||0
18|meta_balanced.result.trace_number|text|0||0
19|meta_balanced.result.return_reason_code|text|0||0
sqlite>
```

whit537:
```sql
sqlite> select count(*) from transactions;
110867
sqlite> select count(*) from transactions where meta_participant_id is null;
66275
sqlite>
```

whit537:
All done.

```sql
sqlite> select count(*) from transactions;
119477
```

It's a 27 MB file, so I don't think I'm going to commit it to the logs repo.

whit537:
> I'm hoping I won't have to run this script again.

Famous last words. I need the unique customer ID, which I think is in the `links`, which I filtered out.

whit537:
The main reason to use SQLite instead of CSV is that the JSON files (there are 1197 of them) contain duplicates, because I exported them from Balanced at least twice, maybe three times. The files are in batches of 100 transactions each, and the dates aren't sequential across files in filename order. Using SQLite makes it easier to collapse down to a single consistent ordering.

Perhaps one SQLite file per month would be easier to write to?

whit537:
I'm chipping away on the `match-balanced.py` script, usually on the bus because it's good offline work. I did end up going with one file per month, and then realized it was easy to go from SQLite to CSV, so we've now got `_balanced.csv` files for each relevant month in [logs/misc/3912](https://github.com/gratipay/logs/tree/backfill/misc/3912). My next goal is to write a `match-balanced.py` that can cleanly match all of the transactions in the Balanced export data with records in our database. We've got data from July, 2012 through October, 2015, and my next step is to get the script to run over the whole of the data without throwing an exception—at least saying "I give up" when it can't find a match. Then it'll be a matter of going through the failures to match, and updating the script (and potentially modifying the database) to find matches.

whit537:
`match-balanced.py` in 4dcbb5ad18c7617b95cd18a82dca6b913b76a751 makes it all the way up to April, 2015 before crapping out.

whit537:
Rebased on master at a756d238de49380ea42f352af2ffc705a8184347. Previous head was 4dcbb5ad18c7617b95cd18a82dca6b913b76a751.

whit537:
```
[gratipay] $ ./tally-backfill.py 
IMPOSSIBLE!!!!!!!!!!!         3119
all set!                     41286
yes                           4752
no                           11145
missing ref and status!          0
mismatched ref and status!       0
missing ref!                   702
mismatched ref!                  0
missing status!                  0
mismatched status!               0
                             61004
[gratipay] $
```

whit537:
b88f1daa27a7d09ab87df152724cec449fb48ad2 is 10x faster! :-)

```
[gratipay] $ time run defaults.env local.env ./match-balanced.py > backfill.two
  File "./match-balanced.py", line 321, in <module>
    main(_matcher, _constraint)
  File "./match-balanced.py", line 314, in main
    process_month(matcher, cid2mat, uid2cid, year, month)
  File "./match-balanced.py", line 174, in process_month
    raise Exception(rec)
Exception: {'links__dispute': '', 'updated_at': '2015-10-10T04:33:22.299846Z', 'links__destination': '', 'currency': 'USD', 'href': '/refunds/RF2MQLF8LISwKAkFO6nC2PZ5', 'meta__exchange_id': '', 'failure_reason': '', 'id': 'RF2MQLF8LISwKAkFO6nC2PZ5', 'links__card': '', 'transaction_number': 'RFKTG-N8E-5DJV', 'failure_reason_code': '', 'links__debit': 'WD6tFwILinRNmeydI6T0NNFw', 'status': 'pending', 'meta__balanced.result.return_reason_code': '', 'description': 'RobertKohut', 'expires_at': '', 'links__credit': '', 'appears_on_statement_as': '', 'links__customer': '', 'links__card_hold': '', 'kind': 'refund', 'meta__state': '', 'created_at': '2015-10-10T04:33:20.547708Z', 'links__settlement': '', 'links__order': '', 'meta__balanced.result.trace_number': '', 'amount': u'5.26', 'meta__participant_id': '', 'voided_at': '', 'links__source': ''}

real    7m9.931s
user    6m51.047s
sys     0m4.654s
[gratipay] $
```

whit537:
I went back and fixed a bug in the JSON to CSV conversion in https://github.com/gratipay/logs/commit/3f78119ae8734a3d8941dd6de07ed981401911e9. Now I intend to go back and re-run the pre-optimized version of `match-balanced.py` because what I'm trying to do is get the same output from the pre-optimized version as the optimized version. Right now they're different—and now they're extra-different with the underlying bug fix.

```
[gratipay] $ ./tally-backfill.py backfill.one
IMPOSSIBLE!!!!!!!!!!!$        3119
all set!$                    41286
yes$                          4752
no$                          11145
missing ref and status!$         0
mismatched ref and status!$      0
missing ref!$                  702
mismatched ref!$                 0
missing status!$                 0
mismatched status!$              0
full of grace                    0
                             61004
[gratipay] $ ./tally-backfill.py backfill.two 
IMPOSSIBLE!!!!!!!!!!!$        3589
all set!$                    40701
yes$                          4752
no$                          11594
missing ref and status!$         0
mismatched ref and status!$      0
missing ref!$                  539
mismatched ref!$                 0
missing status!$                 0
mismatched status!$              0
full of grace                    0
                             61175
[gratipay] $
```

whit537:
I've added sorting to the pre-optimized version so that we can accurately compare its output with the optimized version. I've confirmed that the sorting doesn't change any of the actual assignments.

whit537:
There's definitely still significant differences between the unoptimized and optimized versions of `match-balanced.py`. My hunch is that the optimized version is correct and the unoptimized version is improperly matching *in the past* as well as in the future. More research required ...

whit537:
We're looking at about 10 minutes for one pass through with the basic algorithm in 5c21877bbc3f08ce7285df601affcf4f950bdc10.

```
[gratipay] $ time run defaults.env local.env ./match-balanced.py
We have 57519 transactions to match!
We have 74577 exchanges to match!
..........................................................We found 0 matches!

real    9m34.865s
user    9m30.379s
sys     0m2.083s
[gratipay] $
```

whit537:
We started recording failures in https://github.com/gratipay/gratipay.com/pull/3011, which went into effect in December, 2014.

whit537:
Okay! @kaguillera and I spent yesterday and today rewriting `match-balanced.py`. We're hopeful! 🙏 

```
[gratipay] $ run defaults.env local.env ./match-balanced.py
Loading transactions ... we have 69901 transactions to match!
Loading exchanges ... we have 74577 exchanges to match!
  411 / 60414 |   524 / 64499 | 7936 matches | 13.6% | T-12.5 m              
```

Here's how the script works now:

 1. Load transactions from Balanced and exchanges from Gratipay.
 1. Build mappings of `cid` (Balanced customer id) to transactions, and `uid` (Gratipay user id) to exchanges.
 1. Loop through transactions. Whenever we are able to positively match a transaction with an exchange (based on amount, username, and ~10 second window), drop to an inner loop over all transactions and exchanges for the given (`cid`, `uid`):
  1. Remove all of the local transactions and exchanges from the global lists.
  1. Put them *either* in a global `matched` list *or* in a global `unmatchable` list.
 1. Write matches out to [match CSVs](https://github.com/gratipay/gratipay.com/pull/3912#issuecomment-194485146).
 1. Write unmatchables out to their own files as well.

##### Next Steps

 - [ ] Make sure we are getting accurate numbers for matches, unmatchable transactions, and unmatchable exchanges.
 - [ ] Account for all unmatchable transactions and exchanges.
 - [ ] Perform some analysis and checking of matches to make sure they are right.

```
[gratipay] $ run defaults.env local.env ./match-balanced.py
Loading transactions ... we have 69901 transactions to match!
Loading exchanges ... we have 74577 exchanges to match!
 3213 /  3214 |  1985 /  2462 | 51134 matches | 95.4% | T-28 s                
We found 51134 matches!
[gratipay] $
```

whit537:
More notes to self:

 - Expand ~10 second look-ahead (weigh the effect on runtime with any improvement in match rate).
 - Really pin down the ways that transactions and exchanges can relate.
  - Where do we need to create exchanges that don't exist? Failures, refunds, and reversals, ya?
  - We should account for every single transaction, right?
 - Possible explanations for unmatchable transactions:
  - We didn't track failures until December, 2014
  - We *never* tracked refunds (right?) ... or reversals?
  - Why did we see successful debits in unmatchable? Does the look-ahead window explain that?
 - Possible explanations for unmatchable exchanges:
  - Exchanges include non-Balanced exchanges (Braintree, Coinbase, etc.).
  - Network failures could result in an exchange with no transaction (right?).
 - We will need to output match CSVs by month.
  - Right now all the output data is in one file.
  - It's not sorted by date.
  - Also need to make sure we have `cid` (for failed card holds, it seems).

whit537:
With da46305:

```
[gratipay] $ run defaults.env local.env ./match-balanced.py
Loading transactions ... we have 69901 transactions to match!
Loading exchanges ... we have 74577 exchanges to match!
 3213 /  3214 |  1985 /  2462 | 51134 matches | 95.4% | T-30 s                
We found 51134 matches!
[gratipay] $ wc -l unmatchable.*
     915 unmatchable.ambiguous_card_hold
   15553 unmatchable.dregs
     319 unmatchable.early_failures
   20981 unmatchable.exchanges
     489 unmatchable.non_card_hold_without_cid
   38257 total
```

whit537:
In dregs and exchanges I'm seeing a lot of failures that I would expect to match. Something to look at next ...

kaguillera:
dregs??

whit537:
@kaguillera "Dregs" is the [left-over transactions](https://github.com/gratipay/gratipay.com/blob/867b8fcac9c1b381852d158b064f87cf74923523/match-balanced.py#L269). All of the other categories of failure are also transactions, so I used "dregs" to differentiate from other transaction failures as well as exchanges.

kaguillera:
I remember..for some reason I thought it was an abbreviation for something else

whit537:
http://www.dictionary.com/browse/dregs :-)

kaguillera:
👅

whit537:
With 530fa50fd5e3355284e277f2ac94ceb16276a465:

```
[gratipay] $ time run defaults.env local.env ./match-balanced.py
Loading transactions ... we have 69901 transactions to match!
Loading exchanges ... we have 74577 exchanges to match!
 3213 /  3214 |  1985 /  2462 | 66637 matches | 95.4% | T-30 s
We found 66637 matches!

real    10m49.635s
user    10m34.695s
sys     0m5.490s
[gratipay] $
```

whit537:
We're up from 73.2% to 95.3% matches!

kaguillera:
great what did you do extend the look-ahead time?

whit537:
I actually let it go 60 seconds into the past as well, but just for `failed`. We may want to extend that to `succeeded` as well.

whit537:
🐰 :hole: Can't bring myself to close yet.

whit537:
@kaguillera discovered that we have good date for this in our payday logs! Can we make any quicker progress based on that?

whit537:
Over at https://github.com/gratipay/gratipay.com/pull/4342#issuecomment-283450229 we are seeing signs of some corruption: `braintree-cc` network for some routes used waaaaay before we had Braintree.

whit537:
Rebased, was 43804a36e8994182d2b0d4e5f96d6a90b6af9901.

whit537:
Gosh. Bring @dmk246 up to speed here, and it's getting worse. We're seeing a route for yours truly that is of `unknown` network, and has 219 exchanges on it from 2012-06-01 through 2016-11-17. Amount and notes and circumstances indicate that it in fact represents a number of different networks: Samurai, PayPal, Braintree, Balanced, and Coinbase at the least. 😳

whit537:
Alright, this is A Big Deal&trade;. It's by far our biggest piece of technical debt. This underpins our whole accounting function, so we *have* to get this done. And it's going to be a _lot_ of work. It sounds like @kaguillera is going to dive back into this one, along with @dmk246 as she gets up to speed.

As @dmk246 pointed out IRL, the immediate next task is to ensure that we're not adding new bad data ... which we are: we are not storing `ref` for *any* new transactions—`record_exchange` [doesn't even allow](https://github.com/gratipay/gratipay.com/blob/2053/gratipay/billing/exchanges.py#L231-L262) for it. 😮

!m @kaguillera @dmk246

kaguillera:
Diving back into this. 

> As @dmk246 pointed out IRL, the immediate next task is to ensure that we're not adding new bad data ... which we are: we are not storing ref for any new transactions—record_exchange doesn't even allow for it. 😮

This is taken care of by https://github.com/gratipay/gratipay.com/pull/4361

Now back to figuring out the corrupted data problem. @whit537 I think I will need a snapshot of the tables that are involved...exchanges, routes and exchange_route if that is at all possible.

kaguillera:
@whit537 can I have a the actual records from exchanges and exchange_routes for the month of 2012-06 to begin with. I guess a csv would be good or any format that would be easiest and fastest for you. I don't want to take you away from it what you are working on. Maybe you could drop it some where in logs repo. 
I want to analyze it along with the payday logs and see if I can figure out the source and reason for the data corruption. Then maybe we can come up with a solution.

whit537:
I exported the `exchange_routes` table, id and username from `participants`, and the first three months of `exchanges`.

kaguillera:
tanx...going to start looking at this

kaguillera:
This ticket is getting too large as it pertains to the comments so I am moving the research of the errors to https://github.com/gratipay/gratipay.com/issues/4442

whit537:
Bah! You can never have too many comments! :-)

whit537:
We added `unknown` in https://github.com/gratipay/gratipay.com/pull/3975.

whit537:
We haven't had an unknown network for six months and I'm trying to remember why.

```
gratipay::DATABASE=> select timestamp, network from exchanges e join exchange_routes er on e.route = er.id where network='unknown' order by timestamp desc limit 10;
┌───────────────────────────────┬─────────┐
│           timestamp           │ network │
├───────────────────────────────┼─────────┤
│ 2016-11-17 21:33:03.859138+00 │ unknown │
│ 2016-11-17 21:33:02.467701+00 │ unknown │
│ 2016-11-17 21:33:01.429801+00 │ unknown │
│ 2016-11-17 21:33:00.15087+00  │ unknown │
│ 2016-11-17 21:32:59.618481+00 │ unknown │
│ 2016-11-17 21:32:59.212181+00 │ unknown │
│ 2016-11-17 21:32:58.823265+00 │ unknown │
│ 2016-11-17 21:32:58.47406+00  │ unknown │
│ 2016-11-17 21:32:56.693921+00 │ unknown │
│ 2016-11-17 21:32:56.283582+00 │ unknown │
└───────────────────────────────┴─────────┘
(10 rows)

gratipay::DATABASE=>
```

whit537:
We plugged the `ref` hole in https://github.com/gratipay/gratipay.com/pull/4361. What about the `network` hole?

whit537:
The last unknown networks are for PayPal masspays. Those run through record an exchange which depends on existing routes. How do we have a route with a null network and why did it stop seven months ago?

Last change to routes/associate.json.spt is four months ago, then before that is two years.

https://github.com/gratipay/gratipay.com/blame/24559cb0ee6d9fff77f64fcdba914a19fc0a4b98/www/~/%25username/routes/associate.json.spt

whit537:
```
gratipay::DATABASE=> select count(*) from exchange_routes where network='unknown';
┌───────┐
│ count │
├───────┤
│  1626 │
└───────┘
(1 row)
```

whit537:
We didn't fully plug the `ref` leak in https://github.com/gratipay/gratipay.com/pull/4361. We plugged it for  `braintree-cc` but not for `paypal`, i.e., MassPay.

```
gratipay::DATABASE=> select count(*) from exchanges where ref is null and timestamp::text > '2017-03';
┌───────┐
│ count │
├───────┤
│  1089 │
└───────┘
(1 row)

gratipay::DATABASE=>
```

whit537:
Looking at those 1,089 it is clear from the `note` that the only thing still leaking NULL `ref`s since https://github.com/gratipay/gratipay.com/pull/4361 is MassPay. A PR! A PR!

whit537:
We have four things we need to backfill, two are now using `unknown` (`status` and `network`) and two are still using `NULL` (`route` and `ref`)

In general we want to plug leaks before backfilling. Moving back over to https://github.com/gratipay/gratipay.com/issues/3822 ...

kaguillera:
Once #4518 is merged [this](https://github.com/gratipay/gratipay.com/pull/3912#issuecomment-305856368) should be taken care of.

whit537:
Closing per https://github.com/gratipay/inside.gratipay.com/issues/1196.

Issue #3545, opened 2015-06-11
make Teams editable


whit537:
Reviewing teams is creating pressure to edit teams, as one might expect. :)

#### Notify
- [x] https://gratipay.freshdesk.com/helpdesk/tickets/2879
- [x] https://gratipay.freshdesk.com/helpdesk/tickets/3247

whit537:
+1 from @stevepiercy at https://github.com/gratipay/gratipay.com/issues/3681#issuecomment-135148451:

> when I log in as ~stevepiercy and view my account profile, I expect to see a UI for me to manage Team KnopProject, but cannot find one. I couldn't manage the Team, so I assume one of the Gratipay staff assigned me as Owner on my behalf. Is some public Team management UI is forthcoming?

book:
How will the team owner/admins invite new members to the team? I expect this to be part of "team management" too.

whit537:
@book Yeah, on Gratipay 1.0 there was a page listing all of the members of the team, and team admins had an input on there to locate and add new members. We'll bring this back below #3433.

whit537:
+1 @sils1297 at https://github.com/gratipay/team-review/issues/36#issuecomment-137971240.

whit537:
+1 @duythien in https://github.com/gratipay/team-review/issues/84:

> My team https://gratipay.com/phanbook/ have just active and I want to change logo, but I'm does not see options. Can you help me change logo

whit537:
+1 from @Jaex at https://github.com/gratipay/team-review/issues/79#issuecomment-148617054.

whit537:
+1 from @webschik at https://github.com/gratipay/team-review/issues/83#issuecomment-148604227.

mattbk:
+1 from me

whit537:
+1 @senthilporunan https://github.com/gratipay/team-review/issues/89#issuecomment-150468492

whit537:
How does Team editing interact with [review](http://inside.gratipay.com/howto/review-teams)? If Team pages are changing and we're making yes/no decisions based upon the content thereof, then should we freeze Team pages at some point? Or at least show the history of changes (at least for admins?)?

whit537:
Should changes once a Team has been approved have to go through another approval process?

whit537:
Or should we simply log those as `events` and provide a dashboard to passively review Team change events?

kaguillera:
> simply log those as events and provide a dashboard to passively review Team change events
+1 
[...]

PR #3923, opened 2016-02-20
Edit teams.

aandis:
References #3545

wip.


aandis:
Alright! It's now possible to edit teams. :dancer: 
There are a few ui improvements I have in mind which I'll do once this passes review. Will also add tests if this looks good.

rohitpaulk:
Yay! :beers: Screenshots please?

aandis:
screenshot from 2016-02-22 22 52 23

aandis:
screenshot from 2016-02-22 22 54 06

aandis:
screenshot from 2016-02-22 22 57 02

mattbk:
Does this allow admins to edit teams?

gratipay/models/team.py
```diff
@@ -67,6 +67,26 @@ def insert(cls, owner, **fields):
...
+      placeholders = ', '.join(['%s']*len(vals))
+
+      return self.db.one("""
+
+        UPDATE teams
+           SET ({0}) = ({1})
+         WHERE id = %s
+     RETURNING teams.*::teams
+
+        """.format(cols, placeholders), vals+(self.id,))
```
rohitpaulk: Let's also write to the `events` table with details about the changes - we need admins to be able to view an 'audit log'. I'm digging up the ticket where this was discussed..

aandis: https://github.com/gratipay/gratipay.com/issues/3545#issuecomment-152497341?

rohitpaulk: yess, that one.

gratipay/models/team.py
```diff
@@ -67,6 +67,26 @@ def insert(cls, owner, **fields):
         """, fields)
 
 
+    def update(self, **kw):
+      updateable = frozenset(['name', 'product_or_service', 'homepage',
+                              'onboarding_url', 'todo_url'])
+
+      cols, vals = zip(*kw.items())
+      assert set(cols).issubset(updateable)
```
rohitpaulk: This would return a `500` if someone tries to force another column in there. Let's throw a `400` instead?

aandis: It won't receive extra columns, at least not from the web endpoint. https://github.com/gratipay/gratipay.com/pull/3923/files#diff-8cd27f320d80ef3f1fed6dc52f5feafdR40 makes sure of that. This check is just in case this is called from somewhere else in future in which case they should catch the exception.

gratipay/models/team.py
aandis: will do.

mattbk:
> I was under the impression admins already had a friendly way of doing so. I can add it if it's needed. Would be minor changes.

No, it's been happening through the database, so those of us without db access haven't been able to edit Teams. This PR solves us needing to do that for trivial things, however I think it's probably best to let admins edit anything a normal user can edit (someone can back me up or disagree on that).

aandis:
Done @mattbk :)
> Thanks for doing this

Just doing my share in making gratipay better. :)

aandis:
todos - (mostly after review)

- [ ] write changes to `events` tables.
- [ ] add [rejected team check](https://github.com/gratipay/gratipay.com/pull/3923/files#diff-8cd27f320d80ef3f1fed6dc52f5feafdR30).
- [ ] disable modify btn on ui until changes made in the form.
- [ ] on successful edit, disable modify btn and redirect to team homepage.
- [ ] add tests.

whit537:
!m @aandis @mattbk @rohitpaulk 

I love watching this happen! It's so encouraging to not be a bottleneck! :-)

whit537:
One note I'll throw out there: let's be sure to log the *old* value in `event` and not just the new one. @kaguillera and I learned about this the hard way under https://github.com/gratipay/gratipay.com/pull/3912, where we were trying to reconstruct old usernames and weren't able to because we were (and are still!) logging the *new* value and throwing away the old.

mattbk:
Editing team text fields seems to work for me (as admin).

Might be nice to have either a breadcrumb/menu for going back, or a cancel button if the form is automatically opened.

When I try to upload an image, I get this response:
`{
    "error_code": 500,
    "error_message_long": "Traceback (most recent call last):\n  File \"/home/matt/Documents/github/gratipay.com/env/lib/python2.7/site-packages/algorithm.py\", line 288, in run\n    new_state = function(**deps.as_kwargs)\n  File \"/home/matt/Documents/github/gratipay.com/env/lib/python2.7/site-packages/aspen/algorithms/website.py\", line 115, in get_response_for_resource\n    return {'response': resource.respond(state)}\n  File \"/home/matt/Documents/github/gratipay.com/env/lib/python2.7/site-packages/aspen/http/resource.py\", line 59, in respond\n    content_type, body = super(Dynamic, self).respond(accept, state)\n  File \"/home/matt/Documents/github/gratipay.com/env/lib/python2.7/site-packages/aspen/simplates/__init__.py\", line 169, in respond\n    exec(self.pages[1], context)\n  File \"/home/matt/Documents/github/gratipay.com/www/%team/edit/edit.json.spt\", line 42, in <module>\n    image = data[field].value  # file upload\nAttributeError: 'unicode' object has no attribute 'value'\n",
    "error_message_short": "Internal Server Error"
}`

aandis:
Interesting. I can't reproduce the error with any image types @mattbk. Can you paste the image you were trying to upload here?

mattbk:
It was the first image that came up for "cat" on Google: https://www.petfinder.com/wp-content/uploads/2012/11/140272627-grooming-needs-senior-cat-632x475.jpg

It could be a local issue with my virtual Ubuntu setup. If nobody else is having problems, we can probably carry on.

aandis:
Man, them cats I tell you. :pouting_cat: Works fine for me btw. 

screenshot from 2016-02-24 20 36 30

aandis:
> If nobody else is having problems, we can probably carry on.

Can somebody else please confirm?

mattbk:
@rohitpaulk @whit537 Could one of you check the image upload?

I think I'm not building things correctly on either machine :-\ Just for the record, now I get the same error whether or not I try to change the image.

aandis:
@mattbk can you apply for a new team locally from the home page and check whether you get the error? The image handling section is pretty much the same in both cases.

rohitpaulk:
> Could one of you check the image upload?

Testing..

rohitpaulk:
I get the same error... 

screen shot 2016-02-25 at 9 15 41 pm

rohitpaulk:
I'm noticing a few other errors too - On the homepage, 

screen shot 2016-02-25 at 9 17 30 pm

^ Clicking 'continue' does nothing, and I see an error in my console

screen shot 2016-02-25 at 9 17 24 pm

rohitpaulk:
> I get the same error...

I'm curious to know how it is passing for you, @aandis :smile:

rohitpaulk:
I fixed https://github.com/gratipay/gratipay.com/pull/3923#issuecomment-188848749 in https://github.com/gratipay/gratipay.com/commit/3c191e1bcc03172b9b060697befaf9825e56cf35

mattbk:
> I'm noticing a few other errors too - On the homepage,

I have that too, thought it was a local bug.

aandis:
Hmm. This is weird. I don't see any error messages in the console and team edits work flawlessly.
screenshot from 2016-02-26 00 00 28


However, if I delete `team.js` I can see the console error messages and can also reproduce @mattbk's problem. It seems `team.js` is non existent on your platforms but picked up by my browser alright. Perhaps the order in which the two scripts are loaded are different on our platforms hence `Gratipay.team` ends up undefined for you guys?

aandis:
I can make it work by taking the edit form out of `Gratipay.team` into it's own independent object. But I am really curious atm why this might be happening.

rohitpaulk:
> Perhaps the order in which the two scripts are loaded are different on our platforms hence Gratipay.team ends up undefined for you guys?

Yes, that is the issue. The order of files in https://github.com/gratipay/gratipay.com/blob/master/www/assets/gratipay.js.spt#L5 is indeterminate

gratipay/models/team.py
rohitpaulk: :+1:, makes sense

aandis:
:scream: guess Gratipay still has some tricks up it's sleeve for me. :)

aandis:
alright fixed. This should work now @rohitpaulk @mattbk .

rohitpaulk:
What was the root cause behind https://github.com/gratipay/gratipay.com/pull/3923#issuecomment-188848235?

rohitpaulk:
Fix for https://github.com/gratipay/gratipay.com/pull/3923#issuecomment-188848749 confirmed.

rohitpaulk:
Fix for https://github.com/gratipay/gratipay.com/pull/3923#issuecomment-188848235 confirmed too, nice!

rohitpaulk:
screen shot 2016-02-26 at 12 37 16 am

^ The UI seems to suggest that the maximum size is 256kb, but the error notif says it is 100kb

screen shot 2016-02-26 at 12 39 03 am

Who should I believe? :scream:

aandis:
> What was the root cause behind #3923 (comment)?

Good question! And the explanation is real sweet. So what was happening is, since the js wasn't even loaded for the form(because of `Gratipay.js` not being defined), the [form](https://github.com/gratipay/gratipay.com/pull/3923/files#diff-83a05098ee02ba7143553707e513875fR43) was directly being sent to the endpoint when `submit` was clicked. Since `form` doesn't specify `enctype= multipart/form-data`, the file wasn't being sent. Only the text fields inputs were sent. Hence the `image` didn't exist!

mattbk:
Success for me! (on team edit, haven't tried /new)

Still would like a button/link or auto-return to the team page, though.

aandis:
> Who should I believe? :scream:

https://github.com/gratipay/gratipay.com/blob/master/www/teams/create.json.spt#L83 
Believe the code. 
ALWAYS BELIEVE THE CODE! :smile:

aandis:
> Success for me!

Yay. :)

> haven't tried /new

No need to I guess. It'll work now. The problem was with the order in which js files were picked on different platforms which shouldn't matter now. See https://github.com/gratipay/gratipay.com/pull/3923#issuecomment-188923404

> Still would like a button/link or auto-return to the team page, though.

Alright. noted. Do you have a place in mind where we can position it so it looks good?

mattbk:
I think it would fit the UX better if you return to the team page when you click "modify." That's how it works for editing a ~user profile. Also could use a "cancel" button next to "modify" so a person can back out of the form.

If I knew how to do it, I would jump in :-\

aandis:
> I think it would fit the UX better if you return to the team page when you click "modify."

That's already in my [todos](https://github.com/gratipay/gratipay.com/pull/3923#issuecomment-187315707). Working on it now. :)

mattbk:
!m @aandis.

aandis:
:+1:

mattbk:
Nice. All works for me.

rohitpaulk:
> And the explanation is real sweet. So what was happening is, since the js wasn't even loaded for the form(because of Gratipay.team not being defined), the form was directly being sent to the endpoint when submit was clicked. Since the form doesn't specify enctype = multipart/form-data, the file wasn't being sent. Only the text fields inputs were sent. Hence the image didn't exist!

Aha, I wonder if there is a way to make it work even without JS though. It might not be possible in this case, but worth looking into. If we can make it such that submitting the form before the JS kicks in gives us a slightly worse user experience, but the same functionality (no weird errors) that'd be cool.

js/gratipay/edit_team.js
```diff
@@ -0,0 +1,51 @@
...
+    var data = new FormData();
+    $form.find(":input").each(function() {
+        if($(this).data("original") !== $(this).val()) {
+            modified = true;
+
+            var name = $(this).attr("name");
+            if (name === "image") {
+                data.append(name, $(this)[0].files[0]);
+            } else {
+                data.append(name, $(this).val());
```
rohitpaulk: Why are we looking for a diff here, @aandis? Can't we just use the native submit the plain form with all values (like we do in `new_team.js` https://github.com/gratipay/gratipay.com/blob/master/js/gratipay/new_team.js#L14?

aandis: Why the unnecessary overhead if we are only interested in attributes that changed? Only passing necessary attributes also fits well with the [update](https://github.com/gratipay/gratipay.com/pull/3923/files#diff-aa4aac98a6ec2173f86c51a728870f09R70) method.

aandis: Also, this is supposed to be a `PATCH` not a `PUT` request. The former requires only the attributes being updated to be passed in the request body. I wanted to [use](https://github.com/gratipay/gratipay.com/pull/3923/files#diff-8cd27f320d80ef3f1fed6dc52f5feafdR11) `patch` here, but for some reason it breaks everything. :( The idea here is that if later on we decide to expose this api to public officially, you shouldn't have to do a `GET` (to get all the attributes)  before hitting this endpoint.

rohitpaulk: > The idea here is that if later on we decide to expose this api to public officially, you shouldn't have to do a GET (to get all the attributes and their values) before hitting this endpoint.

I agree.

My question though, is why we're doing it in our internal usage. I'm concerned about this because a JS bug in the logic that figures out which fields have changed would be very hard to track down. Wouldn't this functionality work exactly the same if we weren't generating a diff (just send out all the values) before submitting the form here?

aandis: >  I'm concerned about this because a JS bug in the logic that figures out which fields have changed would be very hard to track down. 

Not sure what you mean. What *kind* of bug? The logic is pretty simple, store the original values and when the form is submitted, compare them with the new values.

rohitpaulk: > when the form is submitted

That's what I'm concerned about. We're relying on the JS logic to post the relevant fields. If we got rid of that and just blindly submitted all the fields instead, would it make a difference to the functionality? 

> What kind of bug?

Where a user edits a field, and our JS logic fails to recognize that the field has been edited - and hence doesn't send the field in the POST request.

aandis: > our JS logic fails to recognize that the field has been edited

Why would it, if it's written to do exactly that? Can you give an example where that happens?

I think you are concerned about https://github.com/gratipay/gratipay.com/pull/3923#issuecomment-189099923 where the form is submitted before the js even loads, which *may* be possible. And to fix that, a simple `enctype = multipart/form-data` on the form is enough. (both on new team and edit team forms). But I still think it's better to send only necessary data if we can.

rohitpaulk: > Why would it, if it's written to do exactly that?

It is know, what happens when sometime in the future someone makes a small edit? :D This is a minor concern - the simpler our code is, the more maintainable it is, and the less bugs we'll have over time :)

rohitpaulk: > And to fix that, a simple enctype = multipart/form-data on the form is enough. (both on new team and edit team forms).

Please add that too

rohitpaulk: > It is now, what happens when sometime in the future someone makes a small edit?

One other reason that I'm pushing this is that our JS tests aren't comprehensive - a bug in our JS isn't going to be covered by tests.

aandis: Cool. I'll get rid of this. :)

aandis: Done in 464d9a117194a2c7fa02c39523dae9553e9b199f

aandis:
Alright. Tests added and rebased on master. :dancer: This is :white_check_mark: from my side. :)

mattbk:
Now I get "Bad Request" when I submit the edit form.

image

mattbk:
I guess I will go through commits to check.
d9a5292 works.

aandis:
oooooooooh. There's a `csrf_token` which gets sent. The tests don't replicate that. :/
Should be fixed now @mattbk

mattbk:
Works! 

The workflow to remove an image is a little weird to me (choose file->cancel), but that can probably be a different PR since most people will be replacing the image, not removing it.

mattbk:
@aandis can you tell me how to run just the tests associated with this PR? I'm relatively new at running tests.

aandis:
> Works!

Ha! I just recently started writing tests after @rohitpaulk's continuous advice and started thinking of them as a source of truth. Lesson learnt => Never trust tests 100%. Especially your own tests. :wink:

> The workflow to remove an image is a little weird to me

You can't really remove a team image since you always need to have one so you'll always be replacing it. If you mean `removing` it from the `form` then maybe yeah.

rohitpaulk:
> Never trust tests 100%. Especially your own tests.

@aandis - There are tests that you *can* trust 100% :) Full-blown integration tests, which actually open a browser and click through, for example.

mattbk:
> You can't really remove a team image since you always need to have one so you'll always be replacing it.

You're right; I didn't click Modify to see that the image doesn't disappear.

aandis:
> There are tests that you can trust 100% :) Full-blown integration tests, which actually open a browser and click through, for example.

Cool! I'll look into the js tests and selenium someday as well. :)

mattbk:
- @aandis's checklist is complete.
- looks like @rohitpaulk's comments have been answered (or reticketed to https://github.com/gratipay/gratipay.com/issues/3930)
- works on my Mac
- works on my virtual Ubuntu
- passes Travis.

Merging.

aandis:
> Merging.

Yay! :)

> Confirmed this works (from above): "One note I'll throw out there: let's be sure to log the old value in events and not just the new one."

One thing I want to point out here before this is deployed is that I decided to *only* log the old *changed* values since the most recent values are already present in the `teams` table, so there's no point logging the new ones. This should be enough to generate the series of edits done for a team. I hope @rohitpaulk  reviewed this already. I don't fully understand the use case of this yet, so if logging the entire tuple of `team` which is edited is more useful to the use case, I can change this.

Change is [here](https://github.com/gratipay/gratipay.com/pull/3923/files#diff-aa4aac98a6ec2173f86c51a728870f09R91) and the tests for that [here](https://github.com/gratipay/gratipay.com/pull/3923/files#diff-6111ddd8c8732521886de4e8a7d13a55R466)

whit537:
>  I decided to *only* log the old *changed* values

:+1: from me.

Thanks for this @aandis, awesome to see forward progress! :dancer:

aandis:
:dancer::dancer:

PR #3931, opened 2016-03-02
Remove Balanced callback

whit537:
This was our *only* callback.


whit537:
I noticed [a Sentry error](https://app.getsentry.com/gratipay/gratipay-com/issues/114288948/) where it appeared that a security researcher was exploring this endpoint. I haven't noticed a vulnerability surface in HackerOne yet but this endpoint has always [given](https://github.com/gratipay/gratipay.com/pull/2620#discussion_r16124314) me the [willies](https://github.com/gratipay/gratipay.com/pull/2620#discussion_r16128266) (publicizing the IPs that we're filtering against!?), and Balanced is dead so this should be removed regardless.

whit537:
Ready for review, @aandis @rohitpaulk et al.

Issue #3932, opened 2016-03-02
No "Edit your team" link


stevepiercy:
Team KnopProject does not have a link to "Edit your team" as shown in #3923. I have tried logged in as either the user ~KnopProject and ~stevepiercy. ~stevepiercy is the owner of the team KnopProject.

aandis:
@stevepiercy #3923 is not live yet. You'll see the link once it's deployed. :)

stevepiercy:
OK, I thought I had missed something. Thanks.

PR #3934, opened 2016-03-03
Team creation fix. Fixes #3929

### LINKED ISSUE #3929: Can't submit new team after changing image.
Can't believe this didn't come up yet. I noticed this while exploring [create.json.spt](https://github.com/gratipay/gratipay.com/blob/master/www/teams/create.json.spt) which inspires the new [edit.json.spt](https://github.com/gratipay/gratipay.com/pull/3923/files#diff-6). 

The way it is written right now, we first write the team details to the db (with a unique generated `slug`) and *then* try to save the team image. If a user uploads an image of size  > 1Mb or an image which is not a jpg or png, the team creation won't be successful as far as the user is concerned and he'll resubmit the team application form with an appropriate image. But when he does again, we would have already created a slug for that team name resulting in a misleading message of `Sorry, there is already a team using <slug>.` when in fact the `slug` was created because we wrote the team details to the db first.

aandis:
1. Fixes #3929.
2. Modifies team update endpoint to process image *before* updating the team.
3. Showed `Unauthorized` message before `Rejected team` and `Closed team` message on team update pages.


aandis:
Can run `select * from teams where image_type is null` after merging this, to check out if there are many invalid teams. If so, can `DELETE` them.

whit537:
Ready for review? Looks like we haven't deployed #3932 yet. Let's maybe land this and then get both out the door.

whit537:
lgtm

whit537:
```
=> select count(*) from teams where image_type is null;
┌───────┐
│ count │
├───────┤
│    27 │
└───────┘
(1 row)
```

Can't delete them, many (all?) are simply old, from before we had image support at all.

aandis:
just 27. So I guess doesn't matter much, especially if they are teams being used.

aandis:
Couple of them, most recent ones, might be from me actually, I was testing this on production to reproduce the error. And maybe from someone else too if they tried https://github.com/gratipay/gratipay.com/issues/3929#issuecomment-190859268

whit537:
Hrm ... I'm finding two that seem to have been bitten by this. Reminds me of a third where I think we got bit by this as well ...

Issue #3880, opened 2015-12-18
Explain character restrictions in team names


mattbk:
E.g., you can't include at least some special characters like ":" as explained by https://gratipay.freshdesk.com/helpdesk/tickets/3677.

mattbk:
Part of https://github.com/gratipay/gratipay.com/issues/3859.

Issue #3426, opened 2015-05-14
Fix teams slug


rohitpaulk:
I reused `slugize` from the Community model in #3412, realized this when I found #3425.

All of the slugs are now lowercase, we should probably give teams the same freedom that users have with `username`.

whit537:
+1

PR #3943, opened 2016-03-06
Team name restrictions

aandis:
#3880

### DIFF
--- a/www/new.spt
+++ b/www/new.spt
@@ -1,7 +1,6 @@
 from datetime import datetime
 
 from aspen import Response
-from gratipay.models.community import slugize
 from gratipay.models.team import Team
 [---]
 request.allow('GET')
@@ -58,6 +57,7 @@ still_migrating = delta > 0
     <input type="hidden" name="csrf_token" value="{{ csrf_token }}">
 
     <label><h2>{{ _("Team Name") }}</h2></label>
+    <p><i>{{ _("Must contain alphabets. Can optionally contain numbers, dashes (-), underscores (_), periods (.), comma (,) and whitespaces.") }}</i></p>
     <input type="text" name="name" required autofocus>
 
     <label><h2>{{ _("Image") }}</h2></label>
--- a/www/new.spt
+++ b/www/new.spt
@@ -57,7 +57,7 @@ still_migrating = delta > 0
     <input type="hidden" name="csrf_token" value="{{ csrf_token }}">
 
     <label><h2>{{ _("Team Name") }}</h2></label>
-    <p><i>{{ _("Must contain alphabets. Can optionally contain numbers, dashes (-), underscores (_), periods (.), comma (,) and whitespaces.") }}</i></p>
+    <p><i>{{ _("At least one letter A through Z, plus numbers, dashes (-), underscores (_), periods (.), comma (,) and whitespace.") }}</i></p>
     <input type="text" name="name" required autofocus>
 
     <label><h2>{{ _("Image") }}</h2></label>
--- a/www/teams/create.json.spt
+++ b/www/teams/create.json.spt
@@ -72,7 +72,7 @@ if request.method == 'POST':
 
     try:
         fields['slug'] = slugize(fields['name'])
-    except AssertionError:
+    except InvalidTeamName:
         raise Response(400, _("Sorry, team name contains invalid characters."))
 
     try:
--- a/www/teams/create.json.spt
+++ b/www/teams/create.json.spt
@@ -4,8 +4,8 @@ from cStringIO import StringIO
 import requests
 from aspen import Response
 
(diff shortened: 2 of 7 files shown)


aandis:
@mattbk this look good?

www/teams/create.json.spt
```diff
@@ -71,6 +71,11 @@ if request.method == 'POST':
                                   ))
 
     try:
+        fields['slug'] = slugize(fields['name'])
+    except AssertionError:
```
rohitpaulk: I think it'd make more sense to throw a custom error - `InvalidTeamName`, maybe

mattbk:
I can't test right now, but looks like you're on the right track.

www/teams/create.json.spt
aandis: wokay! It did cross my mind too but it looked like a bit too much. :)

mattbk:
It's funny that you can't use these characters when creating a new team, but you can edit your team name to include any characters you want.

mattbk:
Hmm, my error messages are stacking up:
screen shot 2016-03-07 at 6 04 33 pm

mattbk:
A team name of all underscores collapses into a team with a name of length zero, and the form completes. The team ends up under Review, but the link goes nowhere.

rohitpaulk:
> It looks like all underscores and periods collapse into "-"

Side-effect of https://github.com/gratipay/gratipay.com/issues/3426, looks like

aandis:
> you can't use these characters when creating a new team, but you can edit your team name to include any characters you want.

There's a reason for this. You can't include those characters while creating the team because we use those characters to create a (unique) `slug` for the team and the current [method](https://github.com/gratipay/gratipay.com/blob/master/gratipay/models/community.py#L8) which does that expects only certain characters. But when editing the team, the `slug` is already created and you can't really change it, so you *should* be free to edit your team name however you want.

www/teams/create.json.spt
aandis: Another reason I didn't do this is that `slugize` is from `community` model. So having `Invalid*Team*Name` in there wouldn't really fit well. Now that you mentioned #3426, this would make sense since `team` model would have a `slugize` of its own.

mattbk:
Okay, thanks, that makes sense.

aandis:
Alright. Fixed #3880 and #3426 in this pr. Ready for review.

mattbk:
It works for me. I'm happy with it as long as someone else thinks the new `slugize` works right. @whit537 @rohitpaulk ?

whit537:
Reviewing ...

whit537:
Rebased on master at af05120f99596edcd61d2a3dda5077dd78247fca. Previous head was 8fa155ebef3a30d8c52d1373d722db2658bcdaa2.

www/new.spt
```diff
@@ -58,6 +57,7 @@ still_migrating = delta > 0
     <input type="hidden" name="csrf_token" value="{{ csrf_token }}">
 
     <label><h2>{{ _("Team Name") }}</h2></label>
+    <p><i>{{ _("Must contain alphabets. Can optionally contain numbers, dashes (-), underscores (_), periods (.), comma (,) and whitespaces.") }}</i></p>
```
whit537: I'm not familiar with this usage of "alphabets." Perhaps it's part of Indian English?

whit537: 323424b72f4c8be8f46e0c088a28c9993b4c887b

aandis: heh. It's petty [common](http://english.stackexchange.com/a/165611/106961) over here. I just realized myself that it's wrong though. :flushed:

whit537: No worries ... maybe we should have an Indian English translation! :dancer:

whit537: Well, "localization" is probably the better word. ;-)

whit537:
Rebased to remove 3c0d72fc31903f4d508a539154def1e1ac822011 in favor of cbe6dd7401d11596818e97d55bfe85cdce179830. Previous head was 48f9ae663ae7146c888a5a946157b3659a0f1e1b.

whit537:
Rebasing again to fix a couple tests ...

whit537:
@aandis I'm ready to merge if you're good with my latest commits ...

aandis:
Commented on https://github.com/gratipay/gratipay.com/commit/3c0d72fc31903f4d508a539154def1e1ac822011#commitcomment-16714701

whit537:
@aandis Back in 2196bd6d2a87799d187cc8cb48f0cbcbd000acef. :)

aandis:
Looks good to me. :+1:

whit537:
!m @aandis

aandis:
!m @whit537

PR #3965, opened 2016-03-15
fix plusses in emails

### LINKED ISSUE #3110: Email address verification email bug when email address contains + (plus sign)
Howdy, 

You sent me an email Subject: "Connect to Foo on Gratipay?" containing a link to verify my email address. That link didn't work, because my email address has a + character in it, and you didn't escape the + in the URL you sent in the email.

Incorrect URL:
https://gratipay.com/Foo/emails/verify.html?nonce=...&email=jay+waterpolice@jays.net

Correct URL:
https://gratipay.com/Foo/emails/verify.html?nonce=...&email=jay%2Bwaterpolice@jays.net

`+` becomes `%2B`

Thank you!

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/7661855-email-address-verification-email-bug-when-email-address-contains-plus-sign?utm_campaign=plugin&utm_content=tracker%2F85909&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F85909&utm_medium=issues&utm_source=github).
</bountysource-plugin>

whit537:
Let's fix #3110. I don't know yet what the issue is, but so far here's a *passing* test for escaping plusses in emails in the links we send out. At what point are those getting unescaped?

### DIFF
--- a/gratipay/utils/__init__.py
+++ b/gratipay/utils/__init__.py
@@ -134,44 +134,25 @@ def get_team(state):
 
 
 def encode_for_querystring(s):
-    prefix = b''
-    if not isinstance(s, bytes):
-        s = s.encode('utf8')
-    else:
-        # Check whether the string is binary or already utf8
-        try:
-            s.decode('utf8')
-        except UnicodeError:
-            prefix = b'.'
-    return prefix + b64encode(s, b'-_').replace(b'=', b'~')
+    """Given a unicode, return a unicode that's safe for transport across a querystring.
+    """
+    if not isinstance(s, unicode):
+        raise TypeError('unicode required')
+    return b64encode(s.encode('utf8'), b'-_').replace(b'=', b'~').decode('ascii')
 
 
 def decode_from_querystring(s, **kw):
-    def error():
+    """Given a unicode computed by encode_for_querystring, return the inverse.
+    """
+    if not isinstance(s, unicode):
+        raise TypeError('unicode required')
+    try:
+        return b64decode(s.encode('ascii').replace(b'~', b'='), b'-_').decode('utf8')
+    except:
         if 'default' in kw:
             # Enable callers to handle errors without using try/except.
             return kw['default']
         raise Response(400, "invalid input")
-
-    try:
-        s = s.encode('ascii')
-    except UnicodeError:
-        return error()
-
-    udecode = lambda a: a.decode('utf8')
-    if s[:1] == b'.':
-        udecode = lambda a: a
-        s = s[1:]
-    s = s.replace(b'~', b'=')
-    try:
-        return udecode(b64decode(s, '-_'))
-    except Exception:
-        try:
-            # For retrocompatibility
-            return udecode(b64decode(s))
-        except Exception:
-            pass
-        return error()
 
 
 def update_cta(website):
--- a/gratipay/utils/__init__.py
+++ b/gratipay/utils/__init__.py
@@ -149,6 +149,7 @@ def b64encode_s(s):
 def b64decode_s(s, **kw):
     def error():
         if 'default' in kw:
+            # Enable callers to handle errors without using try/except.
             return kw['default']
         raise Response(400, "invalid base64 input")
 
--- a/gratipay/utils/__init__.py
+++ b/gratipay/utils/__init__.py
@@ -143,6 +143,10 @@ def encode_for_querystring(s):
 
 def decode_from_querystring(s, **kw):
     """Given a unicode computed by encode_for_querystring, return the inverse.
+
+    We raise Response(400) if the input value can't be decoded (i.e., it's not
+    ASCII, not padded properly, or not decodable as UTF-8 once Base64-decoded).
+
     """
     if not isinstance(s, unicode):
         raise TypeError('unicode required')
--- a/gratipay/utils/__init__.py
+++ b/gratipay/utils/__init__.py
@@ -2,7 +2,7 @@
 
 from __future__ import absolute_import, division, print_function, unicode_literals
 
-from base64 import b64decode, b64encode
+from base64 import urlsafe_b64encode, urlsafe_b64decode
 from datetime import datetime, timedelta
 
 from aspen import Response, json
@@ -138,7 +138,7 @@ def encode_for_querystring(s):
     """
     if not isinstance(s, unicode):
         raise TypeError('unicode required')
-    return b64encode(s.encode('utf8'), b'-_').replace(b'=', b'~').decode('ascii')
+    return urlsafe_b64encode(s.encode('utf8')).replace(b'=', b'~').decode('ascii')
 
 
 def decode_from_querystring(s, **kw):
@@ -147,7 +147,7 @@ def decode_from_querystring(s, **kw):
     if not isinstance(s, unicode):
         raise TypeError('unicode required')
     try:
-        return b64decode(s.encode('ascii').replace(b'~', b'='), b'-_').decode('utf8')
+        return urlsafe_b64decode(s.encode('ascii').replace(b'~', b'=')).decode('utf8')
     except:
         if 'default' in kw:
             # Enable callers to handle errors without using try/except.
--- a/gratipay/utils/__init__.py
+++ b/gratipay/utils/__init__.py
@@ -133,7 +133,7 @@ def get_team(state):
     return team
 
 
-def b64encode_s(s):
+def encode_for_querystring(s):
     prefix = b''
     if not isinstance(s, bytes):
         s = s.encode('utf8')
@@ -146,7 +146,7 @@ def b64encode_s(s):
     return prefix + b64encode(s, b'-_').replace(b'=', b'~')
 
 
-def b64decode_s(s, **kw):
+def decode_from_querystring(s, **kw):
     def error():
         if 'default' in kw:
             # Enable callers to handle errors without using try/except.
--- a/gratipay/utils/__init__.py
+++ b/gratipay/utils/__init__.py
@@ -2,6 +2,7 @@
 
 from __future__ import absolute_import, division, print_function, unicode_literals
 
+from base64 import b64decode, b64encode
 from datetime import datetime, timedelta
 
 from aspen import Response, json
@@ -130,6 +131,46 @@ def get_team(state):
         raise Response(410)
 
     return team
+
+
+def b64encode_s(s):
+    prefix = b''
+    if not isinstance(s, bytes):
+        s = s.encode('utf8')
+    else:
+        # Check whether the string is binary or already utf8
+        try:
+            s.decode('utf8')
+        except UnicodeError:
+            prefix = b'.'
+    return prefix + b64encode(s, b'-_').replace(b'=', b'~')
+
+
+def b64decode_s(s, **kw):
+    def error():
+        if 'default' in kw:
+            return kw['default']
+        raise Response(400, "invalid base64 input")
+
+    try:
+        s = s.encode('ascii')
+    except UnicodeError:
+        return error()
+
+    udecode = lambda a: a.decode('utf8')
+    if s[:1] == b'.':
+        udecode = lambda a: a
+        s = s[1:]
+    s = s.replace(b'~', b'=')
+    try:
+        return udecode(b64decode(s, '-_'))
+    except Exception:
+        try:
+            # For retrocompatibility
+            return udecode(b64decode(s))
+        except Exception:
+            pass
+        return error()
 
 
 def update_cta(website):
--- a/gratipay/utils/__init__.py
+++ b/gratipay/utils/__init__.py
@@ -151,7 +151,7 @@ def decode_from_querystring(s, **kw):
         if 'default' in kw:
             # Enable callers to handle errors without using try/except.
             return kw['default']
-        raise Response(400, "invalid base64 input")
+        raise Response(400, "invalid input")
 
     try:
         s = s.encode('ascii')
--- a/tests/py/test_email.py
+++ b/tests/py/test_email.py
@@ -6,7 +6,7 @@ from gratipay.exceptions import CannotRemovePrimaryEmail, EmailAlreadyTaken, Ema
 from gratipay.exceptions import TooManyEmailAddresses, ResendingTooFast
 from gratipay.models.participant import Participant
 from gratipay.testing.emails import EmailHarness
-from gratipay.utils import emails
+from gratipay.utils import emails, b64encode_s
 
 
 class TestEmail(EmailHarness):
@@ -27,7 +27,8 @@ class TestEmail(EmailHarness):
         return P('/~alice/emails/modify.json', data, auth_as=user, **headers)
 
     def verify_email(self, email, nonce, username='alice', should_fail=False):
-        url = '/~%s/emails/verify.html?email=%s&nonce=%s' % (username, email, nonce)
+        # Email address is base64 encoded in url.
+        url = '/~%s/emails/verify.html?email64=%s&nonce=%s' % (username, b64encode_s(email), nonce)
         G = self.client.GxT if should_fail else self.client.GET
         return G(url, auth_as=username)
 
@@ -49,6 +50,13 @@ class TestEmail(EmailHarness):
         assert last_email['to'][0]['email'] == 'alice@gratipay.com'
         expected = "We've received a request to connect alice@gratipay.com to the alice account on Gratipay"
         assert expected in last_email['text']
+
+    def test_email_address_is_base64_encoded_in_sent_verification_link(self):
+        address = 'alice@gratipay.com'
+        encoded = b64encode_s(address)
+        self.hit_email_spt('add-email', address)
+        last_email = self.get_last_email()
+        assert "~alice/emails/verify.html?email64="+encoded in last_email['text']
 
     def test_verification_email_doesnt_contain_unsubscribe(self):
         self.hit_email_spt('add-email', 'alice@gratipay.com')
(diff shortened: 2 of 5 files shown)


aandis:
I tested this with a couple of emails having plus sign and couldn't reproduce this. It's probably (selectively) happening on mandrill's end? If yes, what can we do about that?

whit537:
Either on Mandrill's end or in actual mail clients (Gmail, etc.). More research required ...

whit537:
@aandis We should just be able to base64-encode the email address for transport across the URL. Working on a commit ...

whit537:
Actually, you wanna do that? :-)

whit537:
@aandis Does it make sense what I'm suggesting? In other words, instead of sending this link in email:

```
https://gratipay.com/Foo/emails/verify.html?nonce=...&email=jay+waterpolice@jays.net
```

We would send this:

```
https://gratipay.com/Foo/emails/verify.html?nonce=...&email=amF5K3dhdGVycG9saWNlQGpheXMubmV0
```

Then we would decode on the receiving end. That should work, no?

rohitpaulk:
^ I like that idea

aandis:
Sounds good. I'll get to this tomorrow.

aandis:
https://github.com/ankane/mailkick/issues/5#issuecomment-87163210 suggests this will work!

whit537:
There we go. :-)

!m @aandis

aandis:
Out of curiosity, why do we send the email in the verification link in the first place? The [`nonce`](https://github.com/gratipay/gratipay.com/blob/master/gratipay/models/participant.py#L398) generated is pretty much unique. Why not add a unique constraint on the `nonce` in db and use the nonce to get the email and verify it? Gets rid of this problem all together.

whit537:
@aandis Also an option. Whatever is lowest-hanging fruit here, I think. We've got bigger fish to fry.

whit537:
Which is to say, follow your :heart:, @aandis! :-)

aandis:
Alright. :)

rohitpaulk:
https://github.com/gratipay/gratipay.com/pull/3022#issuecomment-67180143 for reference

rohitpaulk:
> By requiring the user ID, an attacker can only brute-force one account at a time, instead of being able to attack all accounts at the same time. That means the probability of success becomes lower and constant (it won't grow with the number of users).

 > Right now it's not a big problem, but the probability of finding a good key increases with the number of users.

I know it still might not make a practical difference - but removing the email seems like a step backward to me :)

Changaco:
Email verification nonces aren't much of a security issue since they're not passwords (they don't log you in, they only set the verified bit on the address). Still, you don't want to use a random nonce as both an ID and a password, even when there is no real threat.

@aandis "pretty much unique" isn't enough, if it's not 100% unique then you need extra code to deal with conflicts, otherwise a few unlucky users could be prevented from verifying their emails. So, encoding the address with base64 is simpler.

Two more things:

- it's probably better to have a way to differentiate links in the new and old formats, for example by using another querystring key such as `email64`
- you can't use standard base64, it's not querystring-safe (it can have plusses, and equal signs for padding), in Liberapay we have a `b64encode_s` function

aandis:
Sorry I haven't gotten around to doing this yet. :disappointed:

@Changaco my reason for wording it as "pretty much unique" and not "perfectly unique" was exactly that. There's a very (very) low probability that they can collide, but nonetheless, there *is* a chance and I had it in mind to handle the collisions if we went this route. :)
Anyway, `base_64`  should be the way to go here then. I'll try to push the fix to liberypay as well. :)

Changaco:
Another thing I forgot to mention, including in the original comment that @rohitpaulk quoted, is that a DB index lookup is not guaranteed to run in constant time, so if you lookup a password (nonce or permanent, it doesn't matter) you're potentially exposed to a timing attack.

In short: never lookup a password, never put a `UNIQUE` constraint on a password column, always use `constant_time_compare()`.

> I'll try to push the fix to liberypay as well. :)

I look forward to it. :-)

aandis:
Ready for review. :)

tests/py/test_email.py
```diff
@@ -219,6 +220,13 @@ def test_html_escaping(self):
         assert 'foo&#39;bar' in last_email['html']
         assert '&#39;' not in last_email['text']
 
+    @mock.patch('gratipay.models.participant.Participant.send_email')
+    def test_emails_with_plus_are_linked_properly(self, send_email):
+        send_email.return_value = 1
+        self.alice.add_email("foo+bar@example.com")
+        link = send_email.args[1]
+        assert link.startswith('/~alice/emails/verify.html?email=foo%2Bbar%40example.com&nonce=')
+
```
aandis: How is this still passing? :scream:

aandis: I modified the verification link to contain `email64=<encoded-email>`. Here's the [test](https://github.com/gratipay/gratipay.com/commit/5f9320c4ec20a5bd1d88ed7072e363e60d0c0223#diff-a4f4fd2e3f2cfd01d60403c0c4c3a02fR54)

www/~/%username/emails/verify.html.spt
```diff
@@ -4,14 +4,17 @@ from datetime import timedelta
...
 [-----------------------------------------------------------------------------]
 
 participant = get_participant(state, restrict=False)
 banner = '~' + participant.username
 if participant == user.participant:
-    email = request.qs.get('email', '')
+    if 'email' in request.qs:
+        email = request.qs['email']
+    else:
+        email = b64decode_s(request.qs.get('email64', ''), default='')
```
whit537: The purpose of doing this (which @Changaco suggested at https://github.com/gratipay/gratipay.com/pull/3965#issuecomment-197789029, but didn't provide a reason for) is to handle the race condition when deploying, yes? I.e., when we deploy this, we may have some users who requested an email verification link before we deployed, but aren't going to click on it until after we deploy. Since the old `email` key works fine for most folks, we don't want to fail their verification unnecessarily, yes?

aandis: Yes. We stop sending out unencoded address in verification link but, to be backward compatible, we still support old style verification. Here's a [test](https://github.com/gratipay/gratipay.com/commit/5f9320c4ec20a5bd1d88ed7072e363e60d0c0223#diff-a4f4fd2e3f2cfd01d60403c0c4c3a02fR130) for the same. :)

whit537:
I'm trying to get Mandrill to send mail for me locally. Over in [OSF](https://github.com/CenterForOpenScience/osf.io) they have a dev hook that dumps email to the console so you don't have to have a real mailer configured. That'd be nice to have.

whit537:
I do find the key in `defaults.env` listed as a test key in the Mandrill UI ...

whit537:
Ah, test keys don't send mail. I have a personal key in `local.env` that I needed to uncomment.

whit537:
Okay! Seems to be working, I see the `email64` key in the URL now.

Let's look at the code ...

whit537:
Rebased on master at 14df2fea025a6acb10f5d2a9c861d4e03ead7132. Previous HEAD was 5f9320c4ec20a5bd1d88ed7072e363e60d0c0223.

www/~/%username/emails/verify.html.spt
whit537: Okay, so since verification emails [expire](https://github.com/gratipay/gratipay.com/blob/0ee9aa51064290ce6acbbfca0df23a8a81b1419f/gratipay/models/participant.py#L59) after 24 hours, we could drop `email` and rename `email64` to `email` after this has been deployed for 24 hours, yes?

gratipay/utils/__init__.py
```diff
@@ -132,6 +133,46 @@ def get_team(state):
...
+        return udecode(b64decode(s, '-_'))
+    except Exception:
+        try:
+            # For retrocompatibility
+            return udecode(b64decode(s))
+        except Exception:
+            pass
+        return error()
+
+
```
whit537: These look more complicated than I would expect. We should either add some tests (and possibly comments here) for all the behavior here, or cut out the behavior we don't want/need.

whit537: I more or less understand the `=` to `~` replacement (though I don't understand why it's necessary; `?foo=bar==` seems to survive transport just fine afaict).

I don't understand the `.` prefix and the bytes/unicode conversion.

whit537: I appreciate converting decoding errors into a 400 (to avoid a 500). The use-case for `default` doesn't jump out at me.

whit537: I'm inclined to merge anyway since @Changaco writes good code and the extra code here seems low-risk, just a bit of tech debt.

www/~/%username/emails/verify.html.spt
aandis: I think we should keep it for a while. What if someone's (old style) verification link is already expired by the time he clicks on it? If we keep this, he has the ability to get a new verification email through [this](https://github.com/gratipay/gratipay.com/blob/master/www/~/%25username/emails/verify.html.spt#L76). Probably better to re-ticket this removal and do this after a while when we are confident we are not losing any users because of this.

gratipay/utils/__init__.py
aandis: > I more or less understand the = to ~ replacement (though I don't understand why it's necessary; ?foo=bar== seems to survive transport just fine afaict).

I think it's just to be [safe](http://stackoverflow.com/a/941481/2643846).

> bytes/unicode conversion.

[`Base64`](https://docs.python.org/3/library/base64.html#base64.b64encode) converts *binary* to text. In python 2, a normal string `isinstance` of `bytes`, so that'll work. But in python 3, every string is a `unicode` object which is not `isinstance` of `bytes`. Hence the conversion. Related issue - [liberapay#235](https://github.com/liberapay/liberapay.com/issues/235)

> . prefix

This is a little tricky to explain. If we do encode the string to `utf-8` while encoding to `base64` as mentioned above, then, while decoding from `base64`, we want to [decode](https://github.com/gratipay/gratipay.com/pull/3965/files#diff-29dde85b4bbb7c47ef404ed0ab9f054eR160) the string from `utf-8` as well to get the original string back. *But..* if we do not encode the string to `utf-8` *and* the string had some characters which can't be decoded with `utf-8` ([try-except block](https://github.com/gratipay/gratipay.com/pull/3965/files#diff-29dde85b4bbb7c47ef404ed0ab9f054eR142)) , we [*do not*](https://github.com/gratipay/gratipay.com/pull/3965/files#diff-29dde85b4bbb7c47ef404ed0ab9f054eR162) want to apply the `utf-8` decoding while `base64` decoding. The `.` prefix is just used as a marker for that.
Not sure how much sense that makes, but if you run the encoding/decoding with `'\x80abc'` as an example, it should be clear.

> use-case for default

It's useful for providing a more helpful context specific error message. For example, instead of showing `invalid base64 input` as the error message in email verification, we let the caller handle the error through `default` and show a more helpful error message to the user. Not very relevant in this case maybe, but useful if we use encoding in future at other places as well. :)

www/~/%username/emails/verify.html.spt
whit537: > through [this](https://github.com/gratipay/gratipay.com/blob/master/www/~/%25username/emails/verify.html.spt#L76)

(Quick point of order: in GitHub file links, swap `master` for the [current version number](https://gratipay.com/version.txt) to get a nice permalink, to avoid drift.)

gratipay/utils/__init__.py
whit537: > Related issue - liberapay#235

Heh. You write good code, too! ;-)

tests/py/test_email.py
aandis: @whit537 before merge, we should probably remove this test as well. But I still don't understand why this is passing in the first place. The link now starts with `/~alice/emails/verify.html?email64=`, so the assert *should* fail. Something I don't know about `mock` yet?

gratipay/utils/__init__.py
whit537: > I think it's just to be [safe](http://stackoverflow.com/questions/941468/c-sharp-mvc-trailing-equal-sign-in-url-doesnt-hit-route/941481#941481).

Safe against bugs in ASP.NET MVC? Aspen has bugs, but not that one. :-P

This is a simple enough conversion, though. I'm fine with it. 

> It's useful for providing a more helpful context specific error message.

Why can't the caller `try`/`except` for `Response`?

> bytes/unicode conversion.
> . prefix

Still wrapping my head around this. I'm trying to satisfy myself that, in addition to making sense, this extra complexity doesn't introduce any security vulnerabilities ...

tests/py/test_email.py
whit537: Roger.

gratipay/utils/__init__.py
aandis: > Safe against bugs in ASP.NET MVC

It could happen with [jquery](http://stackoverflow.com/questions/7235874/jquery-query-parameter-parsing-in-query-string-breaks-query-string-up) for all we know. ;)

> Why can't the caller try/except for Response?

Hm. Good idea. We could do that given that this was indeed @Changaco's intention for using `default`. Is that right @Changaco ?

whit537: > It could happen with [jquery](http://stackoverflow.com/questions/7235874/jquery-query-parameter-parsing-in-query-string-breaks-query-string-up) for all we know. ;)

Fair enough. !m @aandis

aandis: Examining some use cases([1](https://github.com/liberapay/liberapay.com/blob/f624590da031943dda62c3ee99b4e2a045bca55a/templates/base.html#L111), ([2](https://github.com/liberapay/liberapay.com/blob/6e98bd14f41a47b8cd9c146a4136e65f4dc02aa2/www/%25username/wallet/payin/%25back_to.spt#L78), [2,1](https://github.com/liberapay/liberapay.com/blob/6e98bd14f41a47b8cd9c146a4136e65f4dc02aa2/www/%25username/wallet/payin/%25back_to.spt#L120)), [3](https://github.com/liberapay/liberapay.com/blob/19ac142d7ee2360b1371c2fae2c5c3f4b8c84552/www/%25username/wallet/payout/%25back_to.spt#L25)), looks like `default` was used as a cleaner way to handle errors and avoid the long `try/except` block.

aandis: That combined with reason mentioned in https://github.com/gratipay/gratipay.com/pull/3965#discussion_r58257449, I think it's okay to keep that as well. :)

gratipay/utils/__init__.py
```diff
@@ -132,6 +133,46 @@ def get_team(state):
...
+    udecode = lambda a: a.decode('utf8')
+    if s[:1] == b'.':
+        udecode = lambda a: a
+        s = s[1:]
+    s = s.replace(b'~', b'=')
+    try:
+        return udecode(b64decode(s, '-_'))
+    except Exception:
+        try:
+            # For retrocompatibility
```
whit537: ... with liberapay's version?

gratipay/utils/__init__.py
whit537: > bytes/unicode conversion.
> . prefix

I don't like that we're so relaxed wrt bytes and unicode here. Seems like a slide back to Python 2 ways of thinking. Why do we need to accept (in `b64encode_s`) and return (from `b64decode_s`) bytes? Why not only accept and return unicode? Our use case for these functions is to transport strings to ourselves across a querystring. We're not trying to send images across the querystring. I propose that we rename the functions to `pack_for_querystring` and `unpack_from_querystring`, and that we constrain them to accept and return only unicodes.

www/~/%username/emails/verify.html.spt
whit537: > Probably better to re-ticket this removal and do this after a while when we are confident we are not losing any users because of this.

Commented in b3422ccd14ebd726f573ff22b329f4bfa0671951.

gratipay/utils/__init__.py
whit537: > looks like `default` was used as a cleaner way to handle errors and avoid the long `try/except` block. 

Commented in 855a86df27a080fb4012e8db078b03e50b12c25a.

whit537: > `pack_for_querystring` and `unpack_from_querystring`

Or `encode_for_querystring` and `decode_from_querystring`, to at least keep the verbs. The point is that this isn't really a Base64 algorithm anymore, it's a higher-order algorithm that happens to use Base64 as one of its parts.

whit537: Implemented as a PR against this one: https://github.com/gratipay/gratipay.com/pull/3977.

tests/py/test_email.py
whit537: >  Something I don't know about `mock` yet?

Yup, you called it: `link` is a mock, which thinks everything is a-okay. Guido [pointed](https://twitter.com/gvanrossum/status/648617596011417600) this up recently.

```
(Pdb) l
246         def test_emails_with_plus_are_linked_properly(self, send_email):
247             send_email.return_value = 1
248             self.alice.add_email("foo+bar@example.com")
249             link = send_email.args[1]
250             import pdb; pdb.set_trace()
251  ->         assert link.startswith('/~alice/emails/verify.html?email=foo%2Bbar%40example.com&nonce=')
252     
253         def test_can_dequeue_an_email(self):
254             larry = self.make_participant('larry', email_address='larry@example.com')
255             larry.queue_email("verification")
256     
(Pdb) pp link
<MagicMock name=u'send_email.args.__getitem__()' id='4383627408'>
(Pdb)
```

gratipay/utils/__init__.py
whit537: This question would be obsolete with #3977.

Changaco: > ... with liberapay's version?

No, compatibility with standard base64.

gratipay/utils/__init__.py
Changaco: > The point is that this isn't really a Base64 algorithm anymore

Yes it is, it's an algorithm to encode/decode arbitrary bytes into/from 64 ASCII chars. The decode function even supports standard base64.

whit537:
Okay! Punchlist:

 - [ ] rebase
 - [ ] remove/update old test—https://github.com/gratipay/gratipay.com/pull/3965#discussion_r58259909

whit537:
Rebased on master at e2c6062d378b74afdba4fa9b7591b0ea487172c8, previous head was 1d7e0cd35848ed7edba2445dd05ad8e68f05de1f.

whit537:
@aandis I'm ready to land this when you are. What's left?

whit537:
Derp. Missed an import. Force-pushed 770f0b2d13276260d1041c70a35c46b1326af7b8 instead of 8a73270ce4a04f750d2974b19ecd0f5ad48f9320.

whit537:
Alright, @aandis, travis is green. We ready to merge? :open_mouth:

aandis:
Yep! :dancer:

PR #3973, opened 2016-03-30
Added comment to defaults.env to avoid invalid reports about public keys.

peterjaric:


whit537:
Thanks @peterjaric! :-)

Issue #3289, opened 2015-03-26
implement a customer identification program (CIP)


whit537:
With the move to a lower-level processing infrastructure (#67, #3377), we need a customer identification program (CIP) as part of our AML program (https://github.com/gratipay/inside.gratipay.com/issues/119). Beyond that, we'll need a customer due diligence (CDD) program: https://github.com/gratipay/inside.gratipay.com/issues/219.

(Ftr, Transpay rejected us partially for not having this (https://github.com/gratipay/gratipay.com/issues/417#issuecomment-78266767), and Stripe wanted this, too (https://github.com/gratipay/gratipay.com/issues/3245#issuecomment-85999520).)
##

Changaco:
Stripe doesn't want the identity of the sender in the way you're implying, it just wants to connect payouts to charges, because a credit card number is sufficient information for the authorities to identify the sender.

What we need is identity information for payins that use anonymous methods like Bitcoin.

What we probably don't need is the identity of users who merely transfer money they've received to another Gratipay account, they're just smokescreens and should be ignored.

whit537:
+1 from Braintree at https://github.com/gratipay/gratipay.com/issues/3287#issuecomment-91681028.

techtonik:
Ids and personal information will makes Gratipay a likely object for cyber attacks. That needs a completely different level of security and effort. I'd really like this stuff to be provided as opt-in for service that really need it with explanation why we need it. Otherwise we are no different from all ubiquitous Spyware 2.0 around there.

whit537:
Taking this off of Balanced Shutdown but leaving as three-star.

whit537:
Back on Balanced Shutdown. I think we seriously risk Citizens rejecting us if we don't have this.

whit537:
> Ids and personal information makes Gratipay a likely object for cyber attacks. That needs a completely different level of security and effort.

Agreed. See https://github.com/gratipay/inside.gratipay.com/issues/214.

> I'd really like this stuff to be provided as opt-in and outsourced validation service that really need it with explanation why we need it.

We can't make it opt-in. We need to verify identity for, at a minimum, all of our customers (receivers), for [these reasons](https://github.com/gratipay/inside.gratipay.com/issues/119#issuecomment-107081615). We were outsourcing this to Balanced, but the point of this ticket is that with our new processing infrastructure we are taking on ownership of our own compliance program.  +1 for explaining why we need to collect and verify identity.

whit537:
[...]

PR #3976, opened 2016-04-01
add schema to store national identity information for ~users

whit537:
&#x2B11; #3994 

#4006 &rarr;

Alright, let's do this. The eating of an :elephant: begins with the first bite! :fork_and_knife:

#### Punchlist

 - [x] require `schema_version` in `info` field
 - [x] add client-side encryption to `info` field(!)
 - [x] log events
 - [x] finish docs
 - [x] clean up commits

#### PRs

 - [x] identity preliminaries: #4004 (against `master`), includes ...
  - [x] fix the Makefile for sphinx: #3996 (against `master`)
  - [x] make way for refactoring participant: #3997 (against `master`)
  - [x] document `add_event`
  - [x] prune dead `crypto` code
 - [x] add new crypto code: #3998 (against `master`)
 - [x] add schema, data, and Python for countries: #3999 (against #3998)
 - [x] add schema and Python for storing and retrieving identities: #4000 (against #3999)
 - [x] add Python for verifying and unverifying identities: #4001 (against #4000)
 - [x] add Python for clearing an identity: #4002 (against #4001)
 - [x] add schema and trigger for `participants.has_verified_identity`: #4003 (against #4002)


whit537:
I'm thinking that we'll use [client-side encryption](http://www.postgresql.org/docs/9.4/static/encryption-options.html) on the `personal_identities.info` field.

At this point we know very little about what exactly we'll need to collect. It may vary from country to country, and I suspect it'll be driven largely by the third-party verification vendor(s) we end up going with. Going with a JSON field for now seemed like the easiest way to get started.

I think we should allow multiple identities per ~user, but only one per ~user per country (we should account for dual citizenship, in other words).

I guess let's log changes to the `events` table, with the old info in the payload as with ... wherever else it was that we did that recently. :-)

whit537:
I'm expecting that eventually we'll add a `organizational_identities` table that links with `teams`.

whit537:
> At a minimum, the bank must obtain the following identifying information from each customer before opening the account:
>
Name.
Date of birth for individuals.
Address.
Identification number.

https://github.com/gratipay/gratipay.com/issues/3289#issuecomment-107086274

whit537:
Turns out we still have an identity form still [hanging around](https://github.com/gratipay/gratipay.com/blob/035e5b755344e5649686a0535710ea68c1020d89/www/~/%25username/identity.spt) post-Balanced.

mattbk:
> I guess let's log changes to the events table, with the old info in the payload as with ... wherever else it was that we did that recently. :-)

https://github.com/gratipay/gratipay.com/pull/3923#issuecomment-187322860

sql/countries.sql
```diff
@@ -0,0 +1,250 @@
...
+INSERT INTO countries (name, code) VALUES ('Vanuatu', 'VUT');
+INSERT INTO countries (name, code) VALUES ('Venezuela (Bolivarian Republic of)', 'VEN');
+INSERT INTO countries (name, code) VALUES ('Viet Nam', 'VNM');
+INSERT INTO countries (name, code) VALUES ('Virgin Islands (British)', 'VGB');
+INSERT INTO countries (name, code) VALUES ('Virgin Islands (U.S.)', 'VIR');
+INSERT INTO countries (name, code) VALUES ('Wallis and Futuna', 'WLF');
+INSERT INTO countries (name, code) VALUES ('Western Sahara*', 'ESH');
+INSERT INTO countries (name, code) VALUES ('Yemen', 'YEM');
+INSERT INTO countries (name, code) VALUES ('Zambia', 'ZMB');
+INSERT INTO countries (name, code) VALUES ('Zimbabwe', 'ZWE');
```
aandis: geez. Why not a single multi row `INSERT` statement?

whit537: Laziness, sorry. Cleaned up in d78a132e9b03392e26b504ec54714ec2fea1b56f. :blush:

whit537:
Rebased on master at 4270aa0. Previous head was 1b97231aa7dff1f6c4232c8c814d1968caec5b33.

whit537:
Lotta discussion over here. A summary:

 - Individuals can renounce their citizenship. We should allow ~users to deactivate the national identities they have stored with us.
 - Countries fork and merge. This would be case of mass deactivation and reregistration.
 - Deactivating one's last `participant_identities` should a) either remove you from all Teams, or b) be prevented if you're still on a Team. Probably the former?
 - Our recent pattern has been to include old info in the `events` table when we log changes. However, we want to keep PII in one place to, we don't want to also store it in the `events` table. What we're thinking is that we'll use our old ctime/mtime pattern (a la `payment_instructions`)—still also logging to `events`, but with the ids of the old and new records rather than the full info.
 - We don't want to overengineer the `info` schema right now. We'll collect (https://github.com/gratipay/gratipay.com/pull/3976#issuecomment-204204890) name, address, date of birth, and identity number (allowing for national id, drivers license, and/or passport). We should version the schema so we're in a better position as we need to start validating it more strictly.

whit537:
MIT flags: https://www.gosquared.com/resources/flag-icons/

kaguillera:
Sprite/css flags alternative as discussed
https://www.flag-sprites.com/

kaguillera:
flags

CSS: (https://gist.github.com/kaguillera/267ee747bd1e7679732b81d7b2f78d10)

whit537:
I'm now questioning why we need to retain old PII. Seems safer to simply delete or overwrite it as requested.

whit537:
More like that (previous was fd6d63599cf657417ba28ad010222fe7279674db).

whit537:
screen shot 2016-04-28 at 11 47 44 pm

whit537:
Okay! Got a good chunk of work done here. I am ruminating on how to carve this up into easily reviewable PRs ...

whit537:
Alright, got some good commits and I made PRs for them. I've update the ticket description here with a list. It's a little wonky because half-way through I started making the PRs against the previous PR instead of `master` because I thought that might be easier, but due to the way we have Travis configured we're not getting CI results unless the PR is against `master`. I've updated this PR to contain all of the commits, and I've pushed them one at a time so we can see Travis results for them.

whit537:
Hmm ... actually, it looks like Travis may still be confused.

whit537:
Okay, I've gone ahead and consolidated the first four commits into a PR, #4004. I think we should start with that as a piece for review and merge, and then we can tackle the other six. I guess you can start reviewing the six either in the commits on this PR or in the separate PRs (again, see the description on this ticket for links). I think it probably makes sense to rebase and merge in ascending order, rather starting at the end and merging back up the chain. In other words, I guess we'll close instead of merging the latter five PRs, making *new* PRs against `master` for each commit as we're ready.

Of course, you may find enough wrong with this code that we'll end up with a new commit history anyway. :-)

whit537:
Okay! I've marked this one and #4004 as ready for review! Here we go! 🐘 🍴💃 

cc: @rohitpaulk @aandis @rorepo @kaguillera @clone1018 et al.

whit537:
We shouldn't allow users to store a national identity without a verified email address.

whit537:
Rebased on master @279b7ca, previous head was 385fae2498842f1c5301f755858ba40d0aed247c.

whit537:
On to #3998! 😮

whit537:
> We shouldn't allow users to store a national identity without a verified email address.

For some reason I thought we already had `participants.has_verified_email`, a la `has_verified_identity` in #4003, but now I don't see it, either in Postgres or in Python. I guess we should insert a PR into the chain that has that, and then use it subsequently. I suppose it should precede https://github.com/gratipay/gratipay.com/pull/4000, because we want to make storage of an identity conditional on it.

whit537:
> I thought we already had `participants.has_verified_email`

It's not a boolean, it's the address itself: `participants.email_address`.

whit537:
Old head was 9907401d90528f061df0928cf61828963550db60.

PR #3977, opened 2016-04-02
implement qs value handling as a higher order encoding

whit537:
Implements https://github.com/gratipay/gratipay.com/pull/3965#discussion_r58274069. Making this a PR against #3965 to hopefully avoid stepping on toes. :-)


whit537:
Ping @aandis. What do you think of this approach?

Changaco:
Not sure if this is a good solution for gratipay or not, but it wouldn't work for liberapay: our use of `b64{en,de}code` is not limited to querystrings, and dropping support for bytestrings could easily result in 500s that wouldn't be caught by the tests, which would be bad for a function whose purpose is to avoid 500s.

whit537:
>  dropping support for bytestrings could easily result in 500s

Passing bytestrings where unicode strings are expected *should* be a 500. If/when we need to encode/decode binary data we should use different functions designed for that purpose.

Changaco:
> Passing bytestrings where unicode strings are expected *should* be a 500.

You want to sacrifice user experience for the sake of code "purity"? If we had static typing then it could be a compile-time error, but we don't, and in this case the test suite can't replace static checking.

whit537:
> You want to sacrifice user experience for the sake of code "purity"?

That's not what I see going on here. If we end up passing a bytestring into `encode_for_querystring` then that's *our* bug (i.e., should result in an HTTP 500 so we find out about it and can fix it), not a result of user input (&rarr; 400). We own/control the data type of the input we receive from users, and we shouldn't conflate byte strings and text (unicode) strings.

Changaco:
Dropping support for bytestrings means that instead of a 200 you'll get a 500. It's not a change from a 400 to a 500.

whit537:
@Changaco Why would we ever pass a bytestring to `decode_from_querystring`?

Changaco:
Does it matter? The point is that it can happen. It's fairly easy to pass `bytes` instead of `str` by mistake, especially in python2.

I'm dropping out of this discussion now, I've said what I had to say. :-)

whit537:
You good to merge this, then, @aandis? What's your take?

aandis:
I think it's okay to keep this (with a few improvements) for now since our use case is only querystring encoding. If and when, we decide to use encoding at other places, we'll modify this. :)

aandis:
> with a few improvements

387a0c0f03f89dae47fe4503dec9040eed4e8dba

gratipay/utils/__init__.py
```diff
@@ -133,45 +133,26 @@ def get_team(state):
...
-def b64decode_s(s, **kw):
-    def error():
+def decode_from_querystring(s, **kw):
+    """Given a unicode computed by encode_for_querystring, return the inverse.
+    """
+    if not isinstance(s, unicode):
+        raise TypeError('unicode required')
+    try:
+        return urlsafe_b64decode(s.encode('ascii').replace(b'~', b'=')).decode('utf8')
+    except:
```
aandis: This is a little confusing. Why does `decode_from_querystring` require `unicode` if it only ends up encoding it into `ascii`? Doesn't that mean, it can only process `ascii`?

whit537: > Doesn't that mean, it can only process `ascii`?

Yes, it will fail with `Response(400)` if passed a `unicode` that can't be encoded as `ascii`. Note that we do `decode` it again following the `b64decode` (as `utf8`), so that this function both takes and returns a `unicode`.

The Python data type of the input value must be `unicode`, and the character set of the input value must be ASCII.

whit537: @aandis Does dd508d8ea119ae8994871885af5f64af064a337f help at all?

whit537:
@aandis 387a0c0f03f89dae47fe4503dec9040eed4e8dba looks good to me.

whit537:
If you're okay with dd508d8ea119ae8994871885af5f64af064a337f then I think we're good to merge, ya?

aandis:
:+1:

PR #3980, opened 2016-04-05
Add script to fix stale cached values

rohitpaulk:
Earlier (in Gratipay 1.0), we used to fund tips using the Gratipay balance that participants had. We used to refresh these values after every Payday. 

Now that we've moved to charging in arrears and never using one's balance to fund tips - the old 'funded' values have to be corrected.

This was picked off https://github.com/gratipay/gratipay.com/pull/3876#issuecomment-164177764 (https://github.com/gratipay/gratipay.com/pull/3876#issuecomment-164177764 for reference)

PR #3981, opened 2016-04-05
Cap unfunded dues

rohitpaulk:
This is a redo of https://github.com/gratipay/gratipay.com/pull/3876 with a cleaner commit log.

### DIFF
--- a/sql/branch.sql
+++ b/sql/branch.sql
@@ -1,4 +1,14 @@
 BEGIN;
     UPDATE payment_instructions SET due = 0 WHERE amount = 0 AND due != 0;
-    UPDATE payment_instructions SET due = 0 WHERE due > 9.41;
+
+    UPDATE payment_instructions pi
+       SET due = 0
+     WHERE due > 9.41
+       AND (                          -- Doesn't have a valid CC
+             SELECT COUNT(*)
+               FROM current_exchange_routes r
+              WHERE r.participant = (SELECT id FROM participants WHERE username = pi.participant)
+                AND network = 'braintree-cc'
+                AND error = ''
+           ) = 0;
 END;
--- a/sql/branch.sql
+++ b/sql/branch.sql
@@ -1,3 +1,4 @@
 BEGIN;
+    UPDATE payment_instructions SET due = 0 WHERE amount = 0 AND due != 0;
     UPDATE payment_instructions SET due = 0 WHERE due > 9.41;
 END;
--- a/sql/branch.sql
+++ b/sql/branch.sql
@@ -1,0 +1,3 @@
+BEGIN;
+    UPDATE payment_instructions SET due = 0 WHERE due > 9.41;
+END;
--- a/sql/payday.sql
+++ b/sql/payday.sql
@@ -186,13 +186,15 @@ CREATE OR REPLACE FUNCTION process_payment_instruction() RETURNS trigger AS $$
               FROM payday_participants p
              WHERE username = NEW.participant
         );
+
         IF (NEW.amount + NEW.due <= participant.new_balance OR participant.card_hold_ok) THEN
             EXECUTE pay(NEW.participant, NEW.team, NEW.amount + NEW.due, 'to-team');
             RETURN NEW;
-        ELSE
+        ELSIF participant.has_credit_card THEN
             EXECUTE park(NEW.participant, NEW.team, NEW.amount + NEW.due);
             RETURN NULL;
         END IF;
+
         RETURN NULL;
     END;
 $$ LANGUAGE plpgsql;
--- a/tests/py/test_billing_payday.py
+++ b/tests/py/test_billing_payday.py
@@ -102,6 +102,28 @@ class TestPayday(BillingHarness):
         assert obama.get_due('TheEnterprise') == D('0.00')
 
     @mock.patch.object(Payday, 'fetch_card_holds')
+    def test_payday_only_adds_to_dues_if_valid_cc_exists(self, fch):
+        Enterprise = self.make_team(is_approved=True)
+        self.obama.set_payment_instruction(Enterprise, '4.00')
+        assert self.obama.get_due('TheEnterprise') == D('0.00')
+
+        fch.return_value = {}
+        Payday.start().run()    # payday 0
+
+        assert self.obama.get_due('TheEnterprise') == D('4.00')
+
+        fch.return_value = {}
+        self.obama_route.update_error("failed") # card fails
+        Payday.start().run()    # payday 1
+
+        assert self.obama.get_due('TheEnterprise') == D('4.00')
+
+        fch.return_value = {}
+        Payday.start().run()    # payday 2
+
+        assert self.obama.get_due('TheEnterprise') == D('4.00')
+
+    @mock.patch.object(Payday, 'fetch_card_holds')
     def test_payday_does_not_move_money_below_min_charge(self, fch):
         Enterprise  = self.make_team(is_approved=True)
         self.obama.set_payment_instruction(Enterprise, '6.00')  # not enough to reach MINIMUM_CHARGE


sql/branch.sql
```diff
@@ -0,0 +1,3 @@
+BEGIN;
+    UPDATE payment_instructions SET due = 0 WHERE due > 9.41;
```
whit537: This means that anyone with dues and a failing credit card is getting a big reset. I think that means we're leaving a little money on the table. An alternative would be to reset due to `9.41`, and then if/when they fix their card, we'd charge them that much instead of having to build back up. Right?

Was that the original intention at https://github.com/gratipay/gratipay.com/pull/3876#discussion-diff-51040424?

whit537: We [aren't]( https://github.com/gratipay/gratipay.com/blob/61fbb2a4baa00d2da2d8cc457a890b4af71abd4d/tests/py/test_billing_payday.py#L119) otherwise resetting to zero after a card fails. Maybe we should make an effort to avoid doing that in the repair step here?

rohitpaulk: Aha, you're right. That was the original intention, yes - But even that diff is just an approximation. 

```sql
UPDATE payment_instructions SET due = floor(9.41/(amount)) * amount WHERE due > 9.41;
```

^ That assumes that the user didn't change the tip amount. For example - let's say the user had a 5 dollar tip amount. 

### Payday 1 

Charged: 0
Due: $5

### Payday 2

Attempted to charge, failed
Due: $10 (our old code does this)

Expected behaviour after our script runs is that the due is reset to **$5**. 

What happens if the user changes the tip amount after Payday 2 though? If they changed the tip amount to $1, let's say - we'll end up setting the due to `floor(9.41/(amount))`, i.e. **$9**

rohitpaulk: Given the two options - I think the resetting to $9.41 option would be safer though...

rohitpaulk: Hmm, maybe we could use https://github.com/gratipay/gratipay.com/issues/3851 to truly solve this?

whit537: > Hmm, maybe we could use #3851 to truly solve this?

I think we'd need a pretty good reason to block this further than we already have.

> I think the resetting to $0 option would be safer though...

Agreed. How many users are affected if we go this route? I suspect it's not very many ...

whit537: Wait ... don't we also want to tie this to `is_funded`? Otherwise we reset *all* dues back to 0, even for folks with a good card attached.

whit537:
```
=> select count(*) from payment_instructions where due > 9.41 and not is_funded;
┌───────┐
│ count │
├───────┤
│   649 │
└───────┘
(1 row)

=> select count(*) from payment_instructions where due > 9.41 and is_funded;
┌───────┐
│ count │
├───────┤
│     5 │
└───────┘
(1 row)
```

rohitpaulk: > don't we also want to tie this to is_funded

Ideally - we shouldn't have to. No person with a valid CC should ever have a due more than 9.41. These look like more or less the same cases we found on https://github.com/gratipay/gratipay.com/pull/3980#issuecomment-206063806. Let's check back here once I've identified the issue there

whit537: The five includes [these four](https://github.com/gratipay/gratipay.com/pull/3980#issuecomment-206063806).

whit537: For the fifth, `amount = 0`, meaning due *should* be reset to zero.

whit537: > Let's check back here once I've identified the issue there

Okay.

rohitpaulk:
Rebased on master.

sql/branch.sql
rohitpaulk: > For the fifth, amount = 0, meaning due should be reset to zero.

Done in https://github.com/gratipay/gratipay.com/pull/3981/commits/cc1d5f10a8cd89b276543309534490e43996f2af

rohitpaulk: > What happens if the user changes the tip amount after Payday 2 though?

I think that we should be able to isolate cases where this has happened.. And if we're looking to get this out the door before https://github.com/gratipay/gratipay.com/issues/3851 is done - the best way to go about this is reset the dues to $0 for only those users. For users who did *not* change the tip amount at all, the original [`UPDATE` statement](https://github.com/gratipay/gratipay.com/pull/3876#discussion-diff-51040424) should work just fine

> Agreed. How many users are affected if we go this route? I suspect it's not very many ...

In total - it'd be a lot of people (`649`, to be exact). But after isolating the ones that we can't properly repair without #3851, it might not be many... Time to run some queries..

whit537:
@rohitpaulk With https://github.com/gratipay/gratipay.com/pull/3981#discussion_r59018518, should we take this out of `Review` for now? Sounds like you want to do a little more digging before we merge this, ya?

sql/branch.sql
rohitpaulk:
```sql
SELECT COUNT(*) 
  FROM payment_instructions 
 WHERE due > 9.41;

#=> 654

SELECT COUNT(*) 
  FROM current_payment_instructions 
 WHERE due > 9.41;                      -- Should be the same as above

#=> 654

SELECT COUNT(*)
  FROM current_payment_instructions cpi
 WHERE due > 9.41
   AND (
            SELECT COUNT(*)                    -- Tips with a different amount exist. 
              FROM payment_instructions pi
             WHERE cpi.participant = pi.participant
               AND cpi.team = pi.team
               AND amount != cpi.amount
       ) > 0;


#=> 34
```

rohitpaulk: > But after isolating the ones that we can't properly repair without #3851, it might not be many...

So that's just 34 people out of 654. I'm going to modify the script to set dues to zero for those 34 people (because we run the risk of overcharging them if we go with an approximation). For the rest of the users, we'll go with the original UPDATE query that sets the due to the highest multiple of the tip amount that is `<` the minimum charge

rohitpaulk:
Ran https://github.com/gratipay/gratipay.com/pull/3981/commits/7a10ee91ef6e3dec29f603471fa4454b2c604bbd on a recent backup, here's the output.

```
➜  gratipay.com git:(new-cap-unfunded-dues) ✗ psql gratipay < sql/branch.sql
BEGIN
UPDATE 13
UPDATE 32
UPDATE 620
COMMIT
```

sql/branch.sql
```diff
@@ -0,0 +1,24 @@
+BEGIN;
+    -- Tips that have been set to zero should have zero dues.
+
+    UPDATE payment_instructions SET due = 0 WHERE amount = 0 AND due != 0;
```
whit537: P.S. Do we do this normally? When I drop my tip to zero, does my due get reset as well?

whit537: [Yes](https://github.com/gratipay/gratipay.com/blob/7a10ee91ef6e3dec29f603471fa4454b2c604bbd/gratipay/models/participant.py#L856).

rohitpaulk: @whit537 - Yes, https://github.com/gratipay/gratipay.com/blob/master/gratipay/models/participant.py#L856

whit537:
Alright, so how does this interact with the straggler at https://github.com/gratipay/gratipay.com/pull/3980#issuecomment-206693875?

whit537:
Will this accidentally reset their due to zero before we've given them enough time to respond?

rohitpaulk:
> Will this accidentally reset their due to zero before we've given them enough time to respond?

That's a good question... Let me check

rohitpaulk:
The particular user - they haven't changed their tip, so we'll be resetting to the multiple below 9.41... which we don't want to do. I guess I should modify the script to exclude people with working credit cards then. On it..

whit537:
> The particular user - they haven't changed their tip, so we'll be resetting to the multiple below 9.41

Confirmed.

> I guess I should modify the script to exclude people with working credit cards then.

:+1:

whit537:
The payment instruction in question has `false` `is_funded`. I would expect `true` if they had a working card, no?

rohitpaulk:
> The payment instruction in question has false is_funded. I would expect true if they had a working card, no?

Hmm, yes...

sql/branch.sql
```diff
@@ -0,0 +1,24 @@
...
+               FROM payment_instructions pi2
+              WHERE pi.participant = pi2.participant
+                AND pi.team = pi2.team
+                AND amount != pi.amount
+           ) > 0;
+
+    -- For the rest of the payment_instructions, we set due to the highest multiple of the
+    -- tip amount that is lower than the minimum charge.
+
+    UPDATE payment_instructions SET due = floor(9.41/(amount)) * amount WHERE due > 9.41;
```
whit537: I'm not sure about this.

 - Shouldn't it be the *lowest* multiple that is *higher*?
 - What about people who had *two* tips set up? We compute minimum due based on all payment instructions together (and "tip" sure is easier to write than "payment instruction" :).

rohitpaulk:
The [last time we checked](https://github.com/gratipay/gratipay.com/pull/3980#issuecomment-206063806), `is_funded` was `true`. What could've changed?

rohitpaulk:
@whit537 - I just ran the query and it still is `true`, can you double check?

whit537:
> What could've changed?

Sorry, I'm working from an old backup. It's `true` on prod.

rohitpaulk:
`SELECT * FROM payment_instructions WHERE participant = 'xxx';` is what I ran

whit537:
Yeah, my bad, sorry.

sql/branch.sql
rohitpaulk: > Shouldn't it be the lowest multiple that is higher?

Ah, yes. True. We charge their CC the day they exceed the minimum charge. 

> What about people who had two tips set up? We compute minimum due based on all payment instructions together 

Hmm, this is starting to get tricky..

rohitpaulk: The only cases where we can properly repair are: 

1. They have only one tip set up
2. They've never changed the tip amount.

whit537: Really I suppose we'd want to replay `paydays` and `payment_instructions` backwards (via the `events` table?) until the point at which we charged them and their card failed.

whit537: > Hmm, this is starting to get tricky..
> Really I suppose we'd want to replay

Well, we've got [620](https://github.com/gratipay/gratipay.com/pull/3981#issuecomment-207423101) folks in this boat, at roughly $10 each, so call it $6,000 we'd be leaving on the table if we hard-reset those 620 to zero. Some—many? most?—of those folks will *never* fix their card and start giving again.

Raises the question of whether dues shouldn't expire after a while anyway. If I forget about Gratipay and come back a year later and re-up, I'd be surprised if I got hit with $10 that had been waiting there for a year. Should we expire dues after a 30 (60, 90) days?

rohitpaulk: Yep, brings us back to https://github.com/gratipay/gratipay.com/pull/3981#discussion_r58665580 I guess

whit537: Alright, I think we cut bait. Meaning, reset everyone back to zero. We have so much to work on, I think our time is better spent on other things. Unless you really want to keep going down this rabbit hole?

rohitpaulk: > Unless you really want to keep going down this rabbit hole?

Not really... 

I'm going to force push the diff to the original state.

whit537: Okay.

rohitpaulk:
Output after running the script before https://github.com/gratipay/gratipay.com/pull/3981/commits/29400fef5046b707e434a97f4a1cb7b684ed202f - 

```
BEGIN
UPDATE 13
UPDATE 652
COMMIT
```

whit537:
Hmm ... how does that related to `is_funded`?

```
=> select count(*) from payment_instructions WHERE due > 9.41 and not is_funded;
┌───────┐
│ count │
├───────┤
│   652 │
└───────┘
(1 row)

=> select count(*) from payment_instructions WHERE due > 9.41 and is_funded;
┌───────┐
│ count │
├───────┤
│     2 │
└───────┘
(1 row)
```

rohitpaulk:
My backup is a day old, let me try getting a new one..

whit537:
The 652 / 2 counts are from prod.

rohitpaulk:
```
gratipay=# select count(*) from payment_instructions WHERE due > 9.41 and not is_funded;
 count
-------
   652
(1 row)

gratipay=# select count(*) from payment_instructions WHERE due > 9.41 and is_funded;
 count
-------
     2
(1 row)
```

```
BEGIN
UPDATE 13
UPDATE 651
COMMIT
```

whit537:
Perfect. Let's do it!

rohitpaulk:
> Perfect. Let's do it!

@whit537 - Something I'm missing? `652 != 651`?

whit537:
@rohitpaulk Yeah, that's okay. One of the two is part of the 13. ;-)

rohitpaulk:
Aha, yes gotcha. :)

rohitpaulk:
Verified. 

```
gratipay=# UPDATE payment_instructions SET due = 0 WHERE amount = 0 AND due != 0;
UPDATE 13
gratipay=# select count(*) from payment_instructions WHERE due > 9.41 and not is_funded;
 count
-------
   651
(1 row)

gratipay=#
```

PR #3983, opened 2016-04-06
fix method whitelisting

whit537:
We're ending up with 500 instead of 405, because of a misordering in the algorithm functions. The error handling needs a `user` object, and also apparently some `i18n`.

This surfaced on [H1-127218](https://hackerone.com/reports/127218).


whit537:
Ping @rohitpaulk @aandis.

PR #3984, opened 2016-04-06
Fix review dashboard, fixes #3937

### LINKED ISSUE #3937: Fix review dashboard
Our admin review dashboard is supposed to list all users who have attached a credit card and have not been whitelisted. 

The logic we use to pick these participants from the database seems flawed though - 

https://github.com/gratipay/gratipay.com/blob/master/www/dashboard/index.spt#L16

The check should be for whether an `exchange_route` exists, not on `braintree_customer_id`

rohitpaulk:
This changes the condition by which we filter users for review.

The new conditions are:

1) The user must have tips setup with a net value > 0. (And this should be calculated without using cached values)
2) The user must have a payin/payout route attached.

### DIFF
--- a/www/dashboard/index.spt
+++ b/www/dashboard/index.spt
@@ -9,11 +9,16 @@ unreviewed = website.db.all("""
 
     SELECT username
          , balance
-         , giving
-         , ngiving_to
-      FROM participants
+         , (SELECT SUM(amount) FROM current_payment_instructions WHERE participant = p.username) AS giving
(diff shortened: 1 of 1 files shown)


rohitpaulk:
With this change, I now only see 5 users on the review dashboard (from a recent backup) - and they include the 4 users found at https://github.com/gratipay/gratipay.com/pull/3980#issuecomment-206063806.

whit537:
!m @rohitpaulk

PR #3996, opened 2016-04-29
fix the Makefile for Sphinx

whit537:
It was depending on swaddle, which is gone in favor of honcho. Still requires that you `pip install sphinx`, but then you can `make html` and `cd _build/html && python -m SimpleHTTPServer &`.

### DIFF
--- a/docs/Makefile
+++ b/docs/Makefile
@@ -3,7 +3,7 @@
 
 # You can set these variables from the command line.
 SPHINXOPTS    =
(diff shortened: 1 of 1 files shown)


whit537:
Ping @rohitpaulk @aandis @kaguillera for review.

whit537:
(Actually, I think you need to `make rst` first; this is all in the `docs` directory.)

whit537:
Closing in favor of #4004.

PR #3997, opened 2016-04-29
make way for refactoring participant

whit537:
This moves `gratipay.participant` from a single file to a submodule directory. The file is so big, I don't want to add more methods to it in #3976. I think instead that we should go back to mixins, which I intend to start implementing on #3976 (I'll rebase if/when this lands).

This includes #3996 so I'll rebase once that's resolved.


whit537:
Closing in favor of #4004.

PR #3998, opened 2016-04-30
implement symmetric encryption

whit537:
&larr; #4004 — #3999 &rarr;

##### Punchlist

- [x] rebase on master once #4004 lands
- [x] fix typo in comment—https://github.com/gratipay/gratipay.com/pull/3998#discussion_r61734031 —done in 2a6bbd4f7c5845f06fa5a3fe67ce398059c9c80a
- [x] audit [Fernet](https://github.com/fernet/spec/blob/master/Spec.md)—https://github.com/gratipay/gratipay.com/pull/3998#issuecomment-216227070
- [x] audit [`cryptography`](https://cryptography.io)—https://github.com/gratipay/gratipay.com/pull/3998#issuecomment-216227070
- [x] sort out key policies
- [x] comment config knob—https://github.com/gratipay/gratipay.com/pull/3998#discussion_r62547619 —done in 34ba9cd7ce4a6d0d6c4cc59889537d2f7ece5d36
- [x] simplify EncryptingPacker init API—https://github.com/gratipay/gratipay.com/pull/3998#discussion_r62547429 —addressed at https://github.com/gratipay/gratipay.com/pull/3998#discussion_r62554979
- [x] test decryption with old key—https://github.com/gratipay/gratipay.com/pull/3998#discussion_r62548278 —done in 5ce0ff87542872569bc472b26e86270355380cf3

Issue #3994, opened 2016-04-20
Bring back takes for Team Gratipay


whit537:
This is the main ticket for the "[Bring Back Take-What-You-Want for Team Gratipay](https://github.com/gratipay/gratipay.com/milestone/14)" milestone.

#### Done

This ticket/milestone is done when we successfully distribute Gratipay Team revenue to ~users as part of payday. The ~~six~~ five ~users we are targeting for the first distribution are: ~~@kaguillera~~ @rohitpaulk @mattbk @aandis @clone1018 @whit537.

##### National Identity

- [x] deploy schema and Python for storing national identity for ~users: #3976
 - [x] implement symmetric encryption: #3998 
 - [x] add countries: ~~#3999~~ &rarr; #4027
 - [x] implement encrypted national identities: ~~#4000~~ &rarr; #4028
 - [x] \[un]verifying identities: ~~#4001~~ &rarr; #4031
 - [x] clearing identities: ~~#4002~~ &rarr; #4032
 - [x] has_verified_identity: ~~#4003~~ &rarr; #4033
- [x] implement fake data for participant identities: #4006 
- [x] deploy UI for publicly displaying country info of ~users: #4007
- [x] deploy UI for storing national identity for ~users: ~~#4008~~ &rarr; #4034
- [x] 👾  bonus PRs: #4035 #4036

##### Takes

- [x] make room for team changes: #4037 
- [x] 👾  convert client-side tests to Splinter: #4079
- [x] make more room for team changes 😁 : #4068 #4070
- [x] bring back the old UI behind a flag: #4072
- [x] start fixing the UI: #4073
- [x] migrate `_mixin_team.py`: #4074
- [x] cherry-pick #4023: #4075
- [x] cherry-pick #4024: #4076
- [x] update some Python APIs: #4099
- [x] bring back the takes UI: #4077
- [x] 👾  bonus PRs: #4108 #4109 #4110
- [x] bring back payday: #4102
- [x] include team members in MassPay: #4111
- [x] ~~review computed values algorithms~~ &rarr; reticketed #4120
- [x] ~~review docs for new mixins~~ &rarr; reticketed #4121
- [x] ~~fill out the test suite for the mixins, prune old tests~~ &rarr; reticketed #4122

##### Finish!

- [x] use UI to add identities
- [x] use Python API to verify identities
- [x] decide on https://github.com/gratipay/inside.gratipay.com/issues/747
- [x] use SQL to set `available` for Team Gratipay
[...]

PR #3999, opened 2016-04-30
add schema and Python for countries

whit537:
&#x2B11; #3994

&#x2B11; #3976

&larr; #3998 — #4000 &rarr;

##### Punchlist

- [x] remove `code3` and `name` fields: https://github.com/gratipay/gratipay.com/pull/3999#issuecomment-217748706
- [x] take #4008 on board
- [ ] remake against `master` once #3998 lands


whit537:
Discovered in #4005 that the official country names aren't suitable for display. E.g., "United States of America (the)" should be "United States" or "the United States" depending on context.

whit537:
Actually, it [turns out]((https://github.com/gratipay/gratipay.com/pull/4005#issuecomment-216083680)) that we already have good names in `locale.countries` ! 👍 

So do we need the names in the `countries` table at all? Flags (788d92d19817260bc356f0961f5c605d4cbc52ec), Babel, and pretty much everything I'm finding uses the two-letter country code. Should we simplify down to a single column? Will @kaguillera let us use the 2-letter country code as the primary—and only—key?

kaguillera:
Yes...that is cool if that is what everything uses

whit537:
> Yes...that is cool if that is what everything uses

Hah! :D Well, on the other hand, this would mean updating all of the subsequent PRs that depend on `country_id`.

screen shot 2016-05-01 at 8 50 44 pm

kaguillera:
I would suggest that we do that now before we have to do that further down the line...is that difficult or too much work??

whit537:
@kaguillera I think we're okay with our own `id` field, and alpha-2 as a separate field. Do you see anything actively bad about that approach?

kaguillera:
No...I am more inclined to that approach.

whit537:
Rebased on #3998 at 049bd48, previous head was fd1ca952020be35c1b9db7c83aa97cd114717625.

whit537:
We can probably drop the `name` field and maybe even `code3` since we'll be using `code2` to index into Babel for names. We'll do the same for flags, and likely any other purpose in the foreseeable future. Holding `name` especially in our database seems redundant and confusing, since we're not actually using that data in the UI as one might expect.

whit537:
Okay, I roughed in #4008. We should look through that for insight on changes we may want to make here—it may make sense to use `code2` more thoroughly after all.

whit537:
Okay, removed `name` and `code3` in 133388ae1b5634368883865b38a47feec5e56ea3. We're not using `name` anywhere, and `code3` was only used in the database. @kaguillera and I decided to keep the separate `id` field on principle. ☺️

whit537:
Remade as #4027.

PR #4001, opened 2016-04-30
add schema and Python for [un]verifying identities

whit537:
&#x2B11; #3994

&#x2B11; #3976

&larr; #4000 — #4002 &rarr;

##### Punchlist

 - [ ] remake against `master` once #4000 lands

PR #4003, opened 2016-04-30
implement has_verified_identity

whit537:
&#x2B11; #3994

&#x2B11; #3976

&larr; #4002 — #4006 &rarr;

##### Punchlist

- [x] move `has_verified_identity` logic up into Python so that it's easier to use in #4018 
- [x] `set_attributes` for `has_verified_identity` so #4018 is easier
- [ ] remake against `master` once #4002 lands

Issue #2449, opened 2014-06-01
verify individual identity programmatically


whit537:
Airbnb uses https://www.jumio.com/.

This could help free us up to use other payment processors in addition to Balanced, including non-U.S. options.
##

whit537:
+1 https://gratipay.freshdesk.com/helpdesk/tickets/1736

whit537:
+1 from Braintree at https://github.com/gratipay/gratipay.com/issues/3287#issuecomment-91681028.

whit537:
Looks like there's a number of companies in this space:

https://www.google.com/search?q=identify+verification+api

whit537:
Here's a list:

* [Trulioo](https://www.trulioo.com/)
* [IDchecker](http://www.idchecker.com/strongid/)
* [Socure](http://www.socure.com/)
* [Veratad IDMatch](http://veratad.com/solutions/identity-verification/)
* [Cloudcheck](http://www.verifidentity.com/cloudcheck/)
* [Experian](http://www.experian.com/decision-analytics/identity-and-fraud/identity-verification-compliance.html)
* [Integrity](http://integrity.aristotle.com/products/)
* [miiCard](http://www.miicard.com/news/201210/miicard-opens-api-first-global-trusted-id-service-platform)
* [Plaid](https://www.plaid.com/)
* [Clearbit](https://clearbit.com/)
* [MicroBilt](http://www.microbilt.com/)
* [BlockScore](https://blockscore.com/)
* [Jumio](https://www.jumio.com/)
* [Intuit](https://developer.intuit.com/docs/0020_customeraccountdata/0005_service_features)
* [Yodlee](http://www.yodlee.com/)
* [Accuity](http://www.accuity.com/)
* [EveryoneAPI](https://www.everyoneapi.com/)
* [IDology](https://www.idology.com/)

whit537:
We need global coverage.

whit537:
Intuit, Yodlee and Plaid (which looks really nice) are bank account verification APIs. On this ticket want identity verification.

whit537:
What's the relationship between bank account verification and ID verification? Why do we want global reach if we're only able to pay out to U.S. bank accounts right now? (Can Zipmark to IAT?)

Cloudcheck is New Zealand only. IDology is US, UK, Canada, and Mexico. EveryoneAPI is really flimsy.

Remaining:

1. [Trulioo](https://www.trulioo.com/)
[...]

PR #4005, opened 2016-05-01
UI for displaying verified country for ~users

whit537:
Part of #3994. Land #3976 first.

#### Punchlist

 - [x] comprehend overlap between countries and available flags
 - [x] implement a default for when we have no flag (Christmas Island)
 - [x] make the flag smaller at smaller page sizes
 - [x] tinker with flag position
 - [x] tinker with flag popping
 - [x] check it with two and three flags
 - [x] consider hard-limiting to three displayed identities (maybe a "more" link? seems really rare for individuals ... but Teams ...)
 - [x] i18n
 - [x] use the `is_verified` filter
 - [ ] clean up commits

#### Pushing Back Upstream

 - [x] clean up names ("United States of America (the)" &rarr; "the United States"): https://github.com/gratipay/gratipay.com/pull/3999#issuecomment-216040547
 - [x] filtering on `is_verified` before slicing: https://github.com/gratipay/gratipay.com/pull/4001#issuecomment-216038499


whit537:
screen shot 2016-04-30 at 9 20 13 pm

whit537:
4f11ad43b5203a573431082fd1bc5dd34070f3f8:

screen shot 2016-05-01 at 7 59 09 am

whit537:
7e3ddbe7c4919dd17caf9c64bd04f81fdc90b0c2:

screen shot 2016-05-01 at 8 09 06 am

whit537:
Two works, but three is a little too much:

screen shot 2016-05-01 at 8 10 05 am

whit537:
Capped to two in 31056d4905a5ee25e635168893366869c37470b1:

screen shot 2016-05-01 at 8 11 40 am

whit537:
> tinker with flag popping

The options aren't great here, because of the way the sprite image is built: since not all flags are the same form factor (I'm looking at you, Nepal), it's hard to put styles on individual flags. I want to keep them on the banner for now since they're part of the basic metadata about a participant. I think they're good enough for now.

whit537:
screen shot 2016-05-01 at 8 15 44 am

whit537:
It'd be helpful for `list_identity_metadata` to be able to filter by `is_verified`, to make it easier to both filter *and* cap.

whit537:
Hmm ... the small flags from www.flag-sprites.com are a little *too* small ...

screen shot 2016-05-01 at 8 34 11 am

whit537:
I think the regular size are better:

screen shot 2016-05-01 at 8 35 50 am

whit537:
Alright, tightened up a bit, will probably look gross for Nepal:

screen shot 2016-05-01 at 8 37 51 am

whit537:
Actually, not too bad.

926e8e684bdfb9597e54b4ecab1ea1d6273ff102:

screen shot 2016-05-01 at 8 50 21 am

screen shot 2016-05-01 at 8 50 25 am

whit537:
I was wondering if this was possible (username crashing into flags):

screen shot 2016-05-01 at 12 30 48 pm

whit537:
"No flag available"?

screen shot 2016-05-01 at 12 52 37 pm
screen shot 2016-05-01 at 12 52 49 pm

whit537:
Or a simple checkmark? "Verified!"

screen shot 2016-05-01 at 1 04 21 pm

whit537:
> comprehend overlap between countries and available flags

```
[gratipay] $ run defaults.env local.env ./flags.py 

Missing flags:
  AQ Antarctica
  AX Åland Islands
  BL Saint Barthélemy
  BQ Bonaire, Sint Eustatius and Saba
  BV Bouvet Island
  CC Cocos (Keeling) Islands (the)
  CW Curaçao
  CX Christmas Island
  FK Falkland Islands (the) [Malvinas]
  GF French Guiana
  GS South Georgia and the South Sandwich Islands
  HM Heard Island and McDonald Islands
  IO British Indian Ocean Territory (the)
  MF Saint Martin (French part)
  MP Northern Mariana Islands (the)
  NF Norfolk Island
  NU Niue
  PM Saint Pierre and Miquelon
  PN Pitcairn
  SH Saint Helena, Ascension and Tristan da Cunha
  SJ Svalbard and Jan Mayen
  SS South Sudan
  SX Sint Maarten (Dutch part)
  TF French Southern Territories (the)
  TK Tokelau
  UM United States Minor Outlying Islands (the)
  WF Wallis and Futuna
  YT Mayotte

Extra flags:
  AN

[gratipay] $ cat flags.py
```

```python
#!/usr/bin/env python2

from gratipay import wireup

env = wireup.env()
db = wireup.db(env)

# copy/pasted from flags.scss
have = set("""\
AD AE AF AG AI AL AM AN AO AR AS AT AU AW AZ BA BB BD BE BF BG BH BI BJ BM BN BO BR BS BT BW BY BZ
CA CD CF CG CH CI CK CL CM CN CO CR CU CV CY CZ DE DJ DK DM DO DZ EC EE EG EH ER ES ET FI FJ FM FO
FR GA GB GD GE GG GH GI GL GM GN GP GQ GR GT GU GW GY HK HN HR HT HU ID IE IL IM IN IQ IR IS IT JE
JM JO JP KE KG KH KI KM KN KP KR KW KY KZ LA LB LC LI LK LR LS LT LU LV LY MA MC MD ME MG MH MK ML
MM MN MO MQ MR MS MT MU MV MW MX MY MZ NA NC NE NG NI NL NO NP NR NZ OM PA PE PF PG PH PK PL PR PS
PT PW PY QA RE RO RS RU RW SA SB SC SD SE SG SI SK SL SM SN SO SR ST SV SY SZ TC TD TG TH TJ TL TM
TN TO TR TT TV TW TZ UA UG US UY UZ VA VC VE VG VI VN VU WS YE ZA ZM ZW""".split())

want = {c.code2: c.name for c in db.all('select * from countries')}

missing = set(want) - have
extra = have - set(want)

print()
print("Missing flags:")
for code2 in sorted(missing):
    print(" ", code2, want[code2])
print()
print("Extra flags:")
for code2 in sorted(extra):
    print(" ", code2)
print()
```

whit537:
Well, I think we can live with that discrepancy. To start with all we need are `IN`, `TT`, and `US`—which we have. :-)

screen shot 2016-05-01 at 1 40 09 pm

💃

whit537:
https://github.com/gratipay/gratipay.com/pull/4005#issuecomment-216053914 addressed in 7a9378ee73033944da000f327b217512293db143

whit537:
The last big thing here is figuring out how we're going to do country names and i18n. It's an exponential problem: we want all countries translated into all languages. And it kinda depends on the phrasing we use: "Verified in the Netherlands," or just "The Netherlands"? Tweaking the wording just a little bit would ripple out into 250 countries x however many languages.

whit537:
We can reduce the surface area somewhat by focusing on particular countries. This gets me thinking ahead to https://github.com/gratipay/gratipay.com/issues/2449: when we start verifying identity programmatically, we're going to have to scope our efforts to just a few countries at first, expanding from there. At some point I put together a list of the countries already represented on Gratipay ... now where is that ...

whit537:
[Found it](https://github.com/gratipay/inside.gratipay.com/issues/256#issuecomment-112926564)!

> [H]ere is a list of 31 countries in which we have had customers in the past (sorted by percentage of our customer base—roughly half our customers have been in the U.S.):
>
United States, Great Britain, Canada, France, Germany, Spain, Poland, Australia, New Zealand, The Netherlands, Finland, India, Brazil, Sweden, Vietnam, Singapore, Pakistan, Japan, Denmark, Czech Republic, Bulgaria, Argentina, Hungary, Korea, Russia, Slovenia, China, Austria, Chile, Ireland, Belgium

whit537:
TTO isn't on there, I note ...

whit537:
> The last big thing here is figuring out how we're going to do country names and i18n.

Hmmm ... https://github.com/onomojo/i18n-country-translations?

whit537:
That's a nice place to start. However, I think we'll have to let go of a construction such as "verified in the Netherlands" and go with something like "verified in: Netherlands."

whit537:
> Hmmm ... https://github.com/onomojo/i18n-country-translations?

Nevermind ... Babel [already](http://babel.pocoo.org/en/latest/locale.html) gives us this!

> Imagine for example you have a list of countries that users can choose from, and you’d like to display the names of those countries in the language the user prefers. Instead of translating all those country names yourself in your application, you can make use of the translations provided by the locale data included with Babel, which is based on the Common Locale Data Repository (CLDR) developed and maintained by the Unicode Consortium.

The [old identities form](https://github.com/gratipay/gratipay.com/blob/1952/www/~/%25username/identity.spt#L116-L118) already uses this. ☺️

whit537:
👍 

screen shot 2016-05-01 at 8 28 11 pm

whit537:
Oh yeeeeeeeaaaaaaahhhhh!

36262a3b15e121d5955181c8c0a1e9af079614da:

hroyzsevfx

whit537:
Rebased on #4003 at 9907401d90528f061df0928cf61828963550db60. Previous head was 36262a3b15e121d5955181c8c0a1e9af079614da.

whit537:
> It'd be helpful for `list_identity_metadata` to be able to filter by `is_verified`, to make it easier to both filter *and* cap.

8b098f93f42f05a7758b3b2e002672c86714c7a2:

screen shot 2016-05-01 at 10 29 49 pm
screen shot 2016-05-01 at 10 29 42 pm

whit537:
Alright, scrapping this one for parts ...

PR #4010, opened 2016-05-04
plumb payroll permissions

whit537:
&#x2B11; #3994

&larr; #4009 — #4012 &rarr;

#### Punchlist 

- [ ] rebase on master once #4009 lands

PR #4011, opened 2016-05-04
plumb payroll

whit537:
&#x2B11; #3994

&larr; #4010

#### Punchlist 

- [ ] rebase on master once #4010 lands


whit537:
Oops, wrong base. Closing for now in favor of #4012. Could reopen since we'll need to re-PR against master anyway.

PR #4025, opened 2016-05-11
bring back takes in payday

whit537:
&#x2B11; #3994

&larr; #4024 — #4026 &rarr;

#### Punchlist 

- [x] rewrite `process_takes`
- [x] insert a `process_baselines` before `process_takes`
- [ ] include members in MassPay; see `get_ready_payout_routes_by_network`
- [ ] clean up test suite (depends on `take_over` in #4023)
- [ ] remake against `master` once #4024 lands

PR #4026, opened 2016-05-11
bring back takes/membership UI

whit537:
&#x2B11; #3994

&larr; #4025

#### Punchlist 

- [ ] remake against `master` once #4025 lands


whit537:
Decided at https://github.com/gratipay/gratipay.com/issues/3994#issuecomment-229949952 to start with the old UI instead of from scratch.

PR #4027, opened 2016-05-11
Add countries

whit537:
Remake of #3999.


whit537:
"Ready?"
"Mmm-hmm."

👍

PR #4055, opened 2016-06-14
Hide gunicorn version

whit537:
This is a slight security improvement.

https://hackerone.com/reports/123742

Verified working in a sandbox:

```
$ curl -IL http://gratipay537.herokuapp.com/
HTTP/1.1 200 OK
Connection: keep-alive
Server: gunicorn
Date: Tue, 14 Jun 2016 16:13:59 GMT
Transfer-Encoding: chunked
X-Content-Type-Options: nosniff
X-Frame-Options: SAMEORIGIN
X-Gratipay-Version: 1967
X-Xss-Protection: 1; mode=block
Content-Type: text/html; charset=UTF-8
Cache-Control: no-cache
Set-Cookie: csrf_token=[]; expires=Tue, 21 Jun 2016 16:13:59 GMT; Path=/
Via: 1.1 vegur

$
```

### DIFF
--- a/Procfile
+++ b/Procfile
@@ -1,1 +1,1 @@
-web: gunicorn gratipay.main:website --bind :$PORT $GUNICORN_OPTS
+web: gunicorn gratipay.main:website --conf hide_gunicorn_version.py --bind :$PORT $GUNICORN_OPTS
--- a/hide_gunicorn_version.py
+++ b/hide_gunicorn_version.py
@@ -1,0 +1,2 @@
+import gunicorn
+gunicorn.SERVER_SOFTWARE = 'gunicorn'


whit537:
Ready for review, @aandis @rohitpaulk.

rohitpaulk:
We're open source - couldn't someone just check our github repo to find out what version of `gunicorn` we're running? 💭

rohitpaulk:
My bad, I just read the hacker one report + comments.

whit537:
@rohitpaulk Yes. However, I find [this rejoinder](https://hackerone.com/reports/123742#activity-844524) more or less convincing: 

> I believe this report warrants additional consideration for two reasons. First, company may be "open" company and thus not worried about exposing component version information but this does increase the likelihood of future exploit. Services like Shodan (www.shodan.io) allow searching by server information, which makes fingerprintable servers some of the first targets when a new exploit is found. Second, finding the version information indicates out-of-date components are in use, my second recommendation should be considered.

Automated exploit tools don't scrape GitHub, but they do look at the `Server` header.

Also, we get this report a lot, so implementing this should cut down some noise.

PR #4046, opened 2016-06-02
switch to {participant,team}_id in payment_instructions

whit537:
Part of #835.

PR #4057, opened 2016-06-15
extend test coverage a bit

whit537:
Salvaged from #4046.


whit537:
Ready for review, @rohitpaulk @aandis et al.

PR #3363, opened 2015-04-25
Remove symlinks - fixes #3318

techtonik:
The tests start to fail on my machine, but it seems that the reason is the test suite logic, not the same as in #3165.

PR #4060, opened 2016-06-15
Fix bug in test_pages where we ignored many pages

whit537:
Crept in in 8313ca6089703ac2bc4db0ee898a425ffc50191b (#3363).

 - [ ] rebase after #4057 lands


rohitpaulk:
Both this and https://github.com/gratipay/gratipay.com/pull/4057 look good to me, merging this and closing that.

PR #4078, opened 2016-07-07
Added Gratipay team to fakedata so pricing page will work

kaguillera:
Was updating my local repo and realized that http://localhost:8537/about/pricing was not working because it expected a Gratipay team to exist in fake_data. 
Only needed minor changes.

### DIFF
--- a/gratipay/utils/fake_data.py
+++ b/gratipay/utils/fake_data.py
@@ -311,12 +311,11 @@ def populate_db(db, num_participants=100, ntips=200, num_teams=5, num_transfers=
     teams = []
     teamowners = random.sample(participants, num_teams)
     for teamowner in teamowners:
-        if num_teams > 1:
-            teams.append(fake_team(db, teamowner))
-        else:
-            # Creating a fake Gratipay Team 
-            teams.append(fake_team(db, teamowner, "Gratipay"))
-        num_teams -= 1            
+        teams.append(fake_team(db, teamowner))
+        
+    # Creating a fake Gratipay Team 
+    teamowner = random.choice(participants) 
+    teams.append(fake_team(db, teamowner, "Gratipay"))
 
     print("Making Payment Instructions")
     npayment_instructions = 0
--- a/gratipay/utils/fake_data.py
+++ b/gratipay/utils/fake_data.py
@@ -115,17 +115,17 @@ def fake_participant_identity(participant, verification=None):
     return country_id
 
 
-def fake_team(db, teamowner):
+def fake_team(db, teamowner, teamname=None):
     """Create a fake team
     """
     isapproved = [True, False]
     productorservice = ['Product','Service']
 
-    teamname = faker.first_name() + fake_text_id(3)
-    teamslugname = faker.city()
-
+    if teamname is None:
+        teamname = faker.first_name() + fake_text_id(3)
+    
     try:
-        teamslug = slugize(teamslugname)
+        teamslug = slugize(teamname)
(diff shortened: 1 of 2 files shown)


gratipay/utils/fake_data.py
```diff
@@ -309,9 +309,14 @@ def populate_db(db, num_participants=100, ntips=200, num_teams=5, num_transfers=
 
     print("Making Teams")
     teams = []
+    num_teams -= 1      # Make room for the fake Grtipay Team
```
rohitpaulk: I don't think this is required, it should be okay if we have `num_teams` *inclusive* of the Gratipay team :)

gratipay/utils/fake_data.py
```diff
@@ -309,9 +309,14 @@ def populate_db(db, num_participants=100, ntips=200, num_teams=5, num_transfers=
 
     print("Making Teams")
     teams = []
+    num_teams -= 1      # Make room for the fake Grtipay Team 
     teamowners = random.sample(participants, num_teams)
     for teamowner in teamowners:
         teams.append(fake_team(db, teamowner))
+    
+    # Creating a fake Gratipay Team 
+    teamowner = random.sample(participants, 1)[0]
```
rohitpaulk: `random.choice(foo)` might look cleaner than `random.sample(foo, 1)[0]` 😉

rohitpaulk:
Other than those two nitpicks, this looks good to me

gratipay/utils/fake_data.py
whit537: `Grtipay` &rarr; `Gratipay`

kaguillera: @rohitpaulk I did that because of the test for the fake data. It checked to see if `num_teams` created were correct. Since I had to created the Gratipay Teams separate from the random generated Teams and Owner I made room so that the `fake_data` tests would still pass without the need to change the test. 
I am open to suggestions as to how to do this more elegantly. I now that it is not the ideal work around. 
Another way I was thinking would look something like this  
```
    print("Making Teams")
    teams = []
    teamowners = random.sample(participants, num_teams)
    for teamowner in teamowners:
        if num_teams > 1:
            teams.append(fake_team(db, teamowner))
        else:
            # Creating a fake Gratipay Team
            teams.append(fake_team(db, teamowner, "Gratipay"))
        num_teams -= 1
```

gratipay/utils/fake_data.py
kaguillera: I guess the above code in the comment would make this line a little cleaner but it adds more line. I was really trying not to change up to much of the code in too many places.

whit537: Can you explain further? If I'm not mistaken, `random.choice(foo)` and `random.sample(foo, 1)[0]` are logically equivalent, and the former saves six bytes over the latter. ;-)

gratipay/utils/fake_data.py
whit537: Discussed IRL. I suggest changing the test rather than this code.

gratipay/utils/fake_data.py
kaguillera: I misunderstood what @rohitpaulk was getting at 😦 ...now on the same page.

whit537:
Looks good to me. Rebasing ...

whit537:
Previous head was 4aa23ba832787fdb95a7748cc5ad5e8474a6556c.

whit537:
Will merge when Travis is green ...

whit537:
!m @kaguillera 

On to the next story! :-)

PR #4068, opened 2016-06-24
Refactor tip migration

whit537:
&#x2B11; #3994

&larr; #4079 ([diff](https://github.com/gratipay/gratipay.com/compare/splinter...refactor-tip-migration)) — #4070 &rarr;

I want to be able to document some attributes on the Team class, but am running afoul of https://github.com/sphinx-doc/sphinx/issues/1641. This works around the bug by moving the `print` function reference out into a different file. 

- [x] rebase once #4037 lands.
- [x] rebase once #4079 lands

PR #4083, opened 2016-07-14
Fix imports in tip migration script

rohitpaulk:


whit537:
A regression from #4068.

!m @rohitpaulk

PR #4077, opened 2016-07-07
bring back the takes UI

whit537:
&#x2B11; #3994

&larr; #4099 ([diff](https://github.com/gratipay/gratipay.com/compare/update-some-python-APIs...preen-takes))

Refactor the new takes/membership code now that #4023 and #4024 are merged into this new line of development.

- [x] rebase once #4076 lands
- [x] don't call `get_take_for` from `set_take_for` (use `old_takes`)
- [x] deduplicate error handling between python and simplates
- [x] rebase once #4090 lands
- [x] prevent owner from joining a Team (new PR?)
- [x] factor out a few other PRs ... #4092 #4093 #4094 #4099
- [x] rebase once #4099 lands
- [x] restrict the distributing page—https://github.com/gratipay/gratipay.com/pull/4077#discussion_r73420701 —done in d39d41101066feb85dc88a38f53eafc88f1c392c
- [x] _don't_ restrict the distributing page :) —https://github.com/gratipay/gratipay.com/pull/4077#discussion_r73428674 —reverted in 1733e21a7dd15f9158d088edcb705927ea75dda9
- [x] back out the `lookup.json` l10n—done in 505c14f34ece459eabc3cb759ead21062205209e
- [x] back out the admin bar l10n— https://github.com/gratipay/gratipay.com/pull/4077#discussion_r73401849 —done in 064b88c70463a28fc3457ea67c1e7b4760099b0e
- [ ] cram the table into the layout—https://github.com/gratipay/gratipay.com/pull/4077#discussion_r73421480
- [ ] clean up sidebar—https://github.com/gratipay/gratipay.com/pull/4077#discussion_r73586902
- [x] ~~nicely~~ prevent owners from adding ineligible members—https://github.com/gratipay/gratipay.com/pull/4077#discussion_r73422575 —done in c0cf218f2027c820c5cb6ee35fc3a93c96ee0bc6 and 51d9739340f18a62ca4be4cb5a8edcaf5886839d
- [x] consider fixing test helper for Chrome 52—https://github.com/gratipay/gratipay.com/pull/4077#discussion_r73425567 —done in 1866b966673278a3f1802f0938bdfa5117bf8da8 2be3a87a7feeebbc08b76bd99bdc7e828cc7c91a aa5a093fa4f5f31885cdde83e46490e7f69ad0ec and 68ac0989fae54f8f828dfbffdcc65ebb45407c60

PR #4090, opened 2016-07-20
Respect non-searchable users choice when doing a lookup

nashe:
This aims to fix the report [H1-118699](https://hackerone.com/reports/118699): don't search using LIKE on the non-searchable users, but only display them in the list if it's an exact match.

Some tests were removed since I'm now testing searchable users with/without an exact match and non-searchable users with/without an exact match.


nashe:
Travis is greeting me with a nice:
> The repository at gratipay/gratipay.com was not found.

And their sign-in button just freezes. I'll wait a little bit but I think everything's fine anyway.

whit537:
Golly! I think we saw that behavior a while ago. I forget where, not finding it in search atm ... maybe @rohitpaulk remembers?

nashe:
@whit537 It's finally working on my side, sounds like just temporary issues.

whit537:
Perhaps related to https://github.com/gratipay/gratipay.com/pull/4077#issuecomment-234244581?

nashe:
I guess, yes.

whit537:
@nashe I'm good to merge if you're good with 75aac2c9fe96ae68ac99150f805575fc90243cb8.

nashe:
It's obviously more readable now 👍

whit537:
Deploying ...

whit537:
Deployed! :-)

nashe:
Thanks for your rectification!

PR #4091, opened 2016-07-20
Add the envtest Makefile rule

nashe:
Due to Makefile's ~~age~~ ~~rigidity~~ behaviour, we don't have any control of the CLI arguments. Instead, we have to use `make var=foo rule` to give `foo` to `$(var)` inside the rule.

Example of usage: `make tests="-k test_lookup" envtest`.


whit537:
lgtm 👍

Ready to merge when Travis is ...

whit537:
Oops. Out of sync ... can we rebase?

PR #4098, opened 2016-07-21
Fix usernames restriction

nashe:
This aims to fix [H1-128121](https://hackerone.com/reports/128121). 

We're currently listing all the existing files inside `www_root` and comparing the username to each file/folder to ensure an user won't try to use it. However, there is a mapping between the files and the website route (`lookup.json.spt` is served as`lookup.json`), so the comparison is little bit more complicated.

I'm just adding a list of forbidden extensions, and it will to to match the list of files with `username`, `username.txt`, ... It won't really affect performances since it's only called when creating an account or changing the username. I would really know if somebody has a more elegant way to do it 🐱


nashe:
Hum I was definitely too sleepy, fixing this.

nashe:
Closing and making the right commit to another branch, sorry.

PR #4100, opened 2016-07-23
Fix usernames restriction

nashe:
This aims to fix [H1-128121](https://hackerone.com/reports/128121). 

We're currently listing all the existing files inside `www_root` and comparing the username to each file/folder to ensure an user won't try to use it. However, there is a mapping between the files and the website route (`lookup.json.spt` is served as`lookup.json`), so the comparison is little bit more complicated.

I'm just adding a list of forbidden extensions, and it will to to match the list of files with `username`, `username.txt`, ... It won't really affect performances since it's only called when creating an account or changing the username. I would really know if somebody has a more elegant way to do it 🐱

### DIFF
--- a/gratipay/models/participant/__init__.py
+++ b/gratipay/models/participant/__init__.py
@@ -1137,8 +1137,8 @@ class Participant(Model, mixins.Identity):
 
         # Don't allow any username which is the name of a
         # file existing on the web_root folder.
-        for ext in ['', '.spt', '.ico', '.txt', '.osdd']:
-            if (lowercased + ext) in gratipay.RESTRICTED_USERNAMES:
+        for name in (lowercased, lowercased + '.spt'):
+            if name in gratipay.RESTRICTED_USERNAMES:
                 raise UsernameIsRestricted(suggested)
 
         if suggested != self.username:
--- a/gratipay/models/participant/__init__.py
+++ b/gratipay/models/participant/__init__.py
@@ -1135,8 +1135,11 @@ class Participant(Model, mixins.Identity):
 
         lowercased = suggested.lower()
 
-        if lowercased in gratipay.RESTRICTED_USERNAMES:
-            raise UsernameIsRestricted(suggested)
+        # Don't allow any username which is the name of a
+        # file existing on the web_root folder.
+        for ext in ['', '.spt', '.ico', '.txt', '.osdd']:
+            if (lowercased + ext) in gratipay.RESTRICTED_USERNAMES:
+                raise UsernameIsRestricted(suggested)
 
         if suggested != self.username:
             try:
--- a/tests/py/test_participant.py
+++ b/tests/py/test_participant.py
@@ -1,6 +1,7 @@
 from __future__ import print_function, unicode_literals
 
 import datetime
+import os
 import random
 
 import mock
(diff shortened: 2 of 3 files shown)


nashe:
@whit537 : do we need something else to be added to this PR?

whit537:
Let's see ...

whit537:
Rebased on master, previous head was 6f054cc658afdb98928d8bf30bc4685db62b5a2c.

whit537:
I want to go over this to make sure our logic precisely matches what Aspen does ...

nashe:
Sure!

gratipay/models/participant/__init__.py
```diff
@@ -1135,8 +1135,11 @@ def change_username(self, suggested):
 
         lowercased = suggested.lower()
 
-        if lowercased in gratipay.RESTRICTED_USERNAMES:
-            raise UsernameIsRestricted(suggested)
+        # Don't allow any username which is the name of a
+        # file existing on the web_root folder.
+        for ext in ['', '.spt', '.ico', '.txt', '.osdd']:
+            if (lowercased + ext) in gratipay.RESTRICTED_USERNAMES:
+                raise UsernameIsRestricted(suggested)
```
whit537: What if we took a different approach entirely, and actually ran `/newusername` through the Aspen request processor? That would match the condition more exactly, and would be more future proof.

whit537: A redirect (to either `/username/` or `/~username`) would indicate that the username is available. A 200 would indicate that it is not.

whit537: That's basically an internal redirect, which we don't have good facilities for in Aspen (https://github.com/AspenWeb/pando.py/issues/143). The test harness is the closest we have, and we don't want to import all that machinery into production. Hmm ...

whit537: Actually, all we care about is dispatch. Maybe we can isolate that?

whit537: Hrm, not easily. Dispatch is embedded in Gratipay's [algorithm](https://github.com/gratipay/gratipay.com/blob/cecb6ce265c7dba325cf4561a5acef948feb727e/gratipay/main.py#L97-L138), and it seems that we'd want to test the real algorithm.

whit537:
I'm trying out an approach based on an internal redirect ...

whit537:
Now I'm giving up on that approach and suggesting 5592272b03cd48fc239725cef6deae89cb014979.

Why do we need to check `.ico`, `.txt`, and `.osdd`? Yes, those file extensions appear in the directory listing, but a username of `robots` won't conflict with `robots.txt`, will it?

nashe:
I completely agree with all of your remarks, my change was somehow clumsy. I need to try to improve my PRs in the future!

(wow, great reactivity of Github concerning the link issue!)

whit537:
Rebased on master. Previous head was d2a25a3633cd33a5653ca0422573d43723742987.

whit537:
@nashe Cool, no worries. PRs are better than no PRs! :)

I'm ready to merge when Travis is green ...

whit537:
"PR early, PR often!" 💃

PR #4103, opened 2016-07-29
Prefix the search term to mitigate text injection

nashe:
This aims to fix [H1-154921](https://hackerone.com/reports/154921).

### DIFF
--- a/www/search.spt
+++ b/www/search.spt
@@ -10,7 +10,7 @@ banner = _("Search")
 results = {}
 
 if query:
-    title = query
+    title = "Results for: " + query 
     q = strip_accents(query)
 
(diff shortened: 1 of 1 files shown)


www/search.spt
```diff
@@ -10,7 +10,7 @@ banner = _("Search")
 results = {}
 
 if query:
-    title = query
+    title = "Results for: " + query
```
whit537: i18n? :)

nashe: Obviously…

whit537:
👍

screen shot 2016-08-03 at 11 38 46 am

screen shot 2016-08-03 at 11 38 28 am

PR #4105, opened 2016-08-08
Prevent content spoofing on "/on/%elsewhere/%account"

ghost:
One old content spoofing was [already addressed on March 2015](https://github.com/gratipay/gratipay.com/pull/2982). Since [this commit](https://github.com/gratipay/gratipay.com/pull/2982/commits/4d7d0dc24302e3209a70cd350269f677571f147b#diff-2ddfd667e590b3912a22e6c01236150fR12), the username is being validated. One 404 error may be returned during this process. 

The reported content injection occurs right there. Its location changed accross the commits and it's now in [`models/account_elsewhere.py`](https://github.com/gratipay/gratipay.com/blob/2e82f49de2ddcb2fd2ffe74bc696e10bf6c8ea24/gratipay/models/account_elsewhere.py#L272).

We can't just add `text-overflow: ellipsis` to the error template, since it would break legit error pages. Truncating the username if it's more than X characters  looks like the best choice: the search functionality has the same behaviour (using `text-overflow`).

My suggestion is the following:
- Don't accept `\t`, `\r`, `\n` characters in the account name, since it can "break" our display.
- If the username's length is less than 15 characters, display it...
- ...else truncate it and append "...".

@whit537: I have some troubles to add a test for the truncation of the username. VCR saved a request for the requests associated with "adhsjakdjsdkjsajdhksda" but I have troubles to add one for my new test case. Making requests will always raise `CannotOverwriteExistingCassetteException: No match for the request <my-request> was found`.

Travis will fail.

### DIFF
--- a/gratipay/models/account_elsewhere.py
+++ b/gratipay/models/account_elsewhere.py
@@ -250,11 +250,16 @@ class AccountElsewhere(Model):
 
 
 def get_account_elsewhere(website, state, api_lookup=True):
+    _ = state['_']
     path = state['request'].line.uri.path
+
     platform = getattr(website.platforms, path['platform'], None)
     if platform is None:
         raise Response(404)
+
     uid = path['user_name']
+    if "\t" in uid or "\r" in uid or "\n" in uid:
+        raise Response(400, _("Invalid character in elsewhere account username."))
     if uid[:1] == '~':
         key = 'user_id'
         uid = uid[1:]
@@ -271,7 +276,8 @@ def get_account_elsewhere(website, state, api_lookup=True):
             user_info = platform.get_user_info(key, uid)
         except Response as r:
             if r.code == 404:
-                _ = state['_']
+                # Prevent Content Spoofing by truncating long usernames
+                uid = (uid[:15] + '...') if len(uid) > 15 else uid
                 err = _("There doesn't seem to be a user named {0} on {1}.",
                         uid, platform.display_name)
                 raise Response(404, err)
--- a/gratipay/models/account_elsewhere.py
+++ b/gratipay/models/account_elsewhere.py
@@ -276,10 +276,8 @@ def get_account_elsewhere(website, state, api_lookup=True):
             user_info = platform.get_user_info(key, uid)
         except Response as r:
             if r.code == 404:
-                # Prevent Content Spoofing by truncating long usernames
-                uid = (uid[:15] + '...') if len(uid) > 15 else uid
-                err = _("There doesn't seem to be a user named {0} on {1}.",
-                        uid, platform.display_name)
+                err = _("Impossible to find one account with this name on {0}.",
+                        platform.display_name)
                 raise Response(404, err)
             raise
         account = AccountElsewhere.upsert(user_info)
--- a/gratipay/models/account_elsewhere.py
+++ b/gratipay/models/account_elsewhere.py
@@ -276,8 +276,7 @@ def get_account_elsewhere(website, state, api_lookup=True):
             user_info = platform.get_user_info(key, uid)
         except Response as r:
             if r.code == 404:
-                err = _("Impossible to find one account with this name on {0}.",
-                        platform.display_name)
+                err = _("Account not found on {0}.", platform.display_name)
                 raise Response(404, err)
             raise
         account = AccountElsewhere.upsert(user_info)
--- a/tests/py/test_elsewhere.py
+++ b/tests/py/test_elsewhere.py
@@ -120,6 +120,12 @@ class TestElsewhere(Harness):
             expected = error % (user_name, platform.display_name)
             assert expected in r.body
 
+    def test_user_pages_with_invalid_characters(self):
+        platforms = [p for p in self.platforms]
+        assert self.client.GxT('/on/'+platforms[0].name+'/AA%0DBB')
+        assert self.client.GxT('/on/'+platforms[0].name+'/AA%0ABB')
+        assert self.client.GxT('/on/'+platforms[0].name+'/AA   BB')
+
     def test_failure_page_accepts_valid_username(self):
         self.client.GET('/on/twitter/Gratipay/')  # normal case will have the db primed
         response = self.client.GET('/on/twitter/Gratipay/failure.html')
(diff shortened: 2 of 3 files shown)


whit537:
@nashe Try deleting the relevant `*.yml` file in `tests/py/fixtures` and rerunning the tests. It should generate a new yml file.

nashe:
@whit537 Sounds aggressive :> I'll try this, thanks!

whit537:
@nashe I believe we explicitly tell VCR _not_ to overwrite cassettes, so that we don't accidentally change them without meaning to.

nashe:
That's what I ~thought while reading [some documentation](https://relishapp.com/vcr/vcr/v/1-9-0/docs/record-modes/once) but I missed the second line (`Record new interactions if there is no cassette file`).

nashe:
Forgot to mark it ready for review.

whit537:
4edbaf7fcee7593dd7f1b317ec21561160fe5a02 is the commit I'm thinking of (though GitHub is linking to a weird PR, in a fork it seems; I've emailed support about that).

@aandis Any chance you could spare a few cycles to help @nashe with this one? :-)

nashe:
cf. my first message,

> We can't just add text-overflow: ellipsis to the error template, since it would break legit error pages. Truncating the username if it's more than X characters looks like the best choice: the search functionality has the same behaviour (using text-overflow).

Maybe there is another solution in CSS (putting the arbitrary value in a fixed-width span and applying `text-overflow` to this element only...) but it's rather ugly and I don't see any improvement rather than a "hard" limit :(

gratipay/models/account_elsewhere.py
```diff
@@ -271,7 +276,8 @@ def get_account_elsewhere(website, state, api_lookup=True):
             user_info = platform.get_user_info(key, uid)
         except Response as r:
             if r.code == 404:
-                _ = state['_']
+                # Prevent Content Spoofing by truncating long usernames
+                uid = (uid[:15] + '...') if len(uid) > 15 else uid
```
whit537: Should be `:12` here if we're adding three characters. Could also try unicode ellipsis, which is only one character. ;-)

whit537:
> cf. my first message,

Gah, sorry for not reading for carefully. Okay, I'm good to go on this if we can get a rebase on master.

aandis:
> We can't just add text-overflow: ellipsis to the error template, since it would break legit error pages. Truncating the username if it's more than X characters looks like the best choice: the search functionality has the same behaviour (using text-overflow).

Agree with @nashe on this.

gratipay/models/account_elsewhere.py
```diff
@@ -250,11 +250,16 @@ def save_token(self, token):
...
+    _ = state['_']
     path = state['request'].line.uri.path
+
     platform = getattr(website.platforms, path['platform'], None)
     if platform is None:
         raise Response(404)
+
     uid = path['user_name']
+    if "\t" in uid or "\r" in uid or "\n" in uid:
+        raise Response(400, _("Invalid character in elsewhere account username."))
```
aandis: Characters would look cleaner here ;)
Btw when do we hit this case and why do we need specifically handle it?

aandis:
We can however get rid of the long error message all together to simplify things. A simple `404 Not Found` should also suffice.

gratipay/models/account_elsewhere.py
nashe: In the PoC provided on HackerOne (I can't paste it in this commence, their website is in maintenance…), the researcher showed that using these characters may help to make new paragraphs and make the phishing look more real. 

I very reluctant to handle the reports related to content spoofing (insignificant risk since we allow arbitrary content on profiles & co—wasn't it created just to grab money on bug bounties? :-)), so I try to be the more restrictive I can on the input to avoid getting new reports showing a phishing message showing a new paragraph and 10 chars of "malicious message" inside.

In addition, I don't see any reason to consider those characters as valid anyway. Do you know one elsewhere allowing one of them?

nashe:
I would prefer at least "Impossible to find this user on PLATFORM". 

We need to take a decision: finish this PR or open another one where we'll just redact the username from the error / the whole error message?

gratipay/models/account_elsewhere.py
nashe: > Characters would look cleaner here ;)

We only need one match (== character) to raise this 400 so… technically I would not put a "s" :-(

gratipay/models/account_elsewhere.py
nashe: Noted. It'll depend of the decision taken after [my last message](https://github.com/gratipay/gratipay.com/pull/4105#issuecomment-241911165)!

nashe:
No response, so I'll commit the "Impossible to find this user on PLATFORM" and wait for somebody to merge—the report is already 6 months old and this PR 20 days old 😞

whit537:
Rebased on master. I'm going to tweak the error message and then merge.

tests/py/test_elsewhere.py
```diff
@@ -110,15 +110,21 @@ def test_user_pages(self, get_user_info):
...
-        user_name = 'adhsjakdjsdkjsajdhksda'
-        error = "There doesn't seem to be a user named %s on %s."
+    def test_user_pages_are_404_for_unknown_elsewhere_user(self):
+        error = "Account not found on %s."
         for platform in self.platforms:
             if not hasattr(platform, 'api_user_name_info_path'):
                 continue
-            r = self.client.GxT("/on/%s/%s/" % (platform.name, user_name))
-            expected = error % (user_name, platform.display_name)
+            r = self.client.GxT("/on/%s/%s/" % (platform.name, 'ijroioifeef'))
```
whit537: Hope nobody ever registers this name! :-P

whit537:
Alright, I think I'm ready to go on this once Travis is green. You good with the latest commits, @nashe?

nashe:
!m @whit537

nashe:
Alright, closing the H1 report. 

!m everybody!

nashe:
@whit537: are you OK with theoretical risk?

whit537:
@nashe Yes. [Looks like](http://inside.gratipay.com/appendices/disclosures) we've been inconsistent in the past with spoofing.

mild: https://hackerone.com/reports/115284
theoretical: https://hackerone.com/reports/154921

But the more recent was theoretical, and I agree with your [assessment](https://github.com/gratipay/gratipay.com/pull/4105#discussion_r75969526) that spoofing reports are low-quality.

nashe:
Done, thanks @whit537 👍

PR #4106, opened 2016-08-19
Limit user statement size

nashe:
This aims to close [H1-127824](https://hackerone.com/reports/127824). When parsing very large user statements, the markdown parser may take too much resources. We are already safe against nested elements attacks.

**Questions**: 
- should we restrict the "product or project" field for teams? The teams are manually reviewed and they don't have any interest in voluntarily slowing down the display of their pages.
- should we add one hard limit on the database field? It's hard to predict, since we (want to) restrict the size before the md-to-html translation process.

### DIFF
--- a/www/~/%username/statement.json.spt
+++ b/www/~/%username/statement.json.spt
@@ -17,7 +17,7 @@ if request.method == 'POST':
 
     # Statements are limited to 256kB to avoid DoS'ing the markdown parser.
     if len(content) > (256 * 1024):
-        raise Response(400, "Statement too big!")
+        raise Response(400, _("Statement too big!"))
 
     participant.upsert_statement(lang, content)
     r = {"html": markdown.render(content)}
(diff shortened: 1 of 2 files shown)


www/~/%username/statement.json.spt
```diff
@@ -15,6 +15,10 @@ if request.method == 'POST':
     if lang not in LANGUAGES_2:
         raise Response(400, "unknown lang")
 
+    # Statements are limited to 256kB to avoid DoS'ing the markdown parser.
+    if len(content) > (256 * 1024):
+        raise Response(400, "Statement too big!")
```
whit537: Since this can show up in the UI, we should i18n it (the "unknown lang" just before should only be hittable from a script).

screen shot 2016-08-22 at 11 26 05 pm

whit537: Done in 671ff18a387d30a361eb15bff603ef4b8a7b01d1.

whit537:
I'm good to merge here once Travis is green again.

nashe:
🍏

PR #4108, opened 2016-08-24
Remove gendered pronoun

whit537:
We should avoid gender-specific pronouns in non-specific contexts.


nashe:
It seems to be the only occurrence of "his", look good to me :-)

PR #4109, opened 2016-08-24
Refactor take_over tests

whit537:
The participant test file was starting to get big, and this was an obvious thing to factor out.


nashe:
Last bonus PR, looks good to me 👍 
Rebasing…

Issue #3846, opened 2015-10-23
add `noindex` to rejected Teams


whit537:
Reticketed from https://github.com/gratipay/team-review/issues/54.

whit537:
I'm taking off the Embarrassing label since this is on a milestone.

kaguillera:
Please note that the PR for this is up for review anyone...

whit537:
Decided against this at https://github.com/gratipay/gratipay.com/pull/4125#issuecomment-255552348.

PR #4125, opened 2016-09-07
Added 'noindex' and 'nofollow' in the meta tag

kaguillera:
PR for #3846. Not sure if/what test is required for this so I did not put one. If anyone thinks that it is necessary let me know.

### DIFF
--- a/www/%team/index.html.spt
+++ b/www/%team/index.html.spt
@@ -20,7 +20,6 @@ is_team_owner = not user.ANON and team.owner == user.participant.username
           href="https://gratipay.com/{{ team.slug }}/" />
     <meta name="ROBOTS" content="NOINDEX, NOFOLLOW" />
     <meta name="twitter:card" content="summary" />
-    <meta name="og:image" content="{{ website.asset('gratipay.opengraph.png') }}" />
     <meta name="og:url" content="https://gratipay.com/{{ team.slug }}/" />
     <meta name="og:type" content="profile" />
     <meta name="og:title" content="{{ team.name }}" />
--- a/www/%team/index.html.spt
+++ b/www/%team/index.html.spt
@@ -18,9 +18,7 @@ is_team_owner = not user.ANON and team.owner == user.participant.username
 {% block head %}
     <link rel="payment" type="text/html" title="Give weekly gift on Gratipay"
           href="https://gratipay.com/{{ team.slug }}/" />
-    {% if team.status == 'rejected' %}
     <meta name="ROBOTS" content="NOINDEX, NOFOLLOW" />
-    {% endif %}
     <meta name="twitter:card" content="summary" />
     <meta name="og:image" content="{{ website.asset('gratipay.opengraph.png') }}" />
     <meta name="og:url" content="https://gratipay.com/{{ team.slug }}/" />
--- a/www/%team/index.html.spt
+++ b/www/%team/index.html.spt
@@ -18,7 +18,9 @@ is_team_owner = not user.ANON and team.owner == user.participant.username
 {% block head %}
     <link rel="payment" type="text/html" title="Give weekly gift on Gratipay"
           href="https://gratipay.com/{{ team.slug }}/" />
+    {% if team.status == 'rejected' %}
     <meta name="ROBOTS" content="NOINDEX, NOFOLLOW" />
+    {% endif %}
     <meta name="twitter:card" content="summary" />
     <meta name="og:image" content="{{ website.asset('gratipay.opengraph.png') }}" />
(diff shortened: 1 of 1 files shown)


mattbk:
Does this apply only to rejected teams?

mattbk:
Travis didn't like that. My question is whether something like 
```
    {% if team.status == 'rejected' %}
    <meta name="ROBOTS" content="NOINDEX, NOFOLLOW" />
    {% endif %}
```

is needed.

www/%team/index.html.spt
```diff
@@ -18,7 +18,9 @@ is_team_owner = not user.ANON and team.owner == user.participant.username
 {% block head %}
     <link rel="payment" type="text/html" title="Give weekly gift on Gratipay"
           href="https://gratipay.com/{{ team.slug }}/" />
+    <meta name="ROBOTS" content="NOINDEX, NOFOLLOW" />
     <meta name="twitter:card" content="summary" />
+    <meta name="og:image" content="{{ website.asset('gratipay.opengraph.png') }}" />
```
aandis: This is already declared [here](https://github.com/gratipay/gratipay.com/pull/4125/files#diff-54fc7e025b41a7185c3ff65e3111b6f8R28).

www/%team/index.html.spt
```diff
@@ -18,7 +18,9 @@ is_team_owner = not user.ANON and team.owner == user.participant.username
 {% block head %}
     <link rel="payment" type="text/html" title="Give weekly gift on Gratipay"
           href="https://gratipay.com/{{ team.slug }}/" />
+    <meta name="ROBOTS" content="NOINDEX, NOFOLLOW" />
```
aandis: We should only do this for rejected teams as pointed out by @mattbk.

aandis:
I like the new review style.

www/%team/index.html.spt
whit537: What about pending teams? We probably don't want those indexed either, now that I think about it.

aandis: Aren't we no indexing too much then? ;)

aandis: More to the point, I don't see why we should do this at all. What problem do you see with someone googling and coming on to a rejected/pending team?

whit537: We'd only index approved teams.

> What problem do you see with someone googling and coming on to a rejected/pending team page?

It can make the team look bad: https://github.com/gratipay/team-review/issues/54.

www/%team/index.html.spt
kaguillera: Sorry that is a mistake...will fix it now

www/%team/index.html.spt
kaguillera: So what is the final decision. Should it be `NOINDEX` or not and should pending be `NOINDEX` also?

whit537: @aandis Do you buy the reasoning? Or do you still think we should leave rejected teams searchable?

whit537: I mean, it's not very open and transparent to only leave the "good stuff" searchable and try to suppress the "bad information."

mattbk: Under the [right to be forgotten](https://en.wikipedia.org/wiki/Right_to_be_forgotten), ~~teams~~ projects should have control over their own information. We leave it up as an example, which I assume falls under the TOS. An option might be to (somehow) anonymize rejected ~~teams~~ projects...but this seems like an easier solution unless someone yells. It's also possible for the owner to edit most of the information, although the `slug` remains.

On this note, are we going to re-review ~~teams~~ projects that were previously rejected once [Decouple](https://github.com/gratipay/gratipay.com/milestone/22) is complete?

aandis: Sorry, I don't agree here. imo rejected a team just means it wasn't compatible with gratipay and what we do here. Nowhere do we reflect on how good the team is or the nature of work they do. How people perceive that rejection is really upto them. Letting people find rejected teams will at least give them the idea about what we do here and what we don't, if nothing else.

aandis: But if you still want to go ahead with this, I'd be fine with that too. Maybe I am the one who doesn't get the clear picture here. ;)

mattbk: *shrug* I don't care either way, it's a policy @whit537 came up with.

whit537:
Alright, rejecting this PR because a) we don't have a strong consensus that we want it, b) it's not done, and c) it's in the way. :-)

kaguillera:
Actually it was finished but I was waiting for the final decision before I put it up for review.

PR #4134, opened 2016-10-04
Fix typo in German translation

pvorb:


whit537:
Thanks, @pvorb! I've merged and deployed the change. Can you confirm whether it worked? If not we may have to [go through Transifex](http://inside.gratipay.com/howto/translate-gratipay).

pvorb:
I can't see the change. It still says "Edieren".

whit537:
Yeah, looks like your change was reverted in 13f9d8b0b77fb9c6653f4380d630c64736f5226e, meaning that Transifex is the source, and our po files are derived from there. I've fixed this in Transifex and redeployed. Mind checking one (hopefully) more time? :)

pvorb:
The change went through.

image

whit537:
Cool. Thank you! :-)

PR #4151, opened 2016-10-20
Implement /on/npm/foo pages

whit537:
Part of #4148.

# Todo

- [x] squish double-header
- [x] hide anchor links icon on headings
- [x] syntax highlight code snippets
- [x] fix tests
- [x] attend to h3-h6
- [x] do something with images
- [ ] see about the broken img src on http://localhost:8537/on/npm/react-router/

# Test

From https://github.com/gratipay/gratipay.com/pull/4135#issuecomment-262672635:

http://localhost:8537/on/npm/async/
http://localhost:8537/on/npm/iframe-resizer/
http://localhost:8537/on/npm/mongoose/
http://localhost:8537/on/npm/nodemon/
http://localhost:8537/on/npm/react/
http://localhost:8537/on/npm/react-helmet/
http://localhost:8537/on/npm/react-modal/
http://localhost:8537/on/npm/react-redux/
http://localhost:8537/on/npm/react-router/
http://localhost:8537/on/npm/react-router-redux/
http://localhost:8537/on/npm/redux/
http://localhost:8537/on/npm/redux-thunk/
http://localhost:8537/on/npm/webpack/

------------------------

Whee! 💃 

screen shot 2016-10-20 at 3 23 57 pm

Issue #4148, opened 2016-10-20
Integrate npm


whit537:
-------------------------------

✈️  **This is the flight deck for the [Integrate npm](https://github.com/orgs/gratipay/projects/5?fullscreen=true.) project.** ✈️ 

-------------------------------

Current open-source crowdfunding options (Kickstarter, Patreon, Gratipay, OpenCollective, etc.) are consumer-grade. Our hunch is that a business-grade product with better aggregation can better serve the companies that want to pay for open source, because companies use hundreds or thousands of open source packages, not just a few.

Picking up from https://github.com/gratipay/gratipay.com/pull/4135#issuecomment-255122149 and https://github.com/gratipay/inside.gratipay.com/issues/852#issuecomment-255098337 ...

For wider context see:

- [The fundamental idea is to get a viral loop going again like we had under Gittipay 1.0.](https://github.com/gratipay/inside.gratipay.com/issues/637#issuecomment-263431528)
- [The burn I feel](http://gratipay.slackarchive.io/gratipay/-/1480699124.000841/1480708926.000944/1480708501000910/)

JavaScript is the most popular language in open source and npm is the most popular package manager for JavaScript. A good first concrete step towards helping companies pay for open source (#4135), therefore, will be to add the ability to pay for any package on npm. Once we have npm deployed, we will have enough experience to inform a partnership with Libraries.io for the rest of the package managers.

### Target

~~Our goal is to announce this feature in my [lightning talk](https://github.com/whit537/openorg/issues/6) on Thursday, October 26 at Red Hat's All Things Open conference (https://github.com/gratipay/inside.gratipay.com/issues/757).~~

Our goal is to incrementally improve this feature throughout the first half of 2017, with an eye towards [OSCON](https://github.com/gratipay/inside.gratipay.com/issues/948) and [$ustain](https://github.com/gratipay/inside.gratipay.com/issues/920) in May.

### Package names to test with

From https://github.com/gratipay/gratipay.com/pull/4135#issuecomment-262672635:

http://localhost:8537/on/npm/async/
http://localhost:8537/on/npm/iframe-resizer/
http://localhost:8537/on/npm/mongoose/
http://localhost:8537/on/npm/nodemon/
http://localhost:8537/on/npm/react/
http://localhost:8537/on/npm/react-helmet/
http://localhost:8537/on/npm/react-modal/
http://localhost:8537/on/npm/react-redux/
http://localhost:8537/on/npm/react-router/
[...]

PR #4117, opened 2016-08-30
Edit TOS to decouple payins and payouts

mattbk:
https://github.com/gratipay/inside.gratipay.com/issues/432#issuecomment-243522286

> I've started a [private ticket](https://github.com/gratipay/legal-tor-ekeland/issues/31) with our [lawyer](https://github.com/copiesofcopies) about revising our Terms of Service to relax our open work requirement. In the interest of time/money, I think the way to proceed is **take a stab at revising them ourselves**, and then bring our revision to him for review.

Stub out a PR for changes.

Related milestones:
- IG: https://github.com/gratipay/inside.gratipay.com/milestone/9
- G: https://github.com/gratipay/gratipay.com/milestone/22
### Todo
- [x] come up with an alias for Participants acting as contributors, so that we're never using "Participant" in a context-dependent way
  - Tried to address in https://github.com/gratipay/gratipay.com/pull/4117/commits/78edd7983ec2ca316db9b5fd0cb5212e763d8822 by creating "Giver"
- [x] more clearly describe the differences between these various roles up from (and clarify that each is a Participant interacting with the platform in a different way)
  - Tried to address in https://github.com/gratipay/gratipay.com/pull/4117/commits/78edd7983ec2ca316db9b5fd0cb5212e763d8822
- [x] revise the individual sections (e.g. the new 3-5) such that each is limited to describing the obligations of one role. So one section for Owners, one for Participant-contributors, and one for Members
  - Tried to address in https://github.com/gratipay/gratipay.com/pull/4117/commits/78edd7983ec2ca316db9b5fd0cb5212e763d8822
- [x] rename "Owners" and "Members" to "Project Owners" and "Project Members." "Members" in particular is ambiguous -- since "members," "participants," and "users" could all potentially be used interchangeably in common speech, people may get confused trying to keep them straight when reading through the terms.
  - Tried to address in  https://github.com/gratipay/gratipay.com/pull/4117/commits/01e8cb2aa068706354023a7d93e0d74a7f7f8e2e
- [x] Probably use different language for the "Work" performed by the project and the work performed by Members for Projects 
  - Tried to address in  https://github.com/gratipay/gratipay.com/pull/4117/commits/01e8cb2aa068706354023a7d93e0d74a7f7f8e2e
- [x] Run changes by @tieguy.
- [ ] Propagate final version to `terms-of-service.spt`.
- [ ] Email new terms and post as a news item. (https://github.com/gratipay/gratipay.com/pull/4117#issuecomment-262336157)

PR #4153, opened 2016-10-22
Load up npm

whit537:
We need to load npm into our database and keep it up to date, in order to enable the `/on/npm/`, `/on/npm/foo/`, and `/search` changes we want to make.


aandis:
This needs to have the work done in #4155 so I'm just gonna grab it from there.

whit537:
@aandis Yeah, and we can drop the `/on/npm/` stuff from this. Probably best to rebase here so we have a cleaner history. Are you comfortable with rebasing? :-)

aandis:
Rebased to remove `/on/npm` work. Previous head was 4d92a2293a6fb7ce63c677fbba7659bf1047a1eb

aandis:
Rebased again on top of #4155 to bring in all work from there. Commits show on https://github.com/gratipay/gratipay.com/pull/4153/commits but not here. weird.

aandis:
Actually let's scratch that rebase. I wanted to use [Package.insert](https://github.com/gratipay/gratipay.com/pull/4155/files#diff-75620ed158b41ddbc9f51b2242527c29R48)  but we'll need a multi insert statement here.

aandis:
@whit537 how do I run the chomp script?

whit537:
> how do I run the chomp script?

@aandis Once you `pip install -e .` (as under `make env`) you should have it on your `PATH` via an `entrypoint` in `setup.py`. We should write tests for all the pieces that go into it, though. I tried to factor it so it's easy to swap out the HTTP request for `all` to test with a much smaller subset of the full JSON.

aandis:
Got it. `/env/bin/chomp`

whit537:
>  I wanted to use `Package.insert` but we'll need a multi insert statement here.

We may need to think differently about initial load and subsequent updates. For initial load we'll have _a lot_ to do and it might be worth looking at [`COPY`](https://www.postgresql.org/docs/9.4/static/populate.html). Remember that for subsequent runs we'll need to `update` as well as `insert`. [Sounds like](http://blog.npmjs.org/post/141905368000/changes-to-npms-unpublish-policy) we may not need to worry about `delete` (once a name is claimed, even if unpublished it gets a security package).

whit537:
I had figured we could pass `-1` as the arg to chomp to mean "reprocess *everything*".

Fetching and processing readmes is going to add significant expense. It may make sense to decouple that and just load the database with `all` and *then* process readmes.

aandis:
> it might be worth looking at COPY

I thought `COPY` could only be used with files. looking into it.

whit537:
So maybe we have these three?

1. A script that takes `all` and spits out whatever `COPY` wants. We run that manually to initially populate the table (will be able to use that same `COPY` method when adding packages from other package managers like RubyGems and PyPI? Or does that only work for populating new/empty tables?).
1. A script that takes `all` and upserts any record with `modified time` newer than the last time the script was run (every 30 minutes?).
1. A script that does `select * from packages where readme is null order by mtime asc limit 1` and then hits the registry api and processes with `markdown.marky`. It runs continuously (either via cron, or with its own eternal loop).

I was thinking (1) and (2) could be variants of the same `chomp` script. Probably (3) is a separate script?

aandis:
> A script that takes all and spits out whatever COPY wants. We run that manually to initially populate the table 

Manual work [shouldn't](http://stackoverflow.com/a/1870002/2643846) be needed.

>> (will we be able to use that same COPY method when adding packages from other package managers like RubyGems and PyPI? Or does that only work for populating new/empty tables?).

> COPY FROM copies data from a file to a table (appending the data to whatever is in the table already)

https://www.postgresql.org/docs/current/static/sql-copy.html

whit537:
> Manual work shouldn't be needed.

Okay. I mostly meant that we can just run that script as a one-off rather than having to install it in a long-lived dyno/cronjob/whatever.

aandis:
> A script that takes all and upserts any record with modified time newer than the last time the script was run (every 30 minutes?).

Do we get `modified_time` for a package inside `all`?

whit537:
> Do we get `modified_time` for a package inside `all`?

[Yes](https://github.com/gratipay/gratipay.com/pull/4151#issuecomment-255460238)!

aandis:
Sounds good. I am gonna start working on 1. Might doze off in an hour. Will pick up tomorrow from where you leave it.

whit537:
Awesome, @aandis, sounds good. I have #4117 just about topped up. I'll jump in here in a bit ...

aandis:
@whit537 built up a bit on 1. Should need to call copy with the data created to insert it. Gonna head to bed now. Will pick it up tomorrow. :)

sql/branch.sql
```diff
@@ -0,0 +1,27 @@
...
+    , long_description_type text                        NOT NULL DEFAULT ''
+    , mtime                 timestamp with time zone    NOT NULL
+    , UNIQUE (package_manager_id, name)
+     );
+
+    CREATE TABLE package_emails
+    ( id            bigserial   PRIMARY KEY
+    , package_id    bigint      NOT NULL REFERENCES packages ON UPDATE CASCADE ON DELETE RESTRICT
+    , email         text        NOT NULL
+    , UNIQUE (package_id, email)
```
whit537: I don't think we want a `UNIQUE` here because a package can have multiple author/maintainer emails, no?

sql/branch.sql
```diff
@@ -0,0 +1,27 @@
+BEGIN;
+
+    CREATE TABLE package_managers
+    ( id        bigserial       PRIMARY KEY
+    , name      text            NOT NULL UNIQUE
+     );
+
+    CREATE TABLE packages
+    ( id                    bigserial                   PRIMARY KEY
+    , package_manager_id    bigint                      NOT NULL REFERENCES package_managers ON UPDATE CASCADE ON DELETE RESTRICT
```
whit537: Pretty sure we don't need the `ON UPDATE ... ON DELETE` since we're referencing an immutable field.

gratipay/utils/markdown.py
```diff
@@ -1,9 +1,13 @@
...
 
 def render(markdown):
     return Markup(m.html(
         markdown,
         extensions=m.EXT_AUTOLINK | m.EXT_STRIKETHROUGH | m.EXT_NO_INTRA_EMPHASIS,
         render_flags=m.HTML_SKIP_HTML | m.HTML_TOC | m.HTML_SMARTYPANTS | m.HTML_SAFELINK
     ))
+
+def marky(markdown):
+    return subprocess.call()
```
whit537: This should get rebased on `master` now that #4154 is in.

whit537:
I'm gonna dive in to this today on the plane.

whit537:
Here's a few commits. More later ...

gratipay/chomp/__init__.py
```diff
@@ -53,6 +53,7 @@ def fetch_catalog(self):
 
 
 def stringify(obj):
+    # XXX What if there is a `\t` in `v`?
```
aandis: Hm. We should probably opt for different [separator](http://initd.org/psycopg/docs/cursor.html#cursor.copy_from) then?

aandis:
@whit537 just a heads up. I did some tests on this today with the full request loaded (which takes quite a while) and although the script formats the data correctly for `copy`, `copy` ends up complaining about special characters in name/description (eg `\r`). So we'll have to get around that too.

whit537:
Cool, thanks.

> (which takes quite a while)

I'm currently looking at [`ijson`](https://pypi.python.org/pypi/ijson) in event-reacting mode. It looks promising.

whit537:
Looks like this version is taking about four minutes to run on my laptop.

It walks the events from `ijson` and emits output serialized for ... something. Either `COPY` or a unified `INSERT`/`UPDATE` is what I'm thinking.

... Yup!

```
processed 369000 packages in 246.22 seconds

real    4m7.086s
user    3m47.453s
sys     0m3.103s
```

whit537:
`ijson` version in b0549d57b5892da969cb445d04981e936b1f9a4e. Requires yajl (`brew install yajl` wfm, ymmv).

whit537:
I think we should try to denormalize emails as well, so that we can load into a single table and not have to muck about with keeping track of inserted ids. That'll also save us having to prune old emails from the emails table. We'll just overwrite with all new data every time based on mtime.

whit537:
Next step is to get this serializing to something that Postgres can do something with. We may be close if we're using `COPY`. Looks like that's going to be about 42 MB for an initial load. I want to think about updates, too, though ...

whit537:
Rebased. Previous head was 99662d0d48ffa63c234cee55f1174944863d3ea1. (More rebasing to come.)

I am tantalizingly close to having an integration test passing for loading npm into the `packages` table! I found a great [technique](http://tapoueh.org/blog/2013/03/15-batch-update.html) that combines `COPY` with a temp table and `with (update) insert`.

I'm still not comfortable with the serialization, though. I'm nervous about injection vectors. Tabs:

https://www.postgresql.org/docs/9.3/static/arrays.html
https://www.postgresql.org/docs/9.3/static/sql-syntax-lexical.html#SQL-SYNTAX-CONSTANTS-GENERIC
http://initd.org/psycopg/docs/cursor.html#cursor.copy_from
http://stackoverflow.com/questions/31519063/cannot-insert-error-array-value-must-start-with-or-dimension-information
http://blog.lerner.co.il/learning-love-postgresql-arrays/

whit537:
Got it!

whit537:
But now I'm back into serialization weirdness ... getting `'testing-package'` as the name instead of `testing-package`

whit537:
Let's get to the bottom of the serialization question. What's the right way to generate the input to `COPY`? Here's the definition of [`copy_from`](https://github.com/psycopg/psycopg2/blob/2_5_1/psycopg/cursor_type.c#L1306-L1409).

whit537:
Here's the `COPY` format string [in use](https://github.com/psycopg/psycopg2/blob/2_5_1/psycopg/cursor_type.c#L1341):

`COPY %s%s FROM stdin WITH DELIMITER AS %s NULL AS %s`

whit537:
[`COPY`](https://www.postgresql.org/docs/9.3/static/sql-copy.html) takes a `FORMAT` option:

> Selects the data format to be read or written: `text`, `csv` (Comma Separated Values), or `binary`. The default is text.

However, psycopg2 doesn't expose it.

whit537:
Here we go: [File Formats > Text Format](https://www.postgresql.org/docs/9.3/static/sql-copy.html#AEN69274).

whit537:
> Currently no adaptation is provided between Python and PostgreSQL types on COPY: the file can be any Python file-like object but its format must be in the format accepted by [PostgreSQL COPY command](http://www.postgresql.org/docs/current/static/sql-copy.html) (data fromat, escaped characters, etc).

http://initd.org/psycopg/docs/usage.html#copy

... but, ooooh!

> [`copy_expert()`](http://initd.org/psycopg/docs/cursor.html#cursor.copy_expert)
> Allows to handle more specific cases and to use all the COPY features available in PostgreSQL.

whit537:
Hmm ... "[The `ARRAY` constructor syntax can also be used](https://www.postgresql.org/docs/9.3/static/arrays.html#ARRAYS-INPUT)," but I still get:

```
psycopg2.DataError: array value must start with "{" or dimension information
CONTEXT:  COPY updates, line 1, column emails: "ARRAY['alice@example.com']"
```

whit537:
Probably "be used" there refers to `INSERT` and not `COPY`.

whit537:
> SELECT E'{"ab\\"c","de\\"f"}'::varchar[] #yeah, double-escape (literal first, then array)

https://www.postgresql.org/message-id/1398875542833-5802008.post@n5.nabble.com

😕

aandis:
@whit537 this is turning out to be more work than either of us thought eh.

whit537:
> PostgreSQL also accepts "escape" string constants, which are an extension to the SQL standard. An escape string constant is specified by writing the letter `E` (upper or lower case) just before the opening single quote, e.g., `E'foo'`.

https://www.postgresql.org/docs/9.3/static/sql-syntax-lexical.html#SQL-SYNTAX-STRINGS-ESCAPE

whit537:
> this is turning out to be more work than either of us thought eh.

As usual ... right @kaguillera? 🐰 :hole: 😁

whit537:
Okay, I don't think we need to worry about escape string constants, because that's not how `COPY` treats the value. It's just an array constant, not an escape string constant.

Okay!

/me preps a commit ...

whit537:
/me wrestles with escaping some more ...

whit537:
Okay, we have three layers here: JSON, Python, and Postgres. Let's work backwards from what we want to send to Postgres to make sure we handle the preceding stages properly.

whit537:
Sooooo ... what do we want to send to Postgres?

whit537:
We want to send `{"foo"}`.

whit537:
> You can put double quotes around any element value, and must do so if it contains commas or curly braces. (More details appear below.)

https://www.postgresql.org/docs/9.3/static/arrays.html#ARRAYS-INPUT

Where are the details below? Ah!

https://www.postgresql.org/docs/9.3/static/arrays.html#ARRAYS-IO

whit537:
Also: what if someone names their package `NULL`?

whit537:
> We want to send `{"foo"}`.

We want to send whatever was in the JSON. So if the JSON has a package name of `foo\\"bar` we want to send `foo\\"bar` to Postgres.

whit537:
Though actually this is only an issue when arraying values. Scalars we pass through as-is.

whit537:
So basically both JSON and Postgres array literals happen to use backslash-escaping. In Python we are seeing the unescaped value as given from `ijson` (which throws a parse error for bad JSON escaping), and then we want to _re_-escape it for Postgres.

What about the order of re-escaping? ... and what's up with backslash in Python raw strings? Ah:

> Specifically, _a raw string cannot end in a single backslash_ (since the backslash would escape the following quote character).

https://docs.python.org/2/reference/lexical_analysis.html#string-literals

whit537:
P.S. Python also uses backslash escaping. :-)

whit537:
Okay, so yeah: order of re-escaping does matter.

**Bad** _(quotes first, then slashes)_:

```
"
\"
\\"
```

**Good** _(slashes first, then quotes)_:

```
"
"
\"
```

whit537:
Okay! Encoding/quoting cleaned up in 13f5f5ffa089e34689418b6f6a6a7bf31d61d260.

Now what? I think it's time to rebase a bunch of code off this PR ... and then are we ready to think about deployment? 😮 

I have another hour left today.

whit537:
> rebase a bunch of code off this PR

I ended up folding it all together in 7cf1d6c8a5eb64f800b301391c8dfd43dfe1c843. Let's see what Travis says ...

whit537:
https://devcenter.heroku.com/articles/background-jobs-queueing

whit537:
I don't find any [PPAs providing `yajl`](https://launchpad.net/ubuntu/+ppas?name_filter=yajl) in the Travis [whitelist](https://github.com/travis-ci/apt-source-whitelist/blob/master/ubuntu.json), for which the approval queue is hopelessly [backlogged](https://github.com/travis-ci/apt-source-whitelist/issues) (`yajl` doesn't [appear](https://github.com/travis-ci/apt-source-whitelist/search?q=yajl&type=Issues&utf8=%E2%9C%93)).

whit537:
https://devcenter.heroku.com/articles/scheduled-jobs-custom-clock-processes

whit537:
Alright, I'm flailing. :-/

Off to https://github.com/gratipay/inside.gratipay.com/issues/757! I won't be back until late this evening, and then tomorrow morning the conference starts. I expect to be back tomorrow after lunch. 

Slipping, slipping away ...

whit537:
Two tasks: Travis, and then Heroku.

whit537:
At this point I'm [using](http://www.ubuntuupdates.org/package/webupd8/precise/main/base/libyajl2) some random PPA I picked up off the floor of the Internet ... at Travis, at least. I'm seeing 12.04 at Travis, 14.04 at Heroku. Let's see about Heroku ...

whit537:
Maybe the [scheduler](https://devcenter.heroku.com/articles/scheduler) is the easiest way to go to install this ...

whit537:
Ooooh ... I guess we also still need the `--if-modified-since` option ...

whit537:
❗️ ❗️ ❗️ 

screen shot 2016-10-26 at 7 09 26 am

whit537:
https://scheduler.heroku.com/dashboard

whit537:
Working up a command for the scheduler ...

```
[gratipay] $ curl https://gist.githubusercontent.com/whit537/fec53fb1f0618b3d5757f0ab687b7476/raw/25de82f6197df49b47d180db0d62b4e8c6f7f9f8/one 2>/dev/null | run defaults.env local.env env/bin/sync-npm serialize /dev/stdin 2>/dev/null
npm,testing-package,A package for testing,"{""alice@example.com"", ""bob@example.com""}"
[gratipay] $
```

whit537:
Setting up an app for testing ...

```
[gratipay] $ git push test chomp:master
Counting objects: 56007, done.
Delta compression using up to 4 threads.
Compressing objects: 100% (19999/19999), done.
Writing objects:  38% (21336/56007), 4.69 MiB | 219.00 KiB/s 
```

Gonna party like it's 19999. :)

whit537:
Hmm ...

```
remote:          Running setup.py install for cffi: started
remote:            Running setup.py install for cffi: finished with status 'error'
remote:            Complete output from command /app/.heroku/python/bin/python -u -c "import setuptools, token
ize;__file__='/tmp/pip-xAp4zy-build/setup.py';exec(compile(getattr(tokenize, 'open', open)(__file__).read().re
place('\r\n', '\n'), __file__, 'exec'))" install --record /tmp/pip-FXZ8Oi-record/install-record.txt --single-v
ersion-externally-managed --compile:
remote:            Package libffi was not found in the pkg-config search path.
remote:            Perhaps you should add the directory containing `libffi.pc'
remote:            to the PKG_CONFIG_PATH environment variable
remote:            No package 'libffi' found
remote:            Package libffi was not found in the pkg-config search path.
remote:            Perhaps you should add the directory containing `libffi.pc'
remote:            to the PKG_CONFIG_PATH environment variable
remote:            No package 'libffi' found
remote:            Package libffi was not found in the pkg-config search path.
remote:            Perhaps you should add the directory containing `libffi.pc'
remote:            to the PKG_CONFIG_PATH environment variable
remote:            No package 'libffi' found
remote:            Package libffi was not found in the pkg-config search path.
remote:            Perhaps you should add the directory containing `libffi.pc'
remote:            to the PKG_CONFIG_PATH environment variable
remote:            No package 'libffi' found
remote:            Package libffi was not found in the pkg-config search path.
remote:            Perhaps you should add the directory containing `libffi.pc'
remote:            to the PKG_CONFIG_PATH environment variable
remote:            No package 'libffi' found
remote:            running install
remote:            running build
remote:            running build_py
remote:            creating build
remote:            creating build/lib.linux-x86_64-2.7
remote:            creating build/lib.linux-x86_64-2.7/cffi
remote:            copying cffi/gc_weakref.py -> build/lib.linux-x86_64-2.7/cffi
remote:            copying cffi/verifier.py -> build/lib.linux-x86_64-2.7/cffi
remote:            copying cffi/backend_ctypes.py -> build/lib.linux-x86_64-2.7/cffi
remote:            copying cffi/recompiler.py -> build/lib.linux-x86_64-2.7/cffi
remote:            copying cffi/commontypes.py -> build/lib.linux-x86_64-2.7/cffi
remote:            copying cffi/__init__.py -> build/lib.linux-x86_64-2.7/cffi
remote:            copying cffi/lock.py -> build/lib.linux-x86_64-2.7/cffi
remote:            copying cffi/vengine_gen.py -> build/lib.linux-x86_64-2.7/cffi
remote:            copying cffi/setuptools_ext.py -> build/lib.linux-x86_64-2.7/cffi
remote:            copying cffi/cparser.py -> build/lib.linux-x86_64-2.7/cffi
remote:            copying cffi/vengine_cpy.py -> build/lib.linux-x86_64-2.7/cffi
remote:            copying cffi/ffiplatform.py -> build/lib.linux-x86_64-2.7/cffi
remote:            copying cffi/api.py -> build/lib.linux-x86_64-2.7/cffi
remote:            copying cffi/model.py -> build/lib.linux-x86_64-2.7/cffi
remote:            copying cffi/cffi_opcode.py -> build/lib.linux-x86_64-2.7/cffi
remote:            copying cffi/_cffi_include.h -> build/lib.linux-x86_64-2.7/cffi
remote:            copying cffi/parse_c_type.h -> build/lib.linux-x86_64-2.7/cffi
remote:            copying cffi/_embedding.h -> build/lib.linux-x86_64-2.7/cffi
remote:            running build_ext
remote:            building '_cffi_backend' extension
remote:            creating build/temp.linux-x86_64-2.7
remote:            creating build/temp.linux-x86_64-2.7/c
remote:            gcc -pthread -fno-strict-aliasing -g -O2 -DNDEBUG -g -fwrapv -O3 -Wall -Wstrict-prototypes -fPIC -DUSE__THREAD -I/usr/include/ffi -I/usr/include/libffi -I/app/.heroku/python/include/python2.7 -c c/_cffi_backend.c -o build/temp.linux-x86_64-2.7/c/_cffi_backend.o
remote:            c/_cffi_backend.c:15:17: fatal error: ffi.h: No such file or directory
remote:             #include <ffi.h>
remote:                             ^
remote:            compilation terminated.
remote:            error: command 'gcc' failed with exit status 1
remote:
remote:            ----------------------------------------
remote:        Command "/app/.heroku/python/bin/python -u -c "import setuptools, tokenize;__file__='/tmp/pip-xAp4zy-build/setup.py';exec(compile(getattr(tokenize, 'open', open)(__file__).read().replace('\r\n', '\n'), __file__, 'exec'))" install --record /tmp/pip-FXZ8Oi-record/install-record.txt --single-version-externally-managed --compile" failed with error code 1 in /tmp/pip-xAp4zy-build/
remote:  !     Push rejected, failed to compile Python app.
remote:
remote:  !     Push failed
remote: Verifying deploy....
remote:
remote: !       Push rejected to tranquil-ocean-34844.
```

whit537:
I think I need to use a different buildpack ...

whit537:
This mostly worked, I had to edit `defaults.env` to get around a goofy error with the "3 minutes" setting.

```
heroku config:set --remote test `cat defaults.env|grep -v '^#'`
```

Then:

```
heroku config:set --remote test BUILDPACK_URL=`heroku config:get BUILDPACK_URL --remote heroku`
```

whit537:
Let's try again!

```
git push test chomp:master
```

...

whit537:
👍 

Now for the database!

```
remote:        Successfully installed Babel-1.3 Jinja2-2.7.3 MarkupSafe-0.23 algorithm-1.0.0 aspen-0.42 aspen-jinja2-0.4 balanced-1.1.1 boto3-1.3.0 botocore-1.4.13 braintree-3.14.0 certifi-0.0.8 cffi-1.6.0 chardet-1.0.1 cryptography-1.3.2 dependency-injection-1.1.0 docutils-0.12 enum34-1.1.4 environment-1.0.0 fake-factory-0.3.2 filesystem-tree-1.0.0 first-2.0.1 futures-3.0.4 gunicorn-18.0 honcho-cfcc7ebb idna-2.1 invoke-0.7.0 ipaddress-1.0.16 iso8601-0.1.4 jmespath-0.9.0 libsass-0.3.0 misaka-1.0.2 oauthlib-0.7.2 postgres-2.2.1 psycopg2-2.5.1 pyasn1-0.1.3 pycparser-2.14 python-dateutil-2.5.3 python-mimeparse-0.1.4 pytz-2014.7 raven-3.1.4 requests-2.6.0 requests-oauthlib-0.4.2 simplejson-2.3.2 six-1.8.0 uritemplate-0.6 wac-0.23 xmltodict-0.8.4
remote: 
remote: -----> Discovering process types
remote:        Procfile declares types -> web
remote: 
remote: -----> Compressing...
remote:        Done: 74.6M
remote: -----> Launching...
remote:        Released v5
remote:        https://tranquil-ocean-34844.herokuapp.com/ deployed to Heroku
remote: 
remote: Verifying deploy... done.
To https://git.heroku.com/tranquil-ocean-34844.git
 * [new branch]      chomp -> master
[gratipay] $
```

whit537:
Gosh but running `make schema` on heroku at all gives me the willies like really the willies. 😳 😧 😖

whit537:
```
ALTER TABLE
ALTER TABLE
ALTER TABLE
DROP VIEW
UPDATE 0
ALTER TABLE
ALTER TABLE
ALTER TABLE
ALTER TABLE
CREATE VIEW
CREATE TRIGGER
BEGIN
DROP VIEW
DROP TABLE
COMMIT
BEGIN
ALTER TABLE
ALTER TABLE
COMMIT
BEGIN
DROP VIEW
DROP TABLE
CREATE TABLE
CREATE VIEW
COMMIT
==============================================================================
Looking for sql/branch.sql ...

BEGIN
CREATE TABLE
COMMIT

==============================================================================
~ $ 
```

Gosh that was terrifying. 😥

whit537:
Okay! Let's load some packages! 💃

screen shot 2016-10-26 at 8 58 44 am

whit537:
Yes! 💃 

```
~ $ curl $URL | sync-npm serialize /dev/stdin | sync-npm upsert /dev/stdin
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100   279  100   279    0     0   6578      0 --:--:-- --:--:-- --:--:--  6642
skipping None
skipping {u'description': '', u'package_manager': 'npm', u'emails': [], u'name': '_updated'}
processed 1 packages in   0 seconds
~ $
```

whit537:
Double yes! 💃 💃 

```
tranquil-ocean-34844::DATABASE=> select * from packages;
┌────┬─────────────────┬─────────────────┬───────────────────────┬────────┬────────────┬─────────────┬────────
│ id │ package_manager │      name       │      description      │ readme │ readme_raw │ readme_type │        
├────┼─────────────────┼─────────────────┼───────────────────────┼────────┼────────────┼─────────────┼────────
│  2 │ npm             │ testing-package │ A package for testing │        │            │             │ {alice@
└────┴─────────────────┴─────────────────┴───────────────────────┴────────┴────────────┴─────────────┴────────
(1 row)

lines 1-7/7 (END) 
```

whit537:
Okay! Let's light this candle! :candle:

whit537:
I've set up a 10-minute scheduler job on the test app. If that works then my plan is to:

- switch to the real URL
- merge and deploy this PR
- schedule a daily run for now

whit537:
YEEEEEEEEEEEEEEEEEEE

screen shot 2016-10-26 at 10 20 07 am

whit537:
I'm going to land this as-is, and reticket loading up readmes as well as incrementalizing updates.

whit537:
Ready to go once Travis is green ...

whit537:
Okay! Here we go! ...

whit537:
!m @aandis 

!m *

whit537:
Coming back from https://github.com/gratipay/gratipay.com/issues/4148#issuecomment-256382855

```
processed 371932 packages in 308 seconds
  File "/app/.heroku/python/bin/sync-npm", line 9, in <module>
    load_entry_point('gratipay', 'console_scripts', 'sync-npm')()
  File "/app/gratipay/package_managers/sync.py", line 137, in main
    globals()[args.command](args)
  File "/app/gratipay/package_managers/sync.py", line 101, in upsert
    'FROM STDIN WITH (FORMAT csv)', fp)
psycopg2.IntegrityError: null value in column "description" violates not-null constraint
DETAIL:  Failing row contains (2, npm, 0, null, , , , {zolikahan@gmail.com}).
CONTEXT:  COPY updates, line 1: "npm,0,,"{""zolikahan@gmail.com""}""
```

whit537:
More later!

whit537:
Local. 👍 

```
gratipay-test=# select count(*) from packages;
┌────────┐
│ count  │
├────────┤
│ 372147 │
└────────┘
(1 row)
```

whit537:
Okay! Now what's up with the tests?

whit537:
Picking up in https://github.com/gratipay/gratipay.com/pull/4158 ...

PR #4129, opened 2016-09-22
Tweak timing for Travis

whit537:
Fixes #4118 (hopefully!)

PR #4154, opened 2016-10-22
Wrap marky-markdown

whit537:
[marky-markdown](https://github.com/npm/marky-markdown#command-line-usage) is the markdown parser that npm uses. We should use it with npm package readmes. There's a [CLI](https://github.com/npm/marky-markdown#command-line-usage) that we should be able to wrap. It doesn't take `stdin` but we can work around that using `/dev/stdin` like so:

```
marky-markdown /dev/stdin
```

### DIFF
--- a/gratipay/utils/markdown.py
+++ b/gratipay/utils/markdown.py
@@ -15,6 +15,5 @@ def render(markdown):
 def marky(markdown):
     """Process markdown the same way npm does.
     """
-    echo = Popen(("echo", markdown), stdout=PIPE)
-    marky = Popen(("marky-markdown", "/dev/stdin"), stdin=echo.stdout, stdout=PIPE)
-    return Markup(marky.communicate()[0])
+    marky = Popen(("marky-markdown", "/dev/stdin"), stdin=PIPE, stdout=PIPE)
+    return Markup(marky.communicate(markdown)[0])
--- a/gratipay/utils/markdown.py
+++ b/gratipay/utils/markdown.py
@@ -1,5 +1,8 @@
+import subprocess
+
 from markupsafe import Markup
 import misaka as m  # http://misaka.61924.nl/
+
 
 def render(markdown):
     return Markup(m.html(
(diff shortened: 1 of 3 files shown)


aandis:
Will calling an external process on each render be performant enough?  Why not use something like https://github.com/lepture/mistune?

aandis:
Aha, github has it's own flavor of markdown and mistune and the likes may not support all those features.

aandis:
Found [py-gfm](https://github.com/Zopieux/py-gfm) but https://github.com/Zopieux/py-gfm#unsupported-features :/

aandis:
[grip](https://github.com/joeyespo/grip#api) is pretty cool but it's no different than what we're trying to do with marky-markdown.

aandis:
do we have tests for this?

aandis:
do we have tests for this?

whit537:
@aandis No tests, no. This won't be called during web requests. We'll call it in an async worker when we recrawl npm (#4153).

whit537:
>> do we have tests for this?
>
> No tests, no.

But we should. :-)

whit537:
@aandis For now can we add even just one test that runs some very simple markdown through `marky` and verifies that it comes out as HTML? We don't need to exhaustively tests marky-markdown's features.

aandis:
:green_apple:

whit537:
> 🍏

Does that mean "Ready for review?" :-)

aandis:
Yep. :)

gratipay/utils/markdown.py
```diff
@@ -1,9 +1,20 @@
...
         render_flags=m.HTML_SKIP_HTML | m.HTML_TOC | m.HTML_SMARTYPANTS | m.HTML_SAFELINK
     ))
+
+
+def marky(markdown):
+    """Process markdown the same way npm does.
+    """
+    echo = Popen(("echo", markdown), stdout=PIPE)
+    marky = Popen(("marky-markdown", "/dev/stdin"), stdin=echo.stdout, stdout=PIPE)
+    return Markup(marky.communicate()[0])
```
whit537: Can't we pass `markdown` into the `communicate` call rather than the `echo` hack?

aandis: Didn't know pipe could work with that. Nice! Done in ecdc8e0205f80ea9b123be0f5673b2e5a2a1dc90

whit537:
Sigh, looks like #4129 didn't work. :-/

whit537:
!m @aandis 💃

PR #4158, opened 2016-10-26
Fix npm sync bug with empty descriptions

whit537:
Part of #4148.


whit537:
[H*ck yeah](https://travis-ci.org/gratipay/gratipay.com/builds/170951770#L1265).

Issue #4133, opened 2016-09-30
Team list for users with multiple teams is poorly spaced


mattbk:
...if the team name is sort of long:
![2016-09-30_11-15-15](https://cloud.githubusercontent.com/assets/3729038/18998683/53f69f70-86ff-11e6-9134-05e32f5c0fd4.png)

(Screenshot captured with ShareX, who are on Gratipay! https://gratipay.com/sharex/)

Issue #3852, opened 2015-10-29
Differentiate participants and projects


whit537:
Right now they're too close visually which is confusing. What's the difference between them?

![screen shot 2015-10-29 at 10 28 08 am](https://cloud.githubusercontent.com/assets/134455/10821248/c73592a2-7e27-11e5-9802-780c1fdc95aa.png)
![screen shot 2015-10-29 at 10 28 02 am](https://cloud.githubusercontent.com/assets/134455/10821249/c73a7d58-7e27-11e5-8181-3dd88349165b.png)
##

whit537:
+1 https://github.com/gratipay/gratipay.com/issues/3853

mattbk:
+1 from https://twitter.com/hadleywickham/status/688105609456058370.

mattbk:
We've talked about parallels between these pages before, such as always having the donation action area on the left side (as shown in above screenshots). So I still think the ~user left sidebar should contain something like "Give to one of ~user's teams."

More extravagantly, how about ~user pages include team page snippets (the main body) that you can scroll through, each with its own donate button? Then each team page could be the reverse, and list the team members, albeit not with the donation links.

**catapultpgh's Teams**
![screen shot 2016-01-19 at 2 51 54 pm](https://cloud.githubusercontent.com/assets/3729038/12431684/642591b4-bebc-11e5-930e-4af3f1844ea6.png)
![screen shot 2016-01-19 at 2 51 54 pm](https://cloud.githubusercontent.com/assets/3729038/12431684/642591b4-bebc-11e5-930e-4af3f1844ea6.png)
![screen shot 2016-01-19 at 2 51 54 pm](https://cloud.githubusercontent.com/assets/3729038/12431684/642591b4-bebc-11e5-930e-4af3f1844ea6.png)

mattbk:
+1 from https://gratipay.freshdesk.com/helpdesk/tickets/4104

whit537:
+1 from @dakur in https://github.com/gratipay/gratipay.com/issues/3993:

[...]

Issue #396, opened 2012-11-27
explore partnership with Payoneer


chadwhitacre:
Reticketed from #126. Payoneer solves international payouts with prepaid cards. Some concern about legality in India (and elsewhere?). What's the API like for payouts?

I contacted sales and have heard back.

PR #4169, opened 2016-10-30
Add team previews on ~user pages

rohitpaulk:


rohitpaulk:
Sharing early so that I can get feedback - I'm not happy with the design, and the CSS is completely hacky. 

Here's how this currently looks:

I've used two users as samples - ~bbatsov who has 4 teams, and ~mattbk who has 2 (one has a long name)

**4 Teams, viewed as public:** 

screen shot 2016-10-31 at 2 01 15 am

**4 Teams, viewed as user:** 

screen shot 2016-10-31 at 2 01 52 am

rohitpaulk:
This is based on the initial idea: https://github.com/gratipay/gratipay.com/issues/3852#issuecomment-257146433

Copy pasting that image here: 

img_20161030_171508

I also want to experiment with cards that are more like the identities page (rectangular, longer horizontally). 

img_20161031_021838

whit537:
SO EXCITING!!!!!!!!!!!!!!! :D

whit537:
Should we show `ngivers` and `receiving`? Or should we show the ~user's `taking` from the team? For now for almost everyone that will be equal to `receiving`, since Gratipay is the only team for which distributions are enabled.

Whichever it is, maybe try the `$Ideal` font there?

What are you unhappy with about the design? What problems are you trying to solve with it and what isnt' being solved?

whit537:
I would say this is close—and definitely better than what we have. How should approach this in relation to the identity cards? If we take the time to properly factor the CSS now to account for both cases, then it should be fairly straightforward to apply to payment methods soon after.

whit537:
> I also want to experiment with cards that are more like the identities page (rectangular, longer horizontally).

👍

whit537:
https://github.com/gratipay/gratipay.com/issues/3852#issuecomment-257177315

> Yep, would be great if we could come up with a pattern that fit all of those. We could call this component a 'card' - it **must** contain an avatar + title text, and optionally **might** contain sub-text. 

👍

rohitpaulk:
> Should we show ngivers and receiving? Or should we show the ~user's taking from the team? For now for almost everyone that will be equal to receiving, since Gratipay is the only team for which distributions are enabled.

I think `ngivers` and `receiving` makes more sense, because they're properties of the team and not the user. Maybe just `receiving` would do? What is/are the most important metric(s) for a team/project on Gratipay? 

Exposing the ~user's breakdown of takes per team definitely has value, but I think that'd better fit in something like a dropdown on hovering over 'Taking' in the chalkboard.

rohitpaulk:
> What are you unhappy with about the design? What problems are you trying to solve with it and what isn't being solved?

I'm talking only about graphic design here: 

- The brown border seems a bit off - in particular, I don't think it interacts with the green banner well. 
- Too much green inside the box - the top banner, the sub-text *and* the divider between the two sub-texts. I think we should look into neutralizing one of those so that the others stand out better. 
- The divider between the two sub-texts is 2 px wide and is set to a 50% opacity green (#396), which looks a bit weird to me.  
- It would be great if we could keep the vertical length the same, using a design that worked well with both short and long names. Maybe try reducing the font a bit?

whit537:
> I think we should look into neutralizing one of those so that the others stand out better.

Seems like the subtext is the least important info—maybe try dropping that to gray + small caps like with the platform name in the social profiles listings?

> It would be great if we could keep the vertical length the same, using a design that worked well with both short and long names. Maybe try reducing the font a bit?

You could also use `overflow: hidden` and then space the lines so that if the team name wraps, it simply pushes the subtext out of the card. They can always click through for that info.

aandis:
@rohitpaulk this looks nice! few improvements I could think of - 
screen shot 2016-10-31 at 11 43 06 pm

1. Maybe try getting rid of those overflows? It's barely visible unless you zoom in but hard to unsee once you see it. ;)
2. I experimented with a black-ish divider (matching the team title color above) which would look ok too if it were a bit thinner.
3. I think the whole card should be clickable and not *just* the team name.

aandis:
> You could also use overflow: hidden and then space the lines so that if the team name wraps, it simply pushes the subtext out of the card. They can always click through for that info.

yep. could also ellipsize the team name which would give you same vertical length for all the cards.

mattbk:
I completely missed this! Looks great.

Will also close https://github.com/gratipay/gratipay.com/issues/4133.

Just a thought; maybe making teams look like they do on the front page or in search would be better?

whit537:
This seems pretty close and would be nice to have. Might pick this up ...

whit537:
Closing in favor of #4353.

Issue #4168, opened 2016-10-29
Disconnect PayPal gives "Looks like you've found a bug! Sorry for the inconvenience, we'll get it fixed ASAP!"


toy:
Account toy, I have only one verified email, when clicking Disconnect button on PayPal page, I get the error.

whit537:
Sorry, @toy. :-/

Don't have time now, but I will look into this when I get a chance if someone else doesn't get to it first ...

toy:
It is not really urgent. I already saw this bug, but assumed that you are getting notifications, but as the bug was not fixed, I've created an issue.

rohitpaulk:
Copying the error from Sentry: 

```
IntegrityError: update or delete on table "exchange_routes" violates foreign key constraint "exchanges_route_fkey" on table "exchanges"
DETAIL:  Key (id)=(10189) is still referenced from table "exchanges".
```

Looks like we're trying to delete the exchange route, but the foreign key constraint between `exchanges` -> `exchange_routes` prevents us from doing so. 

Hmm, maybe we should be soft deleting exchange routes?

whit537:
> assumed that you are getting notifications

We are getting notifications, but we're not staying on top of them. Sorry. 😞

rohitpaulk:
Aha! 

<img width="572" alt="screen shot 2016-10-31 at 6 37 24 pm" src="https://cloud.githubusercontent.com/assets/3893573/19855171/190b5f56-9f99-11e6-9696-671ad4db3ec1.png">

rohitpaulk:
Added by me in https://github.com/gratipay/gratipay.com/commit/0500105b9dc148855b7957963b3949d89a7cc3b4, and @whit537 added the comment in https://github.com/gratipay/gratipay.com/commit/e27a2283211bf9ade2121dec222714110e37b93b. 

I think the best way out here is to update the error to 'invalidated' irrespective of the `network` - and when user tries to add the route again for paypal, we revive the previous `exchange_route`.

rohitpaulk:
Moving implementation details to a PR so that we don't pollute this issue

toy:
Still not possible to disconnect

whit537:
Mind giving this another shot, @toy? We just deployed #4170 in an attempt to fix this.

toy:
@whit537 Thanks! It seems to work, as I was able to disconnect the paypal account.

whit537:
Woo-hoo! Thanks for your patience. :-)

PR #4170, opened 2016-10-31
Fix bug on disconnecting PayPal

rohitpaulk:
#4168

### Todo

- [x] fix accidentally failing tests
- [x] ~~lightly refactor `test_routes.py`~~ proved difficult due to fixture entanglement w/ `BillingHarness`(?)
- [x] fix (rename?) `test_delete_paypal_with_exchanges`
- [x] ensure we can readd a PayPal
- [x] ensure we can revive an old PayPal

### DIFF
--- a/gratipay/models/exchange_route.py
+++ b/gratipay/models/exchange_route.py
@@ -8,13 +8,6 @@ class ExchangeRoute(Model):
 
     typname = "exchange_routes"
 
-    ERROR_INVALIDATED = 'invalidated'
-
-    def __bool__(self):
-        return self.error != self.ERROR_INVALIDATED
-
-    __nonzero__ = __bool__
-
     @classmethod
     def from_id(cls, id):
         r = cls.db.one("""
@@ -78,13 +71,16 @@ class ExchangeRoute(Model):
             # XXX This doesn't sound right. Doesn't this corrupt history pages?
             self.db.run("DELETE FROM exchange_routes WHERE id=%s", (self.id,))
         else:
-            self.update_error(self.ERROR_INVALIDATED)
+            self.db.run("UPDATE exchange_routes SET is_deleted = true WHERE id = %s", (self.id, ))
+            self.set_attributes(is_deleted=True)
 
     def update_error(self, new_error):
         id = self.id
         old_error = self.error
-        if old_error == self.ERROR_INVALIDATED:
+
+        if self.is_deleted:
             return
+
         self.db.run("""
             UPDATE exchange_routes
                SET error = %(new_error)s
--- a/gratipay/models/exchange_route.py
+++ b/gratipay/models/exchange_route.py
@@ -8,8 +8,10 @@ class ExchangeRoute(Model):
 
     typname = "exchange_routes"
 
+    ERROR_INVALIDATED = 'invalidated'
+
     def __bool__(self):
-        return self.error != 'invalidated'
+        return self.error != self.ERROR_INVALIDATED
 
     __nonzero__ = __bool__
 
@@ -76,12 +78,12 @@ class ExchangeRoute(Model):
             # XXX This doesn't sound right. Doesn't this corrupt history pages?
             self.db.run("DELETE FROM exchange_routes WHERE id=%s", (self.id,))
         else:
-            self.update_error('invalidated')
+            self.update_error(self.ERROR_INVALIDATED)
 
     def update_error(self, new_error):
         id = self.id
         old_error = self.error
-        if old_error == 'invalidated':
+        if old_error == self.ERROR_INVALIDATED:
             return
         self.db.run("""
             UPDATE exchange_routes
--- a/gratipay/models/exchange_route.py
+++ b/gratipay/models/exchange_route.py
@@ -66,15 +66,8 @@ class ExchangeRoute(Model):
     def invalidate(self):
         if self.network == 'braintree-cc':
             braintree.PaymentMethod.delete(self.address)
-
-        # For Paypal, we remove the record entirely to prevent
-        # an integrity error if the user tries to add the route again
-        if self.network == 'paypal':
-            # XXX This doesn't sound right. Doesn't this corrupt history pages?
-            self.db.run("DELETE FROM exchange_routes WHERE id=%s", (self.id,))
-        else:
-            self.db.run("UPDATE exchange_routes SET is_deleted = true WHERE id = %s", (self.id, ))
-            self.set_attributes(is_deleted=True)
+        self.db.run("UPDATE exchange_routes SET is_deleted = true WHERE id = %s", (self.id, ))
+        self.set_attributes(is_deleted=True)
 
     def update_error(self, new_error):
         id = self.id
--- a/gratipay/models/exchange_route.py
+++ b/gratipay/models/exchange_route.py
@@ -2,6 +2,7 @@ from __future__ import absolute_import, division, print_function, unicode_litera
 
 import braintree
 from postgres.orm import Model
+from gratipay.models import add_event
 
 
 class ExchangeRoute(Model):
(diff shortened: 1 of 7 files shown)


rohitpaulk:
Copying from #4168: 

Aha! 

screen shot 2016-10-31 at 6 37 24 pm

The `DELETE` statement was added by me in https://github.com/gratipay/gratipay.com/commit/0500105b9dc148855b7957963b3949d89a7cc3b4, and @whit537 added the comment above that in https://github.com/gratipay/gratipay.com/commit/e27a2283211bf9ade2121dec222714110e37b93b. 

I think the best way out here is to update the error to 'invalidated' irrespective of the `network` - and when user tries to add the route again for paypal, we revive the previous `exchange_route`.

rohitpaulk:
I'm leaning towards "Update the ID to the next auto-incremented ID when reviving the previous exchange_route".

rohitpaulk:
Nah, now "Change the condition for current_exchange_routes to not include routes that have error = 'invalidated'" makes more sense.

rohitpaulk:
Dealing with the `invalidated` constant is turning out to be a pain - I think it is time to treat soft-deleted exchange routes as first-citizens by introducing a `is_deleted` field.

rohitpaulk:
> Dealing with the invalidated constant is turning out to be a pain

This leads to bugs and special handling in many places.

- When a card is removed, we currently tell the user 'Your credit card is failing' (we should say that they haven't linked a credit card)

screen shot 2016-11-14 at 8 35 17 pm

We also end up showing them a notification with the same text.

- In the new functionality I'm going to introduce, we'll have to do a lot of logic in the application layer that could've been handled at the DB layer if we had a boolean `is_deleted` field.

whit537:
Sentry: https://sentry.io/gratipay/gratipay-com/issues/189911878/.

whit537:
@rohitpaulk Any chance you could move this forward, or perhaps find a way to fix #4168 as a manual one-off? 🙇

whit537:
Rebased, was 1c6fa2829010ca9965779240c0ebbf2ec422c8da.

whit537:
> revive the previous `exchange_route`

I agree.

> we're just picking the latest ID in that case

Yeah that part is weird. I'd expect a ctime/mtime in there to sort on.

> I think it is time to treat soft-deleted exchange routes as first-citizens by introducing a `is_deleted` field.

👍

whit537:
Looks like we're pretty far along here. I'm picking this up ...

whit537:
Rebased, was f8c90be61d78db4b002d0298d23ef80acb69fde6.

whit537:
Started a todo in the description.

whit537:
I think that's it! Ready for review @nobodxbodon @mattbk @JessaWitzel @rohitpaulk et al.

mattbk:
Works for me after remembering to `make schema`.

gratipay/models/exchange_route.py
```diff
@@ -69,20 +67,35 @@ def insert(cls, participant, network, address, error='', fee_cap=None, cursor=No
...
             braintree.PaymentMethod.delete(self.address)
-
-        # For Paypal, we remove the record entirely to prevent
-        # an integrity error if the user tries to add the route again
-        if self.network == 'paypal':
-            # XXX This doesn't sound right. Doesn't this corrupt history pages?
-            self.db.run("DELETE FROM exchange_routes WHERE id=%s", (self.id,))
-        else:
-            self.update_error('invalidated')
+        with self.db.get_cursor() as cursor:
```
nobodxbodon: To confirm, we don't need to check if network is paypal anymore here?

gratipay/models/exchange_route.py
```diff
@@ -69,20 +67,35 @@ def insert(cls, participant, network, address, error='', fee_cap=None, cursor=No
...
+        self.set_attributes(is_deleted=False)
 
     def update_error(self, new_error):
         id = self.id
         old_error = self.error
-        if old_error == 'invalidated':
+
+        if self.is_deleted:
             return
+
```
nobodxbodon: nit: move this check to the top of method?

gratipay/models/exchange_route.py
```diff
@@ -53,12 +49,14 @@ def from_address(cls, participant, network, address):
         return r
 
     @classmethod
-    def insert(cls, participant, network, address, error='', fee_cap=None, cursor=None):
+    def insert(cls, participant, network, address, error='', fee_cap=None, cursor=None,
+                                                                                 is_deleted=False):
```
nobodxbodon: as we don't really "delete' the route, shall we name it (and in the database) something like 'active'?

tests/py/test_routes.py
```diff
@@ -26,34 +24,77 @@ def test_associate_and_delete_valid_card(self):
...
+
+        assert not self.roman.has_payout_route
+
+    def test_add_new_paypal_address(self):
+        self.add_and_verify_email(self.roman, 'roman@gmail.com')
+        self.add_and_verify_email(self.roman, 'candle@gmail.com')
+
+        self.hit('roman', 'associate', 'paypal', 'roman@gmail.com')
+        self.hit('roman', 'delete', 'paypal', 'roman@gmail.com')
+        self.hit('roman', 'associate', 'paypal', 'candle@gmail.com')
```
nobodxbodon: could you extract above 5 lines to separate method, and reuse in test case below?

www/~/%username/routes/associate.json.spt
```diff
@@ -40,7 +40,13 @@ if change:
     elif network == 'paypal':
         if address not in participant.get_verified_email_addresses():
             raise Response(400, _("Only verified email addresses allowed."))
-        ExchangeRoute.insert(participant, network, address)
+        try:
+            ExchangeRoute.insert(participant, network, address)
+        except IntegrityError as e:
+            if e.pgcode != errorcodes.UNIQUE_VIOLATION:
+                raise
+            existing = ExchangeRoute.from_address(participant, network, address)
```
nobodxbodon: if I'm not mistaken, this occurs when insertion fails. Could we check before insertion instead? And one nit: rename existing to existing_route?

gratipay/models/exchange_route.py
```diff
@@ -69,20 +67,35 @@ def insert(cls, participant, network, address, error='', fee_cap=None, cursor=No
...
-        if self.network == 'paypal':
-            # XXX This doesn't sound right. Doesn't this corrupt history pages?
-            self.db.run("DELETE FROM exchange_routes WHERE id=%s", (self.id,))
-        else:
-            self.update_error('invalidated')
+        with self.db.get_cursor() as cursor:
+            self.db.run("UPDATE exchange_routes SET is_deleted=true WHERE id=%s", (self.id,))
+            payload = dict( id=self.participant.id
+                          , exchange_route=self.id
+                          , action='invalidate route'
```
nobodxbodon: not quite familiar with this 'action' field. I imagine there are a set of actions, each of which should be constants. IMO they are better to be defined in one place, but feel free to ignore if it involves too much refactoring.

rohitpaulk:
Rebased on master.

gratipay/models/exchange_route.py
rohitpaulk: Hmm, I don't think it is worth adding `is_deleted` to the `insert` function. No one (other than in tests) is going to insert a deleted route, are they

www/~/%username/routes/associate.json.spt
rohitpaulk: Checking before insertion makes us prone to race conditions (two people check, both find no route, both try to write, one fails). 

I'm changing `existing` to `existing_route`

gratipay/models/exchange_route.py
rohitpaulk: In fact, we should only be accepting `participant`, `network`, `address` and `fee_cap` when creating an exchange route. Routes are never created with errors, and routes are never created with `is_deleted=true`

gratipay/models/exchange_route.py
rohitpaulk: No, we don't. We've now changed every route to be soft-deleted

gratipay/models/exchange_route.py
rohitpaulk: Event properties are logged freeform JSON, I don't think we reuse these values between multiple functions anywhere

tests/py/test_routes.py
rohitpaulk: IMO that will be more confusing to read

rohitpaulk:
Ready for review, @nobodxbodon

PR #4172, opened 2016-11-04
#3924 remove inapplicable badges; adjust title

nobodxbodon:
Just noticed there are many test failures:
5 failed, 677 passed, 36 xfailed, 17 error
Is there something wrong with my local settings? Or they are just broken?

PR #4173, opened 2016-11-04
Fix build breakage introduced in npm syncing

whit537:
Should address at least some of https://github.com/gratipay/gratipay.com/pull/4172#issue-187266655.

### DIFF
--- a/gratipay/package_managers/sync.py
+++ b/gratipay/package_managers/sync.py
@@ -15,6 +15,11 @@ log = lambda *a: print(*a, file=sys.stderr)
 NULL = uuid.uuid4().hex
 
 
+def import_ijson():
+    import ijson.backends.yajl2_cffi as ijson
+    return ijson
+
+
 def arrayize(seq):
     """Given a sequence of str, return a Postgres array literal str.
     """
@@ -48,7 +53,7 @@ def serialize_one(out, package):
 def serialize(args):
     """Consume raw JSON from the npm registry and spit out CSV for Postgres.
     """
-    import ijson.backends.yajl2_cffi as ijson
+    ijson = import_ijson()
 
     path = args.path
     parser = ijson.parse(open(path))
--- a/tests/py/test_npm_sync.py
+++ b/tests/py/test_npm_sync.py
@@ -4,8 +4,10 @@ from __future__ import absolute_import, division, print_function, unicode_litera
 
 from subprocess import Popen, PIPE
 
+import pytest
+from gratipay.utils import markdown
 from gratipay.testing import Harness
-from gratipay.package_managers import readmes
+from gratipay.package_managers import readmes, sync
 
 
 def load(raw):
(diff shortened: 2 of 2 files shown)


whit537:
I don't like that our tests could silently pass if one of our dependencies goes missing at Travis, but I'm not sure of the best way to address that.

whit537:
Ready for review and merge, @nobodxbodon @kaguillera.

nobodxbodon:
LGTM. I wonder how review process works? The committer waits for at least one contributor giving green light?

whit537:
@nobodxbodon Yeah, if it looks good to you then you can go ahead and merge! :-)

nobodxbodon:
Sorry about the premature merging. I just realized that I overlooked the python parts. As I'm not that familiar with Travis, @kaguillera please review again.
@whit537 how could Travis miss any of our dependencies? Does that mean our code base replies on some thirty-party library that's not recognized by Travis?

whit537:
> how could Travis miss any of our dependencies? Does that mean our code base replies on some thirty-party library that's not recognized by Travis?

We are [installing](https://github.com/gratipay/gratipay.com/blob/53c7b5cca05355d666aff9701e1bd9cbc5381297/.travis.yml#L12-L14) the necessary dependencies at Travis. My concern was that if that dependency installation started failing for some reason, we wouldn't notice right away, because we would simply start silently skipping said tests at Travis.

kaguillera:
@nobodxbodon as you said it looks good to me. @whit537 as far as I see it will only ignore those test are associated with ijson and marky. That seems appropriate.

rohitpaulk:
Umm shouldn't we be ensuring that those dependencies are installed on Travis? We've got full root access, what prevents us from doing so?

whit537:
@rohitpaulk They _are_ [installed](https://github.com/gratipay/gratipay.com/blob/53c7b5cca05355d666aff9701e1bd9cbc5381297/.travis.yml#L12-L14) on Travis today. My point was just that if the deps installation ever breaks (e.g., the places we're installing from change), then we wouldn't notice because the tests would just be skipped and Travis would still be green. No?

rohitpaulk:
I just read through the entire thread - what I meant to say is "Why make this change?". If tests are failing for anyone, the right way to fix it is to install the dependencies (which we could include in our bootstrap scripts) - not to skip tests.

rohitpaulk:
> we wouldn't notice because the tests would just be skipped and Travis would still be green. No?

Exactly, this is the part that I don't like

nobodxbodon:
> If tests are failing for anyone, the right way to fix it is to install the dependencies (which we could include in our bootstrap scripts) - not to skip tests. 

+1 is there a way to config Travis so missing dependency leads to failure instead of skipping? Feels like a common need to me.

whit537:
> the right way to fix it is to install the dependencies

Yeah, but yajl and npm are so heavy. 😞  Options:

- Allow for the non-yajl `ijson` (less performant).
- Move this back out into [`gdr.rocks`](https://github.com/gratipay/gdr.rocks).

PR #4187, opened 2016-11-12
Remove "other receiving options" from ~user profile

### LINKED ISSUE #4174: remove bitcoin part in "Other Giving Options" from user profile
Discussed [here](https://github.com/gratipay/gratipay.com/issues/3924#issuecomment-258261908).

mattbk:
Closes #4174.


mattbk:
There are other mentions in the code:
- https://github.com/gratipay/gratipay.com/search?utf8=%E2%9C%93&q=venmo
- https://github.com/gratipay/gratipay.com/search?q=bitcoin&type=Code&utf8=%E2%9C%93

Thoughts on whether to leave them in, in case we need them again as routes? 

(for this reason, I'll be reverting the last commit, which deletes the Venmo template.)

rohitpaulk:
Hmm, why revert @c50ea53? I think it'd be best to delete it - less code for us to worry about. It'll always be in git history if we have to revive it in the future.

rohitpaulk:
I've pushed a commit to remove the JS code that powered the cryptocoin input

whit537:
> less code for us to worry about

If we're going to do this, let's go the whole way, and remove things like `participant. get_cryptocoin_addresses` and the relevant database fields as well. No?

JessaWitzel:
@whit537  removed the cryptocoin references but what do you mean re: participant? That's in a lot of places in the code and not always in reference to payments?

whit537:
I think maybe a word wrap broke the communication? :)

```
participant.get_cryptocoin_addresses
```

In other words, let's get rid of the `get_cryptocoin_addresses` method on the `Participant` class (not the whole class ;). Is that more sensical?

nobodxbodon:
What's left here? Seems `get_cryptocoin_addresses` is gone in 1d7b10d. Ready to review?

nobodxbodon:
PR LGTM. @mattbk @JessaWitzel @whit537 if any of you has TODO about it, if it's not absolutely necessary could you add comment in code and create new ticket(s) while push this PR first?

whit537:
Let's drop from the database as well to avoid taking on more tech debt.

nobodxbodon:
> Let's drop from the database as well to avoid taking on more tech debt.

Could you clarify please? Is it a quick delete in code or some db management task?

whit537:
@nobodxbodon Fairly quick. There should be some docs in the README. Basically a `sql/branch.sql` like this (untested):

```
BEGIN;
    ALTER TABLE participants DROP COLUMN bitcoin;
END;
```

nobodxbodon:
Is there any compelling reason to do this? If not IMO it's better to keep data than deleting them permanently.

whit537:
The reason would be to avoid carrying around dead weight. "Wait ... why is this in here? Do we actually use this column anywhere?"

JessaWitzel:
I agree with deleting the data from the database. Clean code that makes sense == good.

nobodxbodon:
@whit537 could you review please?

sql/branch.sql
```diff
@@ -0,0 +1,3 @@
+BEGIN;
+    ALTER TABLE participants DROP COLUMN bitcoin;
```
whit537: Are you sure this is right? I made this up (that's what "untested" meant in my [comment](https://github.com/gratipay/gratipay.com/pull/4187#issuecomment-272965641)), and now that I look I don't find a `bitcoin` column in `participants`. I'm actually surprised that Travis is okay with this.

mattbk:
Bitcoin is listed in the `exchange_routes` table along with the others. See https://github.com/gratipay/gratipay.com/pull/4187/files#diff-34d1c11dd5c1d2842284d39e89497ba3L671 as an example.

sql/branch.sql
nobodxbodon: Obviously it's not. reverted in 7bdd6b6.

whit537:
> Bitcoin is listed in the `exchange_routes` table along with the others.

Ah, gotcha. That makes sense, since we used to support actually moving money via that route. In that case we should definitely _keep_ the historical data since we don't want to corrupt participant history pages.

nobodxbodon:
@whit537 I suppose it's ready for review now?

Issue #557, opened 2013-01-29
Email notify users when they are marked suspicious


whit537:
Send an email notification when possible to users when we mark them suspicious.

----------------

## Original

This might be controversial.

I want to have a page at `/about/fraud/untrusted/` which simply lists all participants with  `is_suspicious is true`. As our fraud system evolves (#556), so will this listing. The principle is that untrusted users are announced as such so that false positives have a chance to discover this and attempt to redeem themselves. We can't be like PayPal, an infuriatingly opaque black box if you're a false positive.
##

mattbk:
> Perhaps, we also notify the user (if possible)

+1 to this. Recently interacted with a user at [FD](https://gratipay.freshdesk.com/helpdesk/tickets/5953) who couldn't log in and wasn't notified why--nothing happened when the user tried to log in. Took me a while to think of checking `is_suspicious`. Showing a message would help.

mattbk:
> Perhaps, we also notify the user (if possible) and put a note on their
profile?

If you view the profile when not logged in, it has a pink overlay at the top and the word "Suspicious" below the avatar. That doesn't tell people what to do, however. And since they can't log in, showing them a notification doesn't help.

whit537:
> Recently interacted with a user at FD who couldn't log in and wasn't notified why--nothing happened when the user tried to log in.

For the record, the user [turned out](https://github.com/gratipay/inside.gratipay.com/issues/890#issuecomment-261123662) to be a true positive. ;-)

whit537:
> Perhaps, we also notify the user (if possible) and put a note on their profile?

As discussed at https://github.com/gratipay/gratipay.com/pull/4188#issuecomment-262073906, I like this solution better than the listing of suspicious accounts, because a) it's much less work for those affected to discover that they're affected, and b) it's more respectful of user privacy. A public listing would only seem to make sense if we're trying to warn others about these users or something like that, but that doesn't seem necessary since suspicious users can't do anything on the site (so there's nothing to watch out for).

mattbk:
I'm fine with that, but I think there should be a message (not just a note on the profile) when someone tries to log in and are denied because they're suspicious. My first inkling in such a case would be that the login form is broken, not that I should go look at my own profile page to see what's up.

whit537:
>I'm fine with that, but I think there should be a message (not just a note on the profile) when someone tries to log in and are denied because they're suspicious.

[...]

PR #4188, opened 2016-11-12
Make a list of suspicious accounts

mattbk:
Part of #557. This covers the logic; it’s mostly a duplicate of
`dashboard/index` with the buttons removed (admins can mark/unmark
`suspicious` on the profile itself).

Obviously needs to be put into a site template for ~users to see, and
only needs the list, not the other frame.


mattbk:
`/about/fraud/suspicious`
screen shot 2016-11-12 at 5 56 48 pm

Haven't added to a menu. The other page in `/fraud` isn't in a menu either, AFAICT.

mattbk:
Almost deleted the dashboard version, but I figure that's something that could be modified in other ways.

mattbk:
I suppose I need to add a menu item before this can be reviewed.

mattbk:
We don't have a root page for `/fraud`; what would we put there? Maybe suggestions on how to let us know about potential fraud, including a link to http://inside.gratipay.com/howto/review-accounts? This is the type of thing that the community can pitch in for.

That would probably make the suspicious list a tab on the fraud page.

mattbk:
image

image

mattbk:
Ready for review, @whit537, @kaguillera et alii.

whit537:
Nav layout breaking for me when bolded:

screen shot 2016-11-21 at 4 18 15 pm

mattbk:
> Nav layout breaking for me when bolded:

Would "Fraud" be sufficient?

whit537:
I think conceptually it could be grouped with Security and also AML (maybe under "Safety"?), _but_ ...

The primary [motivation](https://github.com/gratipay/gratipay.com/issues/557#issue-10413440) here is to give suspicious users some way of discovering that they've been marked suspicious, correct?

> The principle is that untrusted users are announced as such so that false positives have a chance to discover this and attempt to redeem themselves. We can't be like PayPal, an infuriatingly opaque black box if you're a false positive.

Is a listing of suspicious accounts the best way to do that? How will suspicious users know to look there? I realize the suspicious listing was my suggestion, but https://github.com/gratipay/gratipay.com/issues/557#issuecomment-12837123 seems like a better solution to the problem:

> notify the user (if possible) and put a note on their profile

That way the suspicious user doesn't have to work so hard to discover the problem.

I think we should abort this PR and go for the notify/note solution. 🐭

mattbk:
> I think we should abort this PR and go for the notify/note solution. 

Fine by me. I learned some things, at least.

Issue #3121, opened 2015-01-15
admins should be able to search everyone


whit537:
`is_searchable` should only apply to public/user search. Hidden users should still be searchable by admins. (I tried finding someone during review on #3118.)
##

kzisme:
Fairly certain this issue relates [here](https://github.com/gratipay/gratipay.com/blob/f9749d98d2853701968adf437e9436deca1741aa/www/search.spt)

rohitpaulk:
Yep, all the queries on that page seem to filter participants by `is_searchable`. We could make those constraints dynamic, and relax them depending on whether the logged in user is an admin or not.

mattbk:
PR is #4190, ready for review.

mattbk:
This accidentally made it in. Explanation in https://github.com/gratipay/gratipay.com/pull/4190#issuecomment-269891592.

PR #4157, opened 2016-10-23
Revise site copy to keep up with ToS changes

whit537:
Part of https://github.com/gratipay/inside.gratipay.com/issues/432.

PR #4190, opened 2016-11-13
Allow admins to search unsearchable users

mattbk:
Closes https://github.com/gratipay/gratipay.com/issues/3121.


nobodxbodon:
Could you build the query in the block above (16-26) conditionally to avoid duplicated code?

nobodxbodon:
I was in a hurry. In case it wasn't clear enough, something like (excuse my python grammar):

```
    if action in (None, 'search_usernames'):
         results['usernames'] = website.db.all("""
             SELECT username, avatar_url, similarity(username, %(q)s) AS rank
               FROM participants
              WHERE username %% %(q)s
                AND claimed_time IS NOT NULL
                """ 
               + (user.ADMIN ? "AND is_searchable" : "")
               + " AND NOT is_closed"
           ORDER BY rank DESC, username
              LIMIT 10
         """, locals())
```

whit537:
Yay for Python from @nobodxbodon! 💃 🐍

mattbk:
I get what you're saying, just hitting a learning curve on how to work your example in so it works.

mattbk:
I was confused because I didn't realize the `+` signs were literal.

```
    if action in (None, 'search_usernames'):
         results['usernames'] = website.db.all("""
             SELECT username, avatar_url, similarity(username, %(q)s) AS rank
               FROM participants
              WHERE username %% %(q)s
                AND claimed_time IS NOT NULL
                """
                +("" if user.ADMIN else " AND is_searchable")
                +""" AND NOT is_closed
           ORDER BY rank DESC, username
              LIMIT 10
         """, locals())
```

mattbk:
Something is still wrong with this, I need to look at it again.

mattbk:
Okay @nobodxbodon, ready for review again with a simpler diff. Thanks for the help.

www/search.spt
```diff
@@ -14,16 +14,17 @@ if query:
     q = strip_accents(query)
 
     if action in (None, 'search_usernames'):
-        results['usernames'] = website.db.all("""
```
whit537: Why change the indentation here? It makes the diff harder to read.

www/search.spt
```diff
@@ -14,16 +14,17 @@ if query:
...
-          ORDER BY rank DESC, username
-             LIMIT 10
-        """, locals())
+         results['usernames'] = website.db.all("""
+             SELECT username, avatar_url, similarity(username, %(q)s) AS rank
+               FROM participants
+              WHERE username %% %(q)s
+                AND claimed_time IS NOT NULL
+                """
+                +("" if user.ADMIN else " AND is_searchable")
```
whit537: Hmm ... could we find a way to do this without string interpolation? (Tbh this is the achilles heel of our db wiring; it's bad practice to construct SQL from strings because of the danger of SQL injection, but it's hard to avoid without a full-blown SQL model in Python, aka SQLAlchemy.)

whit537:
Needs tests.

www/search.spt
mattbk: Accident.

www/search.spt
mattbk: The first commit didn't work like that, FWIW.

mattbk:
After thrashing about, I cleaned up the indents.

mattbk:
This functionality seems to exist...investigating.

- String interpolation version introduced in https://github.com/gratipay/gratipay.com/commit/5ff5121cf7ea9651792d6326a30583505bad7956
- Finessed in https://github.com/gratipay/gratipay.com/commit/888b96b4ac0c03b03fe4281fdaa9c6e5122bdc61 
- Finalized in  https://github.com/gratipay/gratipay.com/commit/52c02fb9e2217b854e31564c47fa5f523f74a784

There's a commit (6230342a4e4e498f09f85bdb33e54f01b2927b00) in #4157 that is even called "Allow admins to search unsearchable users," but it doesn't have the string interpolation version--because it was introduced earlier, in https://github.com/gratipay/gratipay.com/pull/4157/commits/a277c5d38d74005ed7437a7da65fbf95fc111915. I image this made it in there after I neglected to properly stash files when switching branches.

As such, I am closing this PR and the original ticket.

whit537:
!m @mattbk

PR #4200, opened 2016-11-22
Upgrade Sentry client

whit537:
Closes #4191.

Confirmed working with a "[Gratipay.com Sandbox](https://sentry.io/gratipay/gratipaycom-sandbox/)" project DSN in `local.env`.

screen shot 2016-11-21 at 10 58 42 pm

PR #4205, opened 2016-11-24
Fix early error handling

whit537:
I believe all ~~three~~ four of these are related:

1. https://sentry.io/gratipay/gratipay-com/issues/82062971/
1. https://sentry.io/gratipay/gratipay-com/issues/83182753/
1. https://sentry.io/gratipay/gratipay-com/issues/184941452/
1. https://sentry.io/gratipay/gratipay-com/issues/199760784/


whit537:
In all three of those cases, in the context of the `delegate_error_to_simplate` frame, `request` is the empty string(!). (BTW, this is displayed better under the new client, as `''` instead of ``; #4200.) This surprised me. There is a [test](https://github.com/AspenWeb/pando.py/blob/0.42/aspen/algorithms/website.py#L139-L140) at the top of that function for `request is None`, but how can `request` be the empty string?

Well, it turns out that we _do_ initialize `Request` [thus](https://github.com/AspenWeb/pando.py/blob/0.42/aspen/http/request.py#L200), so that some sort of other error could—and seemingly does—result in `request` as the empty string inside `delegate_error_to_simplate`.

I think the fix here in Gratipay is probably a function immediately before `delegate_error_to_simplate` that converts `''` to `None`, with a reticket in Aspen (Pando). However, we should write a test first. Towards that: (1) and (3) are both CRLF injections, and (2) is a bad redirect location, as indicated by the `err` key in the `/app/error.spt` frame context.

whit537:
And I still don't quite understand how `request` is ending up the empty string. If [`Request.from_wsgi`](https://github.com/AspenWeb/pando.py/blob/0.42/aspen/algorithms/website.py#L53) were to raise, I'd expect `request` to go missing from the context.

whit537:
tfw google for `curl crlf injection` and find https://github.com/gratipay/security-qf35us/issues/1 😁

whit537:
I'm trying:

```
curl -I http://localhost:8537/%0dSet-Cookie:Xsrf_token=xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx;
```

whit537:
I am triggering the CRLF protection, but I'm not seeing the empty string request behavior.

whit537:
Skunked. :-/

whit537:
Rebased, was b32bc05d645e2d41509899a2cec3c2293b4c9a45.

whit537:
Only the last of the original four errors is still available in Sentry:

https://sentry.io/gratipay/gratipay-com/issues/199760784/

whit537:
I give up. 😞

PR #4210, opened 2016-11-30
Rip out READMEs

whit537:
Post-#4151.


whit537:
Oops. Can't change the base more than once? :-/

PR #4211, opened 2016-11-30
Rip out READMEs after all

whit537:
Post-#4151. READMEs are too much work to keep up with and keep secure, especially when we want to add 30+ more package managers.

PR #4212, opened 2016-11-30
Stub out /on/npm/foo/ pages

whit537:
The bare minimum left from #4151 after #4211. Includes #4211.

screen shot 2016-12-01 at 9 40 41 am


whit537:
We're going to get spammy security reports about content spoofing.

whit537:
Ready for review, @aandis. Pretty basic. If you think it's not worth merging yet lmk.

whit537:
💃

PR #4156, opened 2016-10-22
Rename Teams to projects

whit537:
Part of https://github.com/gratipay/inside.gratipay.com/issues/432.

Issue #3433, opened 2015-05-15
bring back take-what-you-want


whit537:
Reticketing from #3414 ...

We lost our team distribution algorithm in #3414. We should bring it back by upgrading `transfer_takes` to `process_payroll` in `payday.py`, but first we should ...
- [x] Determine the relationship between Gratipay Teams and their members (https://github.com/gratipay/inside.gratipay.com/issues/242)
- [x] Build a vault to store PII (https://github.com/gratipay/gratipay.com/issues/3504)
- [x] Bring back take-what-you-want _for Team Gratipay_ (#3994)
- [ ] Decide to prioritize this.

#### Notify
- [ ] https://gratipay.freshdesk.com/helpdesk/tickets/2184
- [ ] https://gratipay.freshdesk.com/helpdesk/tickets/2185
- [ ] https://gratipay.freshdesk.com/helpdesk/tickets/2234
- [ ] https://gratipay.freshdesk.com/helpdesk/tickets/2312
- [ ] https://gratipay.freshdesk.com/helpdesk/tickets/2301
- [ ] https://gratipay.freshdesk.com/helpdesk/tickets/2366
- [ ] https://gratipay.freshdesk.com/helpdesk/tickets/2427
- [ ] https://gratipay.freshdesk.com/helpdesk/tickets/3126
- [ ] https://gratipay.freshdesk.com/helpdesk/tickets/6699
##

whit537:
One way to address the question of the relationship between teams and their members would be to add a "percentage withholding" to the `takes` table, which only the team owner could set. On the one hand that would make it clear to team owners that this is something they need to think and decide about, and on the other it leaves it flexible in terms of specific arrangements. We can't wait to model all of the world's employment law before we re-release the payroll feature.

whit537:
> `takes` table

Erm, the `payroll` table, rather.

whit537:
Setting to blocked on https://github.com/gratipay/inside.gratipay.com/issues/242.

whit537:
It's a lot of overhead to set up a company, even if we *can* pull off "[company as a service](https://github.com/gratipay/inside.gratipay.com/issues/324)." We want to ensure that Teams on Gratipay create open work for people, but we don't want to overburden Teams that are just starting out. We should consider allowing Teams to receive up to a certain amount per week before requiring them to go through identity verification (#988) and turn on payroll. Maybe $1,000 per week?

whit537:
[...]

PR #4218, opened 2016-12-02
Rename members to collaborators

whit537:


mattbk:
I might be able to help with this if I knew exactly what the task was.

whit537:
The idea was to harmonize our nomenclature between front-end and backend. However, I'm not sure this is a priority since this only applies to projects with distributing, and Gratipay is the only one of those we have right now. I think we should close this and revisit the matter when we get to https://github.com/gratipay/gratipay.com/issues/3433.

whit537:
#4156 is actually more important—and more difficult.

Issue #1133, opened 2013-07-13
deploy branches to heroku automatically


whit537:
We need this to turbo-charge our development workflow (@corytheboyd and @warreng called it; props). The idea is that any branch in GitHub gets its own public deployment, with its own database with fake data. This should be a subcommand of `gittip`.
##

whit537:
+1 from @techtonik at https://github.com/gratipay/gratipay.com/pull/3718#issuecomment-135723108.

> Should we have a dev.gratipay.com with the WIP site version that is automatically pushed on commits from master?

whit537:
> For each new GitHub pull request (“PR”), Heroku will spin up a disposable “Review App” with a unique URL for the development team to review and test. Now, instead of guessing what the new code might do, reviewers — including designers, testers, and PMs — can receive a link and actually try it out in a browser, before the code reaches staging or production environments.

http://blog.heroku.com/archives/2015/9/3/heroku_flow_pipelines_review_apps_and_github_sync#review-apps-in-the-flow

whit537:
> Today, we are happy to announce the graduation of [Heroku Review apps](https://devcenter.heroku.com/articles/github-integration-review-apps) from an exceptionally popular beta to being generally available to all Heroku users.
>
Review apps are the instant, disposable Heroku app environments that can spin up automatically with each GitHub pull request. They allow developers and their teams to automatically build and test any pull request, updated at every push, at a temporary, shareable URL. When the pull request is closed or merged, the Review app is deleted. GitHub users are notified of all this, right in the pull request web interface.

https://blog.heroku.com/archives/2016/4/18/heroku_review_apps_GA?

nobodxbodon:
The page in the last comment seems to be pulled back. I guess the service is still beta. Will it be free for our purpose?

whit537:
Looks like they have a broken redirect on the blog post. Here's the correct link: https://blog.heroku.com/heroku-review-apps-ga.

whit537:
I've updated the link in my comment.

nobodxbodon:
As the issue isn't user-oriented, reticket to inside.

PR #4228, opened 2016-12-09
Fix Chrome autofill

### LINKED ISSUE #4041: Credit Card input field doesn't play nicely with Chrome's credit card auto fill feature
Chrome has a secure way to store credit card info and auto-fill it. The input field (under /routes/credit-card) doesn't seem to play nicely with it. When I start typing in the field, I expect Chrome to suggest saved credit cards. In the case of Gratipay, the suggestion doesn't happen unless I type a little and then hit backspace.

mattbk:
Closes #4041. I’m guessing this will take several iterations to get right— ~~testing locally is difficult because Chrome won’t autofill unless the connection is secure.~~ Yes, but we get a popup telling us it's insecure, and that's almost as good.


mattbk:
I wonder if a self-signed certificate would work?

mattbk:
I am unable to create a self-signedcertificate that works for me, but I think that's beside the point. The behavior is the same, even with suggested changes: No autofill message comes up until typing at least two numbers and then hitting backspace.

I will research further.

mattbk:
Aha! If I disable Javascript, it wakes right up.

mattbk:
- I can still submit with JS off.
- I still get popup errors on submit when I don't have an expiration date or CVV.
- I do _not_ get an error if the number is bad or doesn't match a known merchant format ("Credit card type is not accepted by this merchant account...").


Cf. http://stackoverflow.com/questions/7248488/what-would-cause-chrome-autofill-to-stop-working

Hmmmm, related? We're using Braintree's validation in `routes.js`: https://bugs.chromium.org/p/chromium/issues/detail?id=585505

Or as simple as not working at all for Braintree? [Credit Card Autofill Does Not Work](https://github.com/braintree/braintree-web/issues/239).

mattbk:
Doesn't seem related to the ZIP and hidden Braintree fields being non-cc fields (as suggested somewhere). I'm going to bed.

mattbk:
Even if I drop all the way down to 
```
        <form id="credit-card">
                <input id="card_number" required autocomplete="cc-number" />
            <button class="selected larger" id="save" type="submit">{{ _("Save") }}</button>
        </form>
```
the behavior is the same. Is there some master JS that is messing with all form data?

mattbk:
It's this function in `routes.js`:
```
    Gratipay.routes.cc.formatInputs(
        $('#card_number'),
        $('#expiration_month'),
        $('#expiration_year'),
        $('#cvv')
    );
```

If I change or remove it (or just the `#card_number` form input ID, the autocomplete works...but it won't let me put the right values into MM/DD or CVV. Sometimes it won't even let me put two digits in the date fields, just one. Weird.

mattbk:
Well, it's this one: `Gratipay.routes.cc.formatInputs()`.

Aha! If I just don't run the CC number through that function, autocomplete works.

Let's commit that and see.

mattbk:
There are still some `cardNumberInput` calls in that function, so should probably see how important they are versus making autocomplete work. Autocomplete seems only to work if we don't mess with the number.

mattbk:
It is still checking whether the card is American Express or not, because it's telling me when to use 3 or 4 digits for CVV...

Also not validating numbers vs letters.

nobodxbodon:
May I ask on which instance you are testing? "review app" mentioned [here](https://github.com/gratipay/gratipay.com/issues/1133)?

mattbk:
That's a question for @whit537, I have no idea. I'm checking this locally.

mattbk:
@nobodxbodon, I honestly don't know what you're asking. :smile:

whit537:
I think the answer is that you're "checking this locally"?

nobodxbodon:
@mattbk actually I was referring to your initial comment
> testing locally is difficult because Chrome won’t autofill unless the connection is secure

So I wondered if you tested the commits on some staging/dev instance instead of locally.

mattbk:
> So I wondered if you tested the commits on some staging/dev instance instead of locally.

I didn't know of anything like that already set up, so I played around with it until I realized that the message that autocomplete won't work actually follows the same behavior. So whether it will actually autocomplete the card doesn't matter, all that should matter is that the message comes up on the first number entered, rather than after n numbers and then a backspace.

I'm not convinced yet that the card number can't be passed through at least some of the subroutines in `Gratipay.routes.cc.formatInputs()`, in which case I need to dig deeper into that function and what it does.

mattbk:
From my experiments, I don't think we can play nice with autocomplete _and_ validate the way we do now. However, https://github.com/braintree/card-validator looks like something we could use.

whit537:
Seems reasonable to me. 👍

mattbk:
Is anyone interested in taking this on? Running out of time in my day to give this the attention it deserves.

(@whit537, this doesn't include you jumping in and wasting another day. I'm trying to focus on only opening things I know I can finish.)

mattbk:
@EdOverflow is there any security reason for us to not support Chrome Autofill, aside from the user storing their credit card info in Chrome itself?

EdOverflow:
@mattbk I usually don't encourage people to use the Chrome Autofill, because of the following: https://github.com/anttiviljami/browser-autofill-phishing/blob/master/readme.md

That said, I don't think it would hurt to implement it on Gratipay. I will take a look when everything is done.

mattbk:
Rebased on master?

whit537:
Ish? Needs work, I think. Want me to try to clean it up?

mattbk:
> Want me to try to clean it up?

If you have a chance. I wasn't going to go any further today, but apparently my drive-by rebasing needs work.

whit537:
I think you maybe did whatever @kaguillera sometimes does. :-)

Look okay now? (Was c88281f1a4871d49b59e29a978a50ef49803ec37.)

whit537:
@mattbk Is this ready for review?

mattbk:
No, I never learned how to integrate  https://github.com/braintree/card-validator. :disappointed:

mattbk:
...and without that, the way to fix is to remove validation during card entry. IMHO this is fine...

mattbk:
I looked at this again. I think it's ready for review. Essentially it doesn't pass the card number through our internal validation, which lets Chrome begin filling in the card number immediately.

mattbk:
image

PR #4221, opened 2016-12-02
Relax open work requirement

whit537:
Ever since the Gratipocalypse, we've required the "Teams" receiving money on Gratipay to go through an application and approval process. One of our main criteria for approval has been that Teams provide what we called "open work":

> Open Work means that the Team provides a clear path for any individual to voluntarily begin contributing to the Team's work and to share in any revenue the Team generates.

Experience has shown this to be too restrictive. We still want to _encourage_ and _emphasize_ open work on Gratipay, but we don't want to _require_ it. The purpose of this project is to relax the open work requirement.

--------------------

✈️  **This is the flight deck for the [Relax Open Work Requirement](https://github.com/orgs/gratipay/projects/5?fullscreen=true.) project.** ✈️ 

--------------------

### Todo

- [ ] update the user interface
  - [x] merge #4117
  - [x] merge #4214
  - [x] merge #4130
  - [x] merge #4157
  - [ ] merge to `master`
- [ ] update the review repo
  - [ ] rename the `team-review` repo to `project-review`
  - [ ] update `TEAM_REVIEW_REPO` in the Heroku environment
  - [ ] merge https://github.com/gratipay/team-review/pull/281
- [ ] update Inside Gratipay
  - [ ] merge https://github.com/gratipay/inside.gratipay.com/pull/861
- [ ] blog post: https://gratipay.news/from-teams-to-projects-45c46718507b
  - [ ] update link to `terms-of-service.md` diff
  - [ ] publish
- [ ] announce on Twitter
- [ ] notify owners of projects that are accepted or under review
- [ ] put notice on website: #4229

PR #4230, opened 2016-12-12
Fix link in About page

whit537:
Bug from #4221.

We don't have Markdown here; use HTML.


whit537:
+1 from @jdorfman in [slack](http://gratipay.slackarchive.io/gratipay/)! 💃

PR #4231, opened 2016-12-13
Replace missed instances of "team" with "project"

mattbk:
On the frontend at least.

### DIFF
--- a/emails/team-approved.spt
+++ b/emails/team-approved.spt
@@ -1,6 +1,6 @@
-{{ _("Team Application Approved!") }}
+{{ _("Project Application Approved!") }}
 [---] text/html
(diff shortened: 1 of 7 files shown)


mattbk:
Ready for review. @JessaWitzel?

JessaWitzel:
I approve yours @mattbk ! I also added two of my own. Do we need to be concerned about the translation templates?

mattbk:
I don't think so. That seems to be handled elsewhere: https://github.com/gratipay/gratipay.com/tree/team-to-project-misc/i18n

mattbk:
Nice catch on the emails.

whit537:
💃

PR #4234, opened 2016-12-15
Add a "Featured" tab on the homepage

whit537:
Following up from [slack](http://gratipay.slackarchive.io/gratipay/-/1481739661.000615/1481839852.000722/1481839754000712/) ...

Right now the homepage lists all approved projects by default, most recently created first. This doesn't give a great first impression, because we end up seeing mostly little projects nobody has ever heard of or is giving to. We don't want to _hide_ such projects per se (we want to be careful of "the rich get richer"), but when people are hitting Gratipay for the first time, they want to see, "Who is on here that I recognize?"

I had an idea to incrementally improve the situation without _too_ much trouble (hopefully), and it seemed to @JessaWitzel like a good next project. So here we are! 💃

Basically I propose that we add a "Featured" tab in addition to the others on the homepage. It would be the default tab, and would show ten projects. Seven of the projects would be drawn randomly from a pool of the most popular projects (maybe `ngivers >= 5`, I think? We're so sad), and three would be taken randomly from amongst the rest. (Gratipay itself should never be a featured project.) Once we have those 10, I think maybe we randomize the order we list them on the Featured tab, for freshness and because otherwise how would we sort them without burying the new/unpopular projects at the bottom?

Apart from an icon (blue ribbon?) for Featured to match the other tabs, I think we should be able to do this only changing `www/index.html.spt` (plus tests?), because we already load all of the teams every time someone hits the homepage (that's an optimization problem we can worry about later).

--------------------------

# Mockup

image

--------------------------

# We Are Sad

```
gratipay::MAROON=> select name, nreceiving_from from teams where nreceiving_from >= 5 order by nreceiving_from desc;
┌──────────────────────────┬─────────────────┐
│           name           │ nreceiving_from │
├──────────────────────────┼─────────────────┤
│ Gratipay                 │             265 │
│ sqlalchemy               │              21 │
│ DevDocs                  │              19 │
│ CIDER                    │              18 │
│ Nuvola Player            │              17 │
│ jsbin                    │              16 │
│ PylonsProject            │              13 │
│ Internet of Ownership    │              13 │
│ Org-mode                 │              12 │
│ Magit                    │              11 │
│ Shields                  │              11 │
│ WDRL                     │               9 │
│ Chocolatey               │               9 │
│ RuboCop                  │               9 │
│ webpack                  │               8 │
│ Mojolicious              │               8 │
│ catapultpgh              │               7 │
│ Pillow                   │               7 │
│ Font Awesome             │               7 │
│ PerlWeekly               │               7 │
│ sudo room                │               7 │
│ simplytest.me            │               6 │
│ Reddit Enhancement Suite │               6 │
│ weblog.sh                │               6 │
│ dokku                    │               5 │
│ SublimeLinter            │               5 │
│ OpenLP                   │               5 │
└──────────────────────────┴─────────────────┘
(27 rows)
```


whit537:
Does that give you enough to get started, @JessaWitzel? Anything jump out at you that needs to be better-specified?

whit537:
Maybe constrain to only list on per owner among the 10?

JessaWitzel:
If this is all about the projects and being able to support a project and the team working on that project what benefit is there to limiting it to one project per owner? Especially if it's random. Are we worried how it will look if 5/10 of them are owned by one person?

whit537:
> Are we worried how it will look if 5/10 of them are owned by one person?

Yeah, that was my only thought. I'm not proposing we limit the whole website to one project per owner, just the rotating featured 10 on the homepage. Eh?

JessaWitzel:
I was originally thinking about resetting daily but I like the idea of resetting weekly. I have no clue how to do that but will explore once I get this tab working. :)

whit537:
Resetting either weekly or daily is going to be a fair bit more work, because we'll need to store the featured listing somewhere other than in memory. My suggestion would be that we get it working on this PR with generated-every-request, and work on persisting the list in a separate PR.

whit537:
[slack](http://gratipay.slackarchive.io/gratipay/-/1483909182.002289/1484010165.002396/1484007448002380/)

JessaWitzel:
I am still having a hard time making enough fake data to test this properly so would LOVE a set of eyes on this before I move forward and some feedback. The icon isn't right, I'm still trying to figure that out. And I haven't written specific tests yet. @mattbk  or @whit537 or anyone would you mind taking a look and giving me feedback?

mattbk:
Is it only showing the tab when there are enough projects that match certain criteria?

JessaWitzel:
@mattbk Yes. I had to make it that way because of the way our testing environment works but also I think it's smart because we are open source so if someone wants to copy our tabs situation but has less teams it will still work for them.

nobodxbodon:
@JessaWitzel I don't think the tab should be hidden when there are fewer than 7 popular + 3 unpopular. IMO it's shown when there's at least one and we show *up to* 7 popular and 3 unpopular.

JessaWitzel:
@nobodxbodon Well if there is only one or two teams then they will all just show up on the Approved tab. What point is a featured tab if it just shows the Approved tab in a random order?

nobodxbodon:
As we have much more than 2 teams now, I suppose you are thinking of reusing it by some other projects. I'd say suit our needs first, and I don't think it's a big task for other project maintainers to customize it to their needs.

Besides, in case we have only 5 or 6 popular, is it better to still show the 'featured' tab?

JessaWitzel:
Ah, I see what you mean. Like if we have 30 projects but only 5 popular. Ok thanks! I will make that change. This works for us currently but I think it makes sense to make that change in case projects fall. I can do that tonight. Any other feedback?

mattbk:
As a workaround for fake data, you can just modify `fake_data.py` locally to change `nreceiving_from=3` to whatever you want.

nobodxbodon:
To make sure we are on the same page, here are some cases in my mind. Could you check?

| existing projects | result |
| ------------ | ------------ |
| no project at all | tab is hidden |
| 1 unpopular project only | tab shows the 1 project |
| 1 popular project only |  tab shows the 1 project |
| 2 populars, 2 unpopulars | tab shows the 2 populars, 2 unpopulars |
| 2 populars, 10 unpopulars | tab shows the 2 populars, 8 unpopulars |
| 8 populars, 1 unpopulars | tab shows 7 populars, 1 unpopular |
| 10 populars, 10 unpopulars | tab shows 7 populars, 3 unpopular |

JessaWitzel:
@nobodxbodon The way I wrote it if the total number of approved projects is less than 10 there is no featured tab because it will just be a randomized version of the approved tab. If, however there are more than 10 projects total then the rest of your logic is true. Does that work?

nobodxbodon:
@JessaWitzel sounds good to me. Thanks.

JessaWitzel:
@mattbk  yes! I want it to be a star. :( I totally looked up the icon and used the right SVG code and everything.

JessaWitzel:
I requested reviews but should probably write tests, yeah?

www/index.html.spt
```diff
@@ -17,16 +22,39 @@ teams = website.db.all("""
...
+unpopular = []
+for project in teams:
+    tabs[project.status]['teams'].append(project)
+    if project.status == 'approved':
+      if project.nreceiving_from >= 5:
+        popular.append(project)
+      else:
+        unpopular.append(project)
+
+if len(teams) < 10:
```
nobodxbodon: And seems here 10 should be compared with number of approved projects?

www/index.html.spt
```diff
@@ -17,16 +22,39 @@ teams = website.db.all("""
...
+    tabs.pop('featured')
+else:
+    popular_sample_size = min(len(popular), 7)
+    unpopular_sample_size = min(len(unpopular), 10-popular_sample_size)
+    if popular_sample_size == 0:
+        featured_projects = random.sample(unpopular, unpopular_sample_size)
+    elif unpopular_sample_size == 0:
+        featured_projects = random.sample(popular, min(10, len(popular)))
+    else:
+        featured_projects = random.sample(popular, popular_sample_size) + random.sample(unpopular, unpopular_sample_size)
```
nobodxbodon: Below seems to be an alternative of L48-55 block:
```
popular_sample_size = min(10 if (len(unpopular) == 0) else 7, len(popular))
unpopular_sample_size = min(len(unpopular), 10-popular_sample_size)
featured_projects = random.sample(popular, popular_sample_size) + random.sample(unpopular, unpopular_sample_size)
```
Please add tests first, if you decide to do some refactoring of this part, to avoid changing original logic.

nobodxbodon: @mattbk I'm not certain where the test cases should go. Could you advise?

JessaWitzel: @whit537 indicated they should be in test_homepage.py on Slack

JessaWitzel: Thank you for the review!

whit537:
I see featured projects! 😱 

screen shot 2017-01-18 at 6 16 11 am

whit537:
> I want it to be a star. :(

The way the icon font works is that we have multiple versions of the same font in four formats—`eot`, `svg`, `ttf`, `woff`—and then some SCSS to [switch between them](https://github.com/gratipay/gratipay.com/blob/93a0ea51f78a32b76b969bd7cd9bdc8c64de91f6/scss/mixins/icons.scss) depending on browser and operating system support. Looks like so far we've changed the SVG one but not the others?

Here's a [somewhat crustabulous link](http://gratipay.slackarchive.io/gratipay/-/1484607178.002808/1484607402.002833/1484607178002808/) to slackarchive where I started braindumping about icomoon. Now that my memory is jogged I believe what I did was export from IcoMoon and then move files into place under `www/assets/fonts` and then manually edit the `icons.scss` file based on whatever was in the ZIP that IcoMoon gives us. 'Nuff to go on?

www/index.html.spt
```diff
@@ -17,16 +22,39 @@ teams = website.db.all("""
...
-    tabs[team.status]['teams'].append(team)
+popular = []
+unpopular = []
+for project in teams:
+    tabs[project.status]['teams'].append(project)
+    if project.status == 'approved':
+      if project.nreceiving_from >= 5:
+        popular.append(project)
+      else:
+        unpopular.append(project)
```
whit537: Four-space indents, [please](http://inside.gratipay.com/howto/format-source-code). 🙇

www/index.html.spt
```diff
@@ -17,16 +22,39 @@ teams = website.db.all("""
...
+
+if len(teams) < 10:
+    tabs.pop('featured')
```
whit537: This works. If you don't need to use the return value from `pop` then you can also do:

```python
del tabs['featured']
```

Just a little Python tip in case you didn't already know ... no need to change this here. :-)

www/index.html.spt
whit537: > a utility method like `get_featured_projects(popular, unpopular)`

A good place for it might be `gratipay/utils/__init__.py`, right below `get_team`.

www/index.html.spt
whit537: Let's go with `test_utils.py` since `utils` seems the natural place to add the function we're talking about. I've added dd0b8ce81863b9e7b8ce75cc4536cba36fd12935 to bring the `test_utils.py` file up to modern standards of organization (one class per function). Hopefully that gives you enough to go on to add tests for the new function

whit537:
Took a pass through. Good progress, @JessaWitzel! Here's what it looks like with production data:

screen shot 2017-01-18 at 6 49 41 am

whit537:
How about we sort the featured tab by amount descending? Do you think that's a good idea?

Also, seems we'll need to do something about the float drop on the tabs. Maybe it's enough to change "Under Review" to "Reviewing"? Would we have to do that elsewhere as well?

mattbk:
I'd rather not sort by anything. I think sorting by receiving implies that this is the reason they are featured? 

!m everyone!

mattbk:
What can I do to help?

JessaWitzel:
Nothing yet @mattbk . Was marching on DC and today had a sale but tomorrow is Gratipay day and I will ping you if I have questions!

www/index.html.spt
JessaWitzel: I didn't know! I've used pop a bunch so that was just what came to me. Thanks!

JessaWitzel:
Ok I think I have fixed the things that need fixing except the standard naming which @mattbk nicely provided me a replacement for. I will write tests before I use that which means the next step is to write tests!

mattbk:
Do we need (n=10) on the featured tab? Also, I see a square icon now.

screen shot 2017-01-24 at 1 58 17 pm

JessaWitzel:
Oh yeah, I haven't fixed that because I don't really understand Chad's instructions. It's always been a square for me. We do not need n= 10 but I think it's a nicer look, for consistency's sake.

JessaWitzel:
Hey @whit537 . This is done except for the icon. 

"what I did was export from IcoMoon and then move files into place under www/assets/fonts and then manually edit the icons.scss file based on whatever was in the ZIP that IcoMoon gives us. 'Nuff to go on?"

I can't even get a ZIP from IcoMoon and don't really understand this process. Does someone wanna just fix the icon or merge as a square?

JessaWitzel:
Wayne seems to think that I need the project file and I cannot select all of our previous icons. Icomoon says "Note: Your changes and your projects only get stored in your browser's cache. Clearing your browsing data will result in losing your projects." Do you have a project stored somewhere that you can send me so that I can add this icon? @whit537

JessaWitzel:
Nevermind. :) Used the existing SVG file to import font icons and selected the one I needed to add. Now everyone can see how my brain works. This is ready for review.

whit537:
Woo-hoo!

Rebased, was 887e4f8ba23e72eada05255b843a52663701961d. Reviewing ...

whit537:
I kicked out #4314 and restarted the build.

whit537:
Rebased again, was 4fcaa25821ba867ae3303ba92b97738a5d718089.

whit537:
I see a star! ⭐️ 🔵  💃 

--------------------------

screen shot 2017-02-01 at 3 08 35 pm

whit537:
How about "Unreviewed" to keep the tabs on one line? cc459901b21c78926d9b8653a7768ce0b0c35b53

------------------

screen shot 2017-02-01 at 3 40 36 pm

whit537:
Do we want to include Gratipay in the pool of featured projects? Is that too forward?

whit537:
Or does it help communicate that Gratipay is funded on Gratipay?

JessaWitzel:
I think it should be like any other project, included when it is chosen at random.

mattbk:
> How about "Unreviewed" to keep the tabs on one line?
Does "In review" fit?

whit537:
I'm ready to merge and deploy when you are, @JessaWitzel @mattbk @TravisCI.

whit537:
> Does "In review" fit?

Probably. You think "Unreviewed" is that bad?

whit537:
"Unreviewed" is actually more parallel to "Approved" and "Rejected," isn't it?

whit537:
/me preps a tweet ...

> We've added a "⭐️ Featured" tab to https://gratipay.com/—mix of popular & up-and-coming projects. (Thanks to @/JessaCarpenter for the PR!)

whit537:
We have a +1 from @JessaWitzel in [slack](http://gratipay.slackarchive.io/gratipay/-/1485964960.004113/1485987872.004225/1485986043004203/).

You good, @mattbk? I'm thinking we can tweak the tab after the fact if need be ...

whit537:
Alright, I'm holding off until tomorrow. I was the last to commit, so I shouldn't be the one to merge.

@mattbk I believe the merge is yours, unless the tab name is a blocker for you.

mattbk:
> "Unreviewed" is actually more parallel to "Approved" and "Rejected," isn't it?

"Approved" and "Rejected" are end-members, "In Review" is transitional. Not a blocker for me.

JessaWitzel:
FIREWORKS!!

PR #4239, opened 2016-12-17
Update cryptography to 1.5.3

EdOverflow:
`cryptography` fixed a security issue in version 1.5.3:

> SECURITY ISSUE: Fixed a bug where HKDF would return an empty byte-string if used with a length less than algorithm.digest_size. Credit to Markus Döring for reporting the issue. CVE-2016-9243


whit537:
Thanks! I'm testing the build locally (it should build cleanly when the network is off, to guarantee that we've vendored all our deps). I'm hitting an error with the version of openssl I'm trying to link, but I think that's a local condition, debugging ...

whit537:
Deploying ...

whit537:
Thanks, @EdOverflow!

If you'd like to join our team more permanently, make an onboarding ticket as [described here](http://inside.gratipay.com/big-picture/welcome). We'd love to have you! :-)

EdOverflow:
I'll look into it. :)

PR #4242, opened 2016-12-19
Tweak the Record an Exchange page

whit537:


www/~/%username/history/record-an-exchange.spt
```diff
@@ -78,7 +78,6 @@ if request.method == 'POST':
 {% extends "templates/profile.html" %}
 {% block content %}
 
-<h2>Record an Exchange</h2>
```
whit537: We get an automatic `h1` from the `title` defined above. The `h2` here is redundant.

whit537:
Cool. Rebased (was 9869b2972d3da32f3e2d75659ff8fcb581011bd0) and ready to merge when you are @nobodxbodon! 👍

PR #4244, opened 2016-12-21
Implement Content-Security-Policy (CSP)

EdOverflow:
This solves the following report: https://hackerone.com/reports/90777

This policy allows:
- Resources from gratipay.com & all gratipay subdomains.
- Fonts from cloud.typography.com.

If you like we could allow GIFs from, for example, [giphy.com](https://giphy.com/). That way users can at least embed GIFs.

Concerning [report-uri.io](https://report-uri.io/), you will need to sign up for that and if you have any issues feel free to contact me.

### DIFF
--- a/gratipay/security/__init__.py
+++ b/gratipay/security/__init__.py
@@ -40,3 +40,9 @@ def add_headers_to_response(response):
     # https://www.owasp.org/index.php/List_of_useful_HTTP_headers
     if 'X-XSS-Protection' not in response.headers:
         response.headers['X-XSS-Protection'] = '1; mode=block'
+
+    # CSP - https://developer.mozilla.org/en-US/docs/Web/HTTP/CSP
+    # Allow resources from gratipay.com & all gratipay subdomains.
+    # Allow fonts from cloud.typography.com.
+    if 'content-security-policy' not in response.headers:
+        response.headers['content-security-policy'] = 'default-src \'self\' *.gratipay.com; font-src cloud.typography.com;'
--- a/gratipay/security/__init__.py
+++ b/gratipay/security/__init__.py
@@ -45,3 +45,11 @@ def add_headers_to_response(response):
     # Allow resources from gratipay.com & all gratipay subdomains.
     # Allow fonts from cloud.typography.com.
     if 'content-security-policy' not in response.headers:
+        response.headers['content-security-policy'] = ( 'default-src \'self\';' 
+                                                        'script-src assets.gratipay.com;'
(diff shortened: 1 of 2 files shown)


whit537:
Thanks, @EdOverflow, will review ...

whit537:
What will be the effect of this on existing legitimate image embeds in user statements? This will cause them to fail, yes?

Why do we need to allow all subdomains of gratipay.com?

EdOverflow:
> What will be the effect of this on existing legitimate image embeds in user statements? This will cause them to fail, yes?

Yep, this will block all existing external images. One option could be to allow some image hosting sites (imgur, giphy, etc.).

> Why do we need to allow all subdomains of gratipay.com?

We don't need to allow all subdomains. I thought this may come in handy for future use, but if you prefer I can specify assets.gratipay.com.

whit537:
> I thought this may come in handy for future use, but if you prefer I can specify assets.gratipay.com.

I think we should keep the set to a minimum and expand as needed. Any caching or other issues to consider if/when adding new subdomains in the future?

> Yep, this will block all existing external images. One option could be to allow some image hosting sites (imgur, giphy, etc.).

This is the bigger issue. We should probably start with a survey/census of existing usage: how many users would be affected? Are there a small set of domains that account for 99% of the usage? Can you do that from the outside, @EdOverflow? Or can you suggest a SQL query I can run for you against our database?

If there are enough users affected, then we should probably wait for [Camo](https://github.com/atmos/camo) ([H1-140387](https://hackerone.com/reports/140387))—unless we have other options? I don't like the idea of breaking images for a significant number of users for what amounts to a theoretical security risk.

EdOverflow:
> I think we should keep the set to a minimum and expand as needed.

I am in total agreement. The updated CSP is more specific. Scott Helme, the owner of [report-uri.io](https://report-uri.io/), has told me in the past that it's better to specify each individual resource.

~~~
'script-src assets.gratipay.com;'
'style-src assets.gratipay.com;'
~~~

For the time being the CSP allows all images: `img-src *;`. I'll look into a better solution. :)

whit537:
I like it. Let's add another `test_ahtr_` in `tests/py/test_security.py` and then I think we're good to go. 👍

EdOverflow:
Voilà! 😃

whit537:
Sweet!💃 

I am merging into a local branch for some final cleanup. You should really [make that onboarding ticket](http://inside.gratipay.com/big-picture/welcome) so we can get you properly set up here. :-)

PR #4248, opened 2016-12-22
Add lang="en" to the html start tag.

EdOverflow:


clone1018:
What if the user is browsing Gratipay using another language https://github.com/gratipay/gratipay.com/tree/master/i18n/core

EdOverflow:
Thanks for the informative response. Does anyone have any suggestions as to how the `lang` attribute could be implemented?

Changaco:
@EdOverflow https://github.com/liberapay/liberapay.com/blob/147/templates/base.html#L2

EdOverflow:
Ah, nice! Thank you @Changaco.

PR #4255, opened 2016-12-23
Add tests for payday behavior when balance is negative

whit537:
In limited situations, we send a participant's Gratipay escrow balance negative. For example, when they receive stolen money that we refund (https://github.com/gratipay/violations/issues/34), or when I forget to post back MassPay and subsequently overpay people (https://github.com/gratipay/inside.gratipay.com/issues/951). In these situations, we want to recoup the loss from future takes, but our current behavior is to charge their credit card (if they have one) to make up the difference. That penalizes the participant for a mistake on our part, so I think the best thing to do is update payday to not charge a card when a pre-payday balance is negative.

**Update:** We ended up working around the charges for the 10 people in question by manually overriding the `error` field on their route to prevent charging. The tests on this PR are still valuable as documentation of existing behavior.

### DIFF
--- a/gratipay/testing/billing.py
+++ b/gratipay/testing/billing.py
@@ -49,6 +49,14 @@ class BillingHarness(Harness, PaydayMixin):
                                            claimed_time='now', email_address='abcd@gmail.com')
         self.homer_route = ExchangeRoute.insert(self.homer, 'paypal', 'abcd@gmail.com')
 
+        # A customer with both Braintree and Paypal attached.
+        self.marge = self.make_participant('marge', is_suspicious=False, verified_in='US',
+                                           claimed_time='now', email_address='marge@example.com')
+        self.marge_paypal_route = ExchangeRoute.insert(self.marge, 'paypal', 'marge@example.com')
+        self.marge_braintree_route = ExchangeRoute.insert(self.marge, 'braintree-cc',
+                                                                               self.marge_cc_token)
+        self.marge_route = self.marge_paypal_route  # so hacky
+
 
     @classmethod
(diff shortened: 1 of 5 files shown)


whit537:
I think that has implications for people who would've received from these people, though.

A &rarr; B
B &rarr; C

We double-pay B and send their balance negative to correct. Next week they receive from A again, but that just brings them up to 0 and they don't have enough to pay C.

Sorry, C. :-(

whit537:
Or can we be smart enough to charge if the net for the payday is positive? Seems like we should be able to.

whit537:
This is complicated because `new_balance < 0` is [precisely how we decide](https://github.com/gratipay/gratipay.com/blob/2024/gratipay/billing/payday.py#L285) which card holds to settle.

whit537:
One possibility would be to floor [`old_balance` and/or `new_balance`](https://github.com/gratipay/gratipay.com/blob/2026/sql/payday.sql#L8-L9) at 0. Would that work cleanly?

whit537:
Alright, this is non-trivial.

whit537:
Spurious build failures? Restarted ...

whit537:
Rebased, was 68e36a2cc8c10b04d3438f3888b563aa4b959853.

whit537:
Ping @mattbk @nobodxbodon @kaguillera I think this should be an easy one to review, because it only touches tests.

whit537:
Withdrawing from review. Tests are failing and it's not obvious to me why.

whit537:
Giving up. 😞

PR #4252, opened 2016-12-22
Fix CSP again

whit537:
Follow-on from #4250 ... here's one more, @EdOverflow.

(Rolled back again in Heroku.)

screen shot 2016-12-22 at 12 27 55 pm

PR #4259, opened 2016-12-23
Add 401 response to about/me pages.

### LINKED ISSUE #4258: about/me/history does not match other pages that require login to view
`about/me/history` (not logged in):
image

`~whit537/history` (not logged in)
image

The latter is a 401 code, the former is not.

mattbk:
Closes #4258.


mattbk:
The Travis failure is not because of something in this PR.
> tests/py/test_history.py::TestHistory::test_iter_payday_events FAILED

Must be related to https://github.com/gratipay/gratipay.com/pull/4252 et alii.

whit537:
How about a test? :-)

mattbk:
!

Either I got it right on the first try, or it didn't even test the right thing. Hmm.

whit537:
> it didn't even test the right thing

It reads to me like you got it right on the first try, though the second test duplicates the first.

Good work! :-)

!m @mattbk

mattbk:
> though the second test duplicates the first.

You are correct. I think I missed a small difference when I copied them and changed the URLs.

whit537:
> I think I missed a small difference when I copied them and changed the URLs.

Well, from the test name it sounded like you intended to vary the user from anon to auth-no-admin or some such.

whit537:
That wouldn't be bad to have. Also, do you think it'd be worth checking subpaths, too? `/about/me/history/` is the one I think we were starting with here?

whit537:
Looks good to me. Blocking on #4252 though, to avoid `master` drifting further before getting back on track.

mattbk:
#4252 is merged, unblocking.

whit537:
Deployed!

screen shot 2016-12-31 at 8 32 47 am

whit537:
!m @mattbk

Issue #4235, opened 2016-12-16
Smarten up rendered descriptions in search


whit537:
When displaying search results, we include a participant's profile statement, which we render as markdown and then excerpt (theoretically we'll do the same for project's some day). Applying styles to the rendered markdown can have unintended side effects. For example, [searching for webpack](https://gratipay.com/search?q=webpack) matches the usernames `webchick` and `Weblate`, and the half-open HTML from the first's description clobbers the markup for the second.

------------------------------

![screen shot 2016-12-16 at 6 09 19 am](https://cloud.githubusercontent.com/assets/134455/21260902/35b6303c-c356-11e6-9519-e870fd931244.png)

---------------------------------
![screen shot 2016-12-16 at 6 16 42 am](https://cloud.githubusercontent.com/assets/134455/21261069/402aafe2-c357-11e6-9ca6-146e9026d7d3.png)

mattbk:
Helpful? https://coderwall.com/p/dnnlea/how-to-truncate-a-string-that-s-markdown-format

whit537:
Hmm ... maybe. a) I think that's Ruby? and b) Why would it work to truncate before escaping instead of after? We don't even want to escape, do we?

Probably the real way to do this is to store an html-stripped version of the text, and search over and display that.

mattbk:
I think your first point is the most valid.

whit537:
I think we should try using http://stackoverflow.com/a/925630 to compute and store a flattened plain-text version of the content, and search/display against that instead of the marked-up version.

whit537:
Or possibly https://github.com/aaronsw/html2text but that seems like overkill.

whit537:
Moving to a PR in #4266.

whit537:
[Fixed](https://gratipay.com/search?q=webpack)! 💃

![screen shot 2017-01-09 at 4 30 14 pm](https://cloud.githubusercontent.com/assets/134455/21784299/ec43e692-d688-11e6-8cad-22ceec67a0fe.png)

PR #4266, opened 2016-12-30
Clean up descriptions in search

whit537:


whit537:
D'oh! I've been writing this for projects, but it's participants where we are showing statements and being affected.

whit537:
I've got a pretty good start on this but am now hitting a bug. Hoping to have this ready for review soon ...

whit537:
Blech. Okay, let's sort this out ...

whit537:
Users store content in Markdown, which can also include HTML.

whit537:
We want to store a rendered-and-stripped version to search against and display in search results.

whit537:
Something is happening with entity references that I don't understand.

whit537:
If I enter `four & twenty`, I expect `four &amp; twenty` to be sent to the browser. Why do I expect that? Is that what's happening? Why or why not?

whit537:
Alright, so rendering `&amp;` preserves the HTML, but then `striptags` [renders](https://github.com/pallets/markupsafe/blob/0.23/markupsafe/__init__.py#L136-L142) it.

whit537:
Alright, so there are two paths to populating the description `span` on the search page:

- excerpt from a search result, and
- head of the statement.

whit537:
This is complicated by the fact that in the first case, we want to highlight the search terms in the description.

whit537:
We want to follow different paths based on username or statement match, due to the requirement to highlight search terms in the latter case.

Truncating always has to happen on `text/plain`, or we risk terminating markup early, which is what led to the original bug described in #4235.

#### Username Match

If we store HTML-escaped `text/plain`, then (2) needs to be smart enough not to result in, e.g., `&am` (cutting off `&amp;`).

1. **store `text/plain`**—no Markdown, no HTML, unescaped
1. **load excerpt**—truncate the statement from the beginning
1. **escape HTML**

#### Statement Match

We can't have Postgres use HTML to demarcate search terms in 2.i, because we need to escape HTML afterwards.

1. **store `text/plain`**—no Markdown, no HTML, unescaped
1. **load excerpt**—let Postgres select a portion that matches the search terms
    1. to support (4), ask Postgres to add highlighting markup
1. **escape HTML**
1. **highlight search terms**—convert the markers we asked Postgres to use into HTML

whit537:
Okay! I've cleaned up the commits here and rebased a new one with all of the `utils` changes I think we're gonna need. My new plan is to store escaped HTML, because that's what Postgres seems to want to work with for full-text search. The [parser](https://www.postgresql.org/docs/9.3/static/textsearch-controls.html#TEXTSEARCH-PARSING-DOCUMENTS) throws away HTML!

```sql
gratipay-test=# select to_tsvector('foo bar');
┌─────────────────┐
│   to_tsvector   │
├─────────────────┤
│ 'bar':2 'foo':1 │
└─────────────────┘
(1 row)

gratipay-test=# select to_tsvector('foo&amp;bar');
┌─────────────────┐
│   to_tsvector   │
├─────────────────┤
│ 'bar':2 'foo':1 │
└─────────────────┘
(1 row)

gratipay-test=# select to_tsvector('<h1>foo</h1> bar');
┌─────────────────┐
│   to_tsvector   │
├─────────────────┤
│ 'bar':2 'foo':1 │
└─────────────────┘
(1 row)

gratipay-test=# select to_tsvector('&lt;h1&gt;foo&lt;/h1&gt; bar');
┌────────────────────────────────┐
│          to_tsvector           │
├────────────────────────────────┤
│ '/h1':3 'bar':4 'foo':2 'h1':1 │
└────────────────────────────────┘
(1 row)

gratipay-test=#
```

whit537:
What that means is we can store _escaped_ statements, and have Postgres use its usual `<b>`, `</b>` to highlight search terms. We just need to take care with truncating statements in the username match case as noted, which is done in 7d845881c5e98b7ca613fbdc0ccd795246fa85af.

whit537:
Okay! Last thing here is to write a `branch.py` script to populate `content_scrubbed` for old statements.

whit537:
We have 10,271 records in the statements table.

whit537:
Hmm ... stymied by a UnicodeDecode error with a local backup that I'm not seeing in production ...

whit537:
I just rebuilt my local environment and I'm still seeing this error. Does the data or code change between prod and local?

whit537:
I've confirmed via `heroku pg:psql` *and* `heroku run python` that the data does not change.

whit537:
Python 2.7.11 in production, 2.7.12 locally.

whit537:
We have pinned deps so _most_ jitter should be dampened.

whit537:
Same `misaka.__version__`: 1.0.2.

whit537:
`sys.getdefaultencoding()` is `ascii` for both.

whit537:
I'm trying to whittle down a minimal test case.

whit537:
Alright this is goofy ah.

whit537:
This fails locally; all parts are necessary.

```python
from gratipay import wireup

import misaka as m
m.html(u'a@.\xa0', extensions=m.EXT_AUTOLINK)

```

```
[gratipay] $ python foo.py 
  File "foo.py", line 4, in <module>
    m.html(u'a@.\xa0', extensions=m.EXT_AUTOLINK)
  File "misaka.pyx", line 59, in misaka.html (src/misaka.c:813)
  File "misaka.pyx", line 232, in misaka.Markdown.render (src/misaka.c:1749)
  File "/Users/whit537/personal/gratipay/gratipay.com/env/lib/python2.7/encodings/utf_8.py", line 16, in decode
    return codecs.utf_8_decode(input, errors, True)
UnicodeDecodeError: 'utf8' codec can't decode byte 0xc2 in position 30: invalid continuation byte
[gratipay] $
```

Does it fail remotely?

whit537:
No!

```
$ heroku run bash
Running bash on ⬢ gratipay... up, run.2570 (Hobby)
~ $ python
Python 2.7.11 (default, Dec  7 2015, 21:16:24)
[GCC 4.8.4] on linux2
Type "help", "copyright", "credits" or "license" for more information.
>>> from gratipay import wireup
>>> import misaka as m
>>> m.html(u'a@.\xa0', extensions=m.EXT_AUTOLINK)
u'<p><a href="mailto:a@">a@</a>.\xa0</p>\n'
>>>
```

whit537:
Alright, so I have local side-effect from importing wireup?

whit537:
To the binary search mobile! 🚙

whit537:
Nothing in `gratipay.wireup` ... I suspect it's the `locale` [gymnastics](https://github.com/gratipay/gratipay.com/blob/2030/gratipay/__init__.py#L7-L17) in `gratipay/__init__.py` ...

whit537:
Clobbering the locale magic *does* make the behavior go away, but why? On prod the locale value seems to be same as locally ...

whit537:
Alright, the locale is set as expected at Heroku, and indeed it appears to not result in Misaka falling over.

whit537:
I think I need to not get to the bottom of this right now. I think I should write the `branch.py` script to work around local errors like this, and trust that it'll work in production.

whit537:
Okay, I only find two failures out of 10,000+ statements. 👍

mattbk:
> two failures out of 10,000+ statements

Vote them off the island?

whit537:
I guess it's not sensitive what the accounts are:

https://gratipay.com/~rummik/
https://gratipay.com/~marwasaf/

We'll just want to check those after migration (though the latter is spam anyway).

whit537:
Confirmed that both exhibit the same 500 behavior locally.

screen shot 2017-01-03 at 3 09 09 pm

whit537:
Okay! Ready for review! 💃 🌻

gratipay/utils/__init__.py
```diff
@@ -202,12 +201,15 @@ def format_money(money):
     return format % money
 
 
-def excerpt_intro(text, length=175, append=u'…'):
-    if not text:
-        return ''
-    if len(text) > length:
-        return text[:length] + append
-    return text
+def truncate(text, target=160, append=' …'):
```
nobodxbodon: If there's no usage with a different 'append', remove it from parameter? Otherwise may need some more test cases with this parameter?

whit537: The recursive case is such a usage.

whit537: Tests and cleanup in c245baade8f24284b6dd2817d135a52d14bf682c! :-)

whit537:
Rebased, was 14a042b3734ce17e3aca56f14ddd09b4fb225c22.

whit537:
Rebased, was 9e844a364561c530191c5a875cad05fc644331f4.

How do we get this reviewed, @gratipay? This is the last big PR on the "[Relax Open Work Requirement](https://github.com/orgs/gratipay/projects/5?fullscreen=true)" project!

gratipay/models/team/__init__.py
```diff
@@ -167,31 +167,31 @@ def get_payment_distribution(self):
...
+              UPDATE teams
+                 SET ({0}) = ({1})
+               WHERE id = %s
+              """.format(cols, placeholders), vals+(self.id,)
+            )
+            add_event(c, 'team', dict( action='update'
+                                     , id=self.id
+                                     , **old_value
+                                      ))
+            self.set_attributes(**kw)
```
nobodxbodon: To make sure, this only changes formatting and not code?

sql/branch.sql
```diff
@@ -0,0 +1,11 @@
...
+
+    ALTER TABLE statements ADD COLUMN content_scrubbed text NOT NULL DEFAULT '';
+
+    DROP TRIGGER search_vector_update ON statements;
+    CREATE TRIGGER search_vector_update
+        BEFORE INSERT OR UPDATE ON statements
+        FOR EACH ROW EXECUTE PROCEDURE
+        tsvector_update_trigger_column(search_vector, search_conf, content_scrubbed);
+
+END;
```
nobodxbodon: Could you add some comments about what's this for?

tests/py/test_markdown.py
```diff
@@ -6,50 +7,99 @@
 
 class TestMarkdown(Harness):
 
-    # render
+    # r - render
```
nobodxbodon: In past project in Java, we tried to avoid abbreviations in naming to increase readability.

Maybe rename the file to test_render and remove all 'render' string in the methods?

tests/py/test_search.py
```diff
@@ -0,0 +1,77 @@
...
+devotion&mdash;that we here highly resolve that these dead shall not have died
+in vain&mdash;that this nation, under God, shall have a new birth of
+freedom&mdash;and that government of the people, by the people, for the people,
+shall not perish from the earth.</p>
+
+"""
+
+
+class Tests(Harness):
+
```
nobodxbodon: nit: rename class name to a less general one?

www/search.spt
```diff
@@ -1,16 +1,20 @@
...
+    if not excerpt:
+        return Markup('')
+    return Markup('&middot; ' + excerpt)
 
 def get_processed_excerpt(participant):
-    statement = participant.get_statement(request.accept_langs)[0] or ''
+    """For a username match, take the first part of the statement.
+    """
+    statement = participant.get_statement(request.accept_langs, scrubbed=True)[0] or ''
+    statement = truncate(statement, 128, '')
```
nobodxbodon: why 128 here but 256 in original code?

gratipay/models/team/__init__.py
whit537: Yes. When I started this PR I thought I was going to be changing this file, so I made f0b74e10c0c25febfd24a6e7a7bd16f025bd52f2 in order to standardize the whitespace first. I didn't end up changing the code but I left the formatting change in place.

tests/py/test_markdown.py
whit537: In general I agree, but I think abbreviating is okay in this case. By now it's a fairly standard pattern in our test suite to abbreviate the name of the function we're testing in the block where we're testing it. In one sense this improves readability by allowing you to be more descriptive about the purpose of the test in the rest of the test name without going past our 100-char line limit. It also makes it easy to run the subset of the test suite dealing with those tests using `py.test -k test_r_`.

tests/py/test_search.py
whit537: Why? Again, this is a standard pattern in our test suite, and is basically just boilerplate. I would rather use py.test's dependency injected [fixtures](http://docs.pytest.org/en/latest/fixture.html) (we've had a good experience with them [in Aspen/Pando](https://github.com/AspenWeb/pando.py/tree/0.42/tests)), but we're too far along the class-based test path to change now. Better to minimize boilerplate if we can't avoid it entirely. :)

www/search.spt
whit537: 128 is plenty to fill the UI slot. No reason to send more than we're ever going to display.

sql/branch.sql
whit537: 1c761aae327b8e17b03b06221298e78c3c0141b0, ya?

tests/py/test_markdown.py
whit537: ... aaaaaaaaand addressed in 7e6778c16d42e720c33b330539afe645b5f8e494. :-)

Easy enough to do `py.test test_markdown.py::TestRender`.

tests/py/test_search.py
whit537: Fixed in 7e6778c16d42e720c33b330539afe645b5f8e494. 😊

whit537:
Comments addressed. Back to you, @nobodxbodon! :-)

nobodxbodon:
Frankly it seems proper to me to let committer submit, especially if the committer is project member and at least another member has approved it. Because 1) committer usually has more motivation in pushing 2) committer knows better when to push (tests passed, everything tested locally, etc)

BTW jut to be on the same page, reviewer will review again if committer adds more commits after reviewer approved. I'll check  7e6778c in our case, as it's added after my approving.

whit537:
I think it's better for someone other than the primary/last commit author to merge, because it spreads responsibility for the PR across two (or more) people. It makes it a _collective_ change rather than an individual one.

whit537:
> 1) committer usually has more motivation in pushing 

To me that's a reason _not_ to let the primary/last committer merge. We want to spread power and responsibility, not concentrate it.

> 2) committer knows better when to push (tests passed, everything tested locally, etc)

You can also see when the tests are passing, both at Travis and locally. Not sure what the etc. is. ;-)

> BTW jut to be on the same page, reviewer will review again if committer adds more commits after reviewer approved.

Yes, the person merging needs to approve of all changes they're merging.

whit537:
@nobodxbodon If you're not comfortable merging, you may live this for someone else to review and merge. :-)

whit537:
[Consent](http://inside.gratipay.com/big-picture/brand) all around! 💃

nobodxbodon:
Maybe better to be discussed somewhere else, but IMO the reviewing process itself already makes the PR a collective change and responsibility distributed, even if reviewer just approve without having any comments. 

I want this PR in and I've no problem merging it, but I wouldn't feel comfortable without checking with you if you feels good with the PR yourself. It's not because I don't want to share responsiblity (as my reviewing and approval already did, as mentioned above), but because from my experience committer in most cases have more knowledge about the implementation details of the PR, and they may want to do some tests or re-check before actual pushing, especially if there are some additiontal changes during the review. Besides, waiting for tests passing is not a good use of reviewer's time, not to mention the test could fail.

Last but not least, I'm all with consent, but there's surely another side of coin. As you know better about the reviewing speed, IMO one reviewer's approval is enough at current state, unless committer strongly believes the code needs to be reviewed by particular member. In general I prefer [fail early fail fast fail often](http://tracks.roojoom.com/u/shirly-ronen-harel,60/agilelean-fail-early-fail-fast-fail-often,493).

whit537:
> I want this PR in and I've no problem merging it, but I wouldn't feel comfortable without checking with you if you feels good with the PR yourself.

Perfect attitude! 👍  I feel good with this PR, and so we have converged! [SYN, SYN-ACK, ACK](http://www.inetdaemon.com/tutorials/internet/tcp/3-way_handshake.shtml). The "normal" Gratipay behavior at this point is for you to merge. 🙇  :-)

whit537:
(There are occasional cases where we stretch the process to keep up velocity, but I don't see us to be in one such situation right now. :)

whit537:
💃

!m @nobodxbodon :D

nobodxbodon:
> (There are occasional cases where we stretch the process to keep up velocity, but I don't see us to be in one such situation right now. :)

Before this I was used to getting feedback in one day or less after sending out PR in review process. I don't think I can accomplish that here myself, but will try.

whit537:
The two from https://github.com/gratipay/gratipay.com/pull/4266#issuecomment-270211124 still look fine.

whit537:
Branch file cleared in https://github.com/gratipay/gratipay.com/commit/da2bd9c0cab3b45dee8217bc048b2ce8f5bf3e9e.

PR #4268, opened 2016-12-31
Tighten up the banner on the homepage

whit537:
Follow-on from #4267.

- Drops the old banner JPG and renames the new PNG
- Tweaks the homepage copy and positioning
- Brings back "About" in the top nav

----------------

screen shot 2016-12-30 at 9 06 43 pm

Issue #1167, opened 2013-07-17
revamp checkout flow


whit537:
The idea with this ticket is to produce a modal checkout flow that can be used on both gittip.com profiles as well as embedded w/ a "Donate" button on other websites.

Some assets for this are over here: https://layervault.com/jeromegn/Gittip/Widget/tipping.sketch

@jeromegn built the one below for Clarity.fm (visible live on [www.danmartell.com](http://www.danmartell.com/), "Schedule a Call" on the sidebar). See #737 for another example and more of the motivation behind this ticket. This should be hosted on gttp.co and work will go in that repo, but please keep tickets in here.

![screen shot 2013-07-17 at 4 54 04 pm](https://f.cloud.github.com/assets/134455/815290/be8cae62-ef25-11e2-85ac-a915a1ab3164.png)

whit537:
+1 from @elliottw at #3676:

>Have a widget that groups can put on there site, that allows non-gratipay users to easily sign up and donate to that particular group all within their site.
>
<img width="1035" alt="screen shot 2015-08-11 at 12 23 20 am" src="https://cloud.githubusercontent.com/assets/1633097/9190090/489c472a-3fbf-11e5-932a-b64ef22987e6.png">
>
I think Gratipay should be as prominent as it can without being annoying, and we can even have a button like "now go explore other projects to support!".
>
The current thing holding me back from moving a lot of catapultpgh members to gratipay, is that a lot of our base are not developers and I have had to spend 15 minutes onboarding people on to Gratipay to donate, which means lower donation rates not because people don't want to donate, but because they feel they it's very complex and confusing and they are signing up for some big thing when all they want to do is give catapult 10 bucks a months.

And @kzisme:

> I was thinking about something like this for a stint of time. I see a lot of other companies using widgets/buttons on their Twitch streams. Many other people would be interested if these buttons were introduced and the signup process was streamlined.

And @mattbk:

>I like the idea of streamlining the registration/giving process while offsite.
>
Would it have to wait for #1052?

whit537:
This is an important project that needs a champion.

elliottw:
just trying to wrap my ahead around the progress on this issue. @domenkozar looks like you'd a bunch of work. What else needs to get finished?
Do we have a functioning coded prototype?
Has everyone agreed on a user workflow?

My coworking space, @cataputlpgh would move everyone to gratipay if we can get this figured out, so I'm highly motivated!

domenkozar:
[...]

PR #4269, opened 2017-01-01
Clarify that social logins are also social account creators

mattbk:
As noted at https://github.com/gratipay/gratipay.com/pull/4268#issuecomment-269844784 and https://github.com/gratipay/gratipay.com/issues/1167#issuecomment-269845813, I think it's confusing that there's no "register" or "create account" instructions, only "sign in."

I'm torn between "register," which sounds technical yet accurate and "create account," which makes sense unless you think about the social media _accounts_ that can be used to sign in. "Wait, am I creating a Twitter account?" "Create account" is also longer and breaks the UI in the sidebar especially.


mattbk:
Food for thought on why I don't want to use "Sign in or Sign up": [it's visually confusing](http://uxmovement.com/buttons/why-sign-up-and-sign-in-button-labels-confuse-users/).

mattbk:
screen shot 2016-12-31 at 10 29 38 pm

mattbk:
Alternative wording if we're not tied to "Sign in":

screen shot 2016-12-31 at 10 33 49 pm

mattbk:
- `min-width: 165px` if "Log in or Sign up" is used (insert at https://github.com/gratipay/gratipay.com/blob/4b5b0b3e287424a3e5b4179d49eddac44064f7e4/scss/components/sign_in.scss#L101).
- `min-width: 176px` if "Sign in or Register is used.
- Sidebar looks okay regardless, probably because the dropdown menu isn't connected to the button, and has rounded corners to help separate it.

Although with translations, these pixel values won't match up anyway.

nobodxbodon:
How about using "Login with", like in below:

screen shot 2017-01-01 at 11 38 15 pm

mattbk:
Foursquare still has a different "Sign up for Foursquare" link.

That Zoho screenshot leaves me wondering how I create an account in the first place.  It's the same issue I see with Gratipay.

mattbk:
Knew I should have waited...

mattbk:
Where is it getting "Log in" from? Before it was "Sign in." Cache?

```
=================================== FAILURES ===================================
_____________ Tests.test_homepage_renders_copy_correctly_for_anon ______________
self = <test_homepage.Tests testMethod=test_homepage_renders_copy_correctly_for_anon>
    def test_homepage_renders_copy_correctly_for_anon(self):
        assert self.css('#content h1').html == 'Projects'
>       assert self.css('#header .sign-in button').html.strip()[:7] == 'Log in or Sign up'
E       AssertionError: assert 'Log in ' == 'Log in or Sign up'
E         - Log in 
E         + Log in or Sign up```

mattbk:
> Where is it getting "Log in" from?

Compares only part of the string because there is HTML after. So set the string length equal to the length of your test string.

whit537:
I like using a consistent nomenclature for signing up, in, and out, so I'd rather not introduce "log *" or "register."

How about Medium's "Sign in / Sign up"?

screen shot 2017-01-19 at 10 51 22 am
screen shot 2017-01-19 at 10 52 05 am

Or even "Sign in/up"? Is that too terse?

screen shot 2017-01-19 at 10 54 36 am
screen shot 2017-01-19 at 10 54 52 am

Or if we want to be geeky, "Sign {in,up}". :)

screen shot 2017-01-19 at 10 55 07 am

mattbk:
I copied Medium. It's less of a big deal than I originally thought, because there's only one button, not two.

PR #4277, opened 2017-01-06
Improve git clone process, save time, bandwidth and disk space

PeterDaveHello:
Repo size reduce from 2084 KB to 1156 KB, we can save time, bandwidth and disk space, and a `checkout` command/process from this change.


whit537:
Thanks! Will merge when green ...

PR #4279, opened 2017-01-06
Dockerfile fixes

PeterDaveHello:
The old one was broken in not only one part, so just do a quick refactor to 

1. Make it work (I mean - at least the docker image can be built)
2. Optimize image build time spent - `5m43.605s` -> `4m15.379s` on my computer
3. Optimize the image size, from `839.8 MB` to `776 MB`


whit537:
You are now our resident Docker expert! :-)

No-one else is really using this right now that I know of ...

PR #4280, opened 2017-01-06
Shellsciprt syntax check / lint in Travis CI

PeterDaveHello:


.travis.yml
```diff
@@ -4,6 +4,7 @@ addons:
 before_install:
  - git branch -vv | grep '^*'
  - pwd
+ - find . -name "*.sh" -exec bash -n {} \;
```
whit537: What does `bash -n` do?

PeterDaveHello: Kind of syntax check/lint method with shellscript:
https://stackoverflow.com/questions/171924/how-do-i-check-syntax-in-bash-without-running-the-script

PR #4281, opened 2017-01-06
Travis ci tuning

PeterDaveHello:


.travis.yml
```diff
@@ -14,17 +16,18 @@ cache:
   directories:
     - env/bin
     - env/lib/python2.7/site-packages
+    - node_modules
```
whit537: There might've been a reason we took that out, I forget. I guess we'll find out again if we hit it again. :)

.travis.yml
```diff
@@ -14,17 +16,18 @@ cache:
...
- - make env
+ - make env -j2
  - env/bin/pip install --upgrade ijson==2.3.0
 before_script:
  - echo "DATABASE_URL=dbname=gratipay" | tee -a tests/local.env local.env
  - echo "REQUIRE_YAJL=true" | tee -a tests/local.env local.env
  - psql -U postgres -c 'CREATE DATABASE "gratipay";'
  - if [ "${TRAVIS_BRANCH}" = "master" -a "${TRAVIS_PULL_REQUEST}" = "false" ]; then rm -rfv tests/py/fixtures; fi
-script: LD_LIBRARY_PATH=/usr/local/lib make bgrun test doc
+script: LD_LIBRARY_PATH=/usr/local/lib make bgrun test doc -j2
```
whit537: Since our Makefile is mostly phony targets I doubt this will make a difference, but from what I can tell it doesn't hurt. Did you actually benchmark?

PeterDaveHello: The deviation on sharing CI system like travis.org is always too high, so I didn't benchmark it, if the makefile is good enough, it won't hurt at all, if we got some problem, it may mean that  we should improve the makefile dependency control, IMO.

PR #4232, opened 2016-12-13
Send email for Terms notification

whit537:
Includes a script that we can reuse next time we need to send out a transactional email to everyone.

PR #4282, opened 2017-01-06
Finish sending out Terms notification

whit537:
Picking up from https://github.com/gratipay/gratipay.com/pull/4232#issuecomment-271026845 ...

- [x] Update the participants [query](https://github.com/gratipay/gratipay.com/blob/2034/gratipay/utils/emails/queue_branch_email.py#L50-L57) to only select those who have signed in (`session_expires`) or show up in `payments` or `exchanges` within the past [180 days](https://github.com/gratipay/gratipay.com/pull/4232#issuecomment-270976120).
- [ ] Mail all of them, tracking bounces in SQS.
- [ ] Manually retrieve bad addresses from the queue—**within 14 days!**—and drop them to unverified in the database.
- [x] Reticket proper bounce handling with a webhook back to https://gratipay.com/: #4284.

### DIFF
--- a/gratipay/utils/emails/queue_branch_email.py
+++ b/gratipay/utils/emails/queue_branch_email.py
@@ -7,6 +7,23 @@ import random
 from gratipay import wireup
 
 
+def get_recently_active_participants(db):
+    return db.all("""
+        SELECT p.*::participants
+          FROM participants p
+         WHERE email_address is not null
+           AND claimed_time  is not null
+           AND is_closed     is not true
+           AND is_suspicious is not true
+           AND ( SELECT count(*)
+                   FROM exchanges e
+                  WHERE e.participant = p.username
+                    AND timestamp > now() - '180 days'::interval
+                ) > 0
+      ORDER BY p.username ASC
+    """)
+
+
 def main(_argv=sys.argv, _input=raw_input, _print=print):
     """This is a script to enqueue global site notification emails.
 
@@ -47,14 +64,7 @@ def main(_argv=sys.argv, _input=raw_input, _print=print):
         prompt("Are you REALLY ready?")
         prompt("... ?")
         _print("Okay, you asked for it!")
-        participants = db.all("""
-            SELECT p.*::participants
-              FROM participants p
-             WHERE email_address is not null
-               AND claimed_time  is not null
-               AND is_closed     is not true
-               AND is_suspicious is not true
-        """)
+        participants = get_recently_active_participants(db)
     else:
         _print("Okay, just {}.".format(username))
         participants = db.all("SELECT p.*::participants FROM participants p "
--- a/tests/py/test_email.py
+++ b/tests/py/test_email.py
@@ -284,6 +284,43 @@ def queue_branch_email(username, _argv=None, _input=None, _print=None):
     return retcode, stdout, stderr
 
 
+class TestGetRecentlyActiveParticipants(EmailHarness):
+
+    def check(self):
+        return _queue_branch_email.get_recently_active_participants(self.db)
+
+    def test_gets_recently_active_participants(self):
+        alice = self.make_participant('alice', claimed_time='now', email_address='a@example.com')
+        self.make_exchange('braintree-cc', 50, 0, alice)
+        assert self.check() == [alice]
+
+    def test_ignores_participants_with_no_exchanges(self):
+        self.make_participant('alice', claimed_time='now', email_address='a@example.com')
+        assert self.check() == []
+
+    def test_ignores_participants_with_no_recent_exchanges(self):
+        alice = self.make_participant('alice', claimed_time='now', email_address='a@example.com')
+        self.make_exchange('braintree-cc', 50, 0, alice)
+        self.db.run("UPDATE exchanges SET timestamp = timestamp - '181 days'::interval")
+        assert self.check() == []
+
+    def test_keeps_participants_straight(self):
+        alice = self.make_participant('alice', claimed_time='now', email_address='a@example.com')
+        self.make_exchange('braintree-cc', 50, 0, alice)
+
+        bob = self.make_participant('bob', claimed_time='now', email_address='b@example.com')
+        self.make_exchange('braintree-cc', 50, 0, bob)
+
+        carl = self.make_participant('carl', claimed_time='now', email_address='c@example.com')
+        self.make_exchange('braintree-cc', 50, 0, carl)
+        self.db.run("UPDATE exchanges SET timestamp = timestamp - '181 days'::interval "
+                    "WHERE participant='carl'")
+
+        self.make_participant('dana', claimed_time='now', email_address='d@example.com')
+
+        assert self.check() == [alice, bob]
+
+
 class TestQueueBranchEmail(EmailHarness):
 
     def nsent(self):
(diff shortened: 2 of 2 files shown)


whit537:
Reticketed https://github.com/gratipay/gratipay.com/issues/4284 for automated bounce processing.

whit537:
I'm thinking through how to constrain the list of participants we're mailing. What we want to avoid is blasting everyone because we want an excuse to blast everyone (turning this ToS notice into a marketing message, essentially). This [fellow thinks](https://medium.com/@jlkoepke/we-can-change-these-terms-at-anytime-the-detritus-of-terms-of-service-agreements-712409e2d0f1) that it is enough to have a notice on the website for a certain period of time (though he also says we should require ourselves to notify users in this way; cf. https://github.com/gratipay/inside.gratipay.com/issues/926). I suppose that means we can constrain the email blast to a list of more truly active users.

mattbk:
>If we determine, in our sole discretion, that a change to these Terms is material, we will notify you of the change through the Service or by emailing you at the address associated with your account. 

I never noticed that "material" clause there before. I _guess_ that means you can only notify people who are active, because otherwise it doesn't apply to them?

whit537:
@mattbk in [slack](http://gratipay.slackarchive.io/gratipay/-/1483908310.002222/1483984259.002326/1483984259002326/):

> I think Gratipay as a whole errs a little too much on the side of "never email users." What's the harm? If a "inactive" user receives an email, they'll either a) trash it, b) get mad they received it, and continue being inactive, or c) visit the site and see what's up. Don't the benefits of (c) outweigh the potential downsides of (b)?
>
> From my reading, it seems like "selling" (even if it's just a TOS notification) to people who are already users is much easier than selling to people who have never been users before. These are people who have already signed up, maybe entered credit card information, or are otherwise invested.

whit537:
When I get an email from a company I haven't interacted with in two years, I consider it an annoying interruption. It makes me think less of the company. "Sheesh. Don't you know I don't care about you?"

> Don't the benefits of (c) outweigh the potential downsides of (b)?

I guess for me I feel like they don't, is what I'm saying. 😞

Let's see how many participants we're talking about ...

As written, we're looking at ~5500 ...

whit537:
I find 4645 distinct participants in the exchanges table from all time.

whit537:
That's `select * from (select distinct on (participant) participant, count(*) from exchanges group by participant order by participant) _ order by count desc;`.

whit537:
I find 529 active in the past 180 days.

```
gratipay::MAROON=> select count(*) from (select distinct on (participant) participant, count(*) from exchanges where timestamp > now() - '180 days'::interval group by participant order by participant) _ order by count desc;
┌───────┐
│ count │
├───────┤
│   529 │
└───────┘
(1 row)
```

whit537:
So ... 10%.

whit537:
I'm really comfortable emailing that set of users with the terms notice. I think mailing the full 5500 is more risk than it's worth. We don't want to be spammers. 😞

whit537:
@mattbk Can you live with mailing our 529 recently active users?

whit537:
Actually, though, how many of those recent exchangers do we have email addresses for?

whit537:
Answer: 406.

```sql
SELECT * FROM (
    SELECT p.username
         , p.email_address
         , ( SELECT count(*)
               FROM exchanges
              WHERE participant=p.username
                AND timestamp > now() - '180 days'::interval
            ) AS nexchanges
      FROM participants p
     WHERE email_address is not null
       AND claimed_time  is not null
       AND is_closed     is not true
       AND is_suspicious is not true
) _ WHERE nexchanges > 0 ORDER BY nexchanges desc;

```

whit537:
I feel really good about mailing this list.

mattbk:
Yes.

whit537:
Okay! I've added a commit to refactor a nearby test suite per https://github.com/gratipay/gratipay.com/pull/4266#discussion_r95221517, and another to constrain the list of participants we're mailing to (with tests). Code is ready for review, @mattbk @nobodxbodon!

nobodxbodon:
I'll hold on reviewing till tests pass.

whit537:
Rebased and test suite (hopefully) cleaned up.

whit537:
Green! :-)

gratipay/utils/emails/queue_branch_email.py
```diff
@@ -7,6 +7,23 @@
 from gratipay import wireup
 
 
+def get_recently_active_participants(db):
+    return db.all("""
+        SELECT p.*::participants
```
nobodxbodon: not familiar with this :: syntax. could you please share a reference about its usage?

whit537:
@mattbk Eeeeeeee! Merge? :)

gratipay/utils/emails/queue_branch_email.py
```diff
@@ -7,6 +7,23 @@
...
+          FROM participants p
+         WHERE email_address is not null
+           AND claimed_time  is not null
+           AND is_closed     is not true
+           AND is_suspicious is not true
+           AND ( SELECT count(*)
+                   FROM exchanges e
+                  WHERE e.participant = p.username
+                    AND timestamp > now() - '180 days'::interval
```
nobodxbodon: It's same as 
```
NOW() - INTERVAL '180 days'
```
? Do we have preference here? Excuse me about rusty postgresql.

tests/py/test_email.py
```diff
@@ -284,27 +284,62 @@ def _print(string, file=None):
     return retcode, stdout, stderr
 
 
+class TestGetRecentlyActiveParticipants(EmailHarness):
+
+    def check(self):
+        return _queue_branch_email.get_recently_active_participants(self.db)
+
+    def test_gets_recently_active_participants(self):
+        alice = self.make_participant('alice', claimed_time='now', email_address='a@example.com')
```
nobodxbodon: create some helper method as there are many occurences here?

```
self.make_participant(__some_name___, claimed_time='now', email_address='a@example.com')
```

gratipay/utils/emails/queue_branch_email.py
whit537: The double-colon is a [type cast](https://www.postgresql.org/docs/9.3/static/sql-expressions.html#SQL-SYNTAX-TYPE-CASTS). For the particular usage here, see the [postgres.py ORM docs](http://postgres-py.readthedocs.io/en/latest/#introducing-table-types).

gratipay/utils/emails/queue_branch_email.py
whit537: Yes, they're equivalent.

tests/py/test_email.py
whit537: How about 95d866144f11deb3e18be40e989769277d277a86?

whit537: ... and ba6043e47ebd13c546d26e566d230310e04ed1cb.

whit537:
Back to you, @nobodxbodon ...

whit537:
Check 'em, friends. :-)

- [ ] @whit537
- [ ] @nobodxbodon 
- [ ] @mattbk

whit537:
Meh, overlaps with GitHub's (over-engineered) UI for approval.

I am good to go, @nobodxbodon! :-)

whit537:
💃

whit537:
Picking up in https://github.com/gratipay/inside.gratipay.com/issues/972 ...

PR #4290, opened 2017-01-12
Tinker with timing for Travis

whit537:
Trying to avoid spurious test failures.

https://travis-ci.org/gratipay/gratipay.com/builds/191469577

screen shot 2017-01-12 at 5 47 16 pm


whit537:
Eh, @mattbk @nobodxbodon @kaguillera?

mattbk:
Is it running one test before the previous test has finished?

whit537:
> Is it running one test before the previous test has finished?

No, it's a webdriver test, so I guess we're not giving the browser enough time to load the list of participants to add before trying to click on the first one in the list.

mattbk:
Let's try it and see.

PR #4291, opened 2017-01-13
Improve logic in account closing

whit537:
Impetus was bugs discovered at https://github.com/gratipay/gratipay.com/issues/3602#issuecomment-272541653.

- fix bugs in balance checking (`> 0` vs `!= 0`)
- allow admins to close accounts w/ balance
- refactor test suite a little bit
- add tests

#### Resolve

- [ ] https://sentry.io/gratipay/gratipay-com/issues/204504679/

PR #4292, opened 2017-01-13
Show as closed

whit537:
Following on from #4291 (which is included here), this adds a danger zone to the top of the project page, which is still accessible to owners and admins.

screen shot 2017-01-13 at 6 53 16 pm

### DIFF
--- a/tests/py/test_close.py
+++ b/tests/py/test_close.py
@@ -11,9 +11,7 @@ from gratipay.models.participant import Participant
 from gratipay.testing import Harness, D,P
 
 
-class TestClosing(Harness):
-
-    # close
+class TestClose(Harness):
 
     def test_close_closes(self):
         alice = self.make_participant('alice', claimed_time='now')
@@ -24,6 +22,16 @@ class TestClosing(Harness):
         alice = self.make_participant('alice', claimed_time='now', balance=D('10.00'))
         with pytest.raises(alice.BalanceIsNotZero):
             alice.close()
+
+    def test_close_can_be_overriden_for_balance_though(self):
+        alice = self.make_participant('alice', claimed_time='now', balance=D('10.00'))
+        alice.close(require_zero_balance=False)
+        assert P('alice').is_closed
+
+    def test_close_can_be_overriden_for_negative_balance_too(self):
+        alice = self.make_participant('alice', claimed_time='now', balance=D('10.00'))
+        alice.close(require_zero_balance=False)
+        assert P('alice').is_closed
 
     def test_close_fails_if_still_owns_a_team(self):
         alice = self.make_participant('alice', claimed_time='now')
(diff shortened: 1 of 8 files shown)


whit537:
Another one, @mattbk et al.

tests/py/test_close.py
```diff
@@ -25,6 +23,16 @@ def test_close_fails_if_still_a_balance(self):
         with pytest.raises(alice.BalanceIsNotZero):
             alice.close()
 
+    def test_close_can_be_overriden_for_balance_though(self):
+        alice = self.make_participant('alice', claimed_time='now', balance=D('10.00'))
+        alice.close(require_zero_balance=False)
+        assert P('alice').is_closed
+
+    def test_close_can_be_overriden_for_negative_balance_too(self):
+        alice = self.make_participant('alice', claimed_time='now', balance=D('10.00'))
```
nobodxbodon: -10.00?

tests/py/test_close.py
```diff
@@ -56,22 +66,85 @@ def test_can_post_to_close_page(self):
...
+        assert 'Personal Information' in body
+        assert 'Careful!' in body
+
+    def test_posting_with_balance_fails_for_owner(self):
+        self.make_participant('alice', claimed_time='now', balance=10)
+        response = self.client.PxST('/~alice/settings/close', auth_as='alice')
+        assert response.code == 400
+        assert not P('alice').is_closed
+
+    def test_posting_with_balance_succeeds_for_admin(self):
```
nobodxbodon: append "_with_negative_balance" in method name?

tests/py/test_close.py
whit537: Good catch! Fixed in 0d2deb66d7f6f53de07355d235115f2db68a3c62.

tests/py/test_close.py
whit537: Meh. We have tests above for the full matrix of {positive,negative} * {owner,admin}. The point here is to test with _any_ balance. The fact that this one is negative while the previous test uses a positive balance is just extra.

whit537: Because here we're running through the simplate, where the positive/negative isn't so much an issue as the permissions for various roles.

whit537:
I'm good to go here if you're good with 0d2deb66d7f6f53de07355d235115f2db68a3c62, @nobodxbodon. 👍

tests/py/test_close.py
nobodxbodon: There's a principle in testing that making test case mimimal/simplest to achieve the test goal. Any additional conditions would add noise and unnecessery maintanance burden (consider other contributors having the same confusion as me). 

Of course that's in theory :)

nobodxbodon:
!m @whit537 I'm good with 0d2deb6

mattbk:
- This is replacing #4291?
- Where does the balance go if you close?

mattbk:
> Where does the balance go if you close?

Oh, I see. It just hangs out in the database, but the account is closed. I guess we're assuming that this isn't likely to be a problem, because people will either want their money before closing or it will be small amounts. It should be easy enough to look up how much "abandoned" money there is in closed accounts in the future.

I'm satisfied with this. Pushing the button.

whit537:
@mattbk Participants can't close their account with a balance outstanding, but #4291 allows admins to do so, which is necessary for https://github.com/gratipay/gratipay.com/issues/3602#issuecomment-272322002.

PR #4298, opened 2017-01-17
Create a range of nreceiving_from values

mattbk:
For things like #4234, would be more realistic to have a range for `nreceiving_from` rather than a constant value.


mattbk:
I'm wondering if this needs more than I think it does...

mattbk:
Yes! Of course there is. Dropping this because I don't have time to push it through.

Issue #4297, opened 2017-01-16
Spec out claim/unclaim/reclaim packages


whit537:
-------------------------------

⚠️  **Flight deck for the [Claim/unclaim/reclaim packages](https://github.com/orgs/gratipay/projects/14?fullscreen=true) project moved to https://github.com/gratipay/gratipay.com/pull/4305.** ⚠️ 

-------------------------------

## Todo

- [ ] set up npm syncing—besides just needing to stay in sync, people are going to want to update the email address on file in their package.json and then "immediately" see the new one show up to continue the flow
- [ ] make room on the `/on/npm/foo/` page for the claiming workflow—move the name to the banner h1, and replace the wombat with the description: #4304
- [ ] link packages to teams: #4155
- [ ] opt in or out from `/on/npm/foo/` pages via email verification
- [ ] add payment widget on `/on/npm/foo/` pages

-------------------------------

⚠️  **Flight deck for the [Claim/unclaim/reclaim packages](https://github.com/orgs/gratipay/projects/14?fullscreen=true) project moved to https://github.com/gratipay/gratipay.com/pull/4305.** ⚠️ 

-------------------------------

whit537:
Here are some [mockups](https://gomockingbird.com/projects/j8evmau/4gXVnC) for claiming packages:

![screen shot 2017-01-19 at 1 14 43 pm](https://cloud.githubusercontent.com/assets/134455/22119370/5fdd84ec-de49-11e6-88b4-251d7ee4057a.png)
![screen shot 2017-01-19 at 1 14 59 pm](https://cloud.githubusercontent.com/assets/134455/22119371/5fe28a3c-de49-11e6-8c3b-9957c1a95835.png)
![screen shot 2017-01-19 at 1 15 29 pm](https://cloud.githubusercontent.com/assets/134455/22119372/5fe68bf0-de49-11e6-836b-96715821defe.png)

whit537:
The list of emails in the second mockup would be the ones we pull from the package.json. Perhaps we can indicate somehow (is bold enough?) that one an email is already added/verified/primary for the participant.

Are we cool with autoapproving all npm packages as projects on Gratipay?

whit537:
Once a package is claimed, it behaves _almost_ like a regular project. Claimed package profiles look like a project page. Participants can give to them, and owners can edit them. They can be unclaimed just like projects can be closed. Claimed projects show up on the homepage under the various tabs, and in search results with other projects.

Here are the differences:

[...]

PR #4305, opened 2017-01-24
Integrate initial package claiming PRs

whit537:
This is an integration branch/PR for the initial implementation of package claiming. Part of #4427.

### Specs

See https://github.com/gratipay/gratipay.com/issues/4297#issuecomment-273900610.

### Mocks

From https://github.com/gratipay/gratipay.com/issues/4297#issuecomment-273854209 (click for full-size):

[](https://cloud.githubusercontent.com/assets/134455/22119370/5fdd84ec-de49-11e6-88b4-251d7ee4057a.png) [](https://cloud.githubusercontent.com/assets/134455/22119371/5fe28a3c-de49-11e6-8c3b-9957c1a95835.png) [](https://cloud.githubusercontent.com/assets/134455/22119372/5fe68bf0-de49-11e6-836b-96715821defe.png)

### PRs (in merge order)

Numbers are [workflow steps](https://github.com/gratipay/gratipay.com/issues/4297#issuecomment-273900610).

- [x] (1), (2) add sign in w/ copy to `/on/npm/foo/` for anon: #4306 #4307 #4308 #4323
- [x] stub out `Package` to hold Python API: #4155
- [x] (3)—(6) add confirmation form to `/on/npm/foo/` for auth when unclaimed: #4309
- [x] (7), (8) implement `Package.send_confirmation_email`: #4335
- [x] (9) add `/on/npm/react/claim` endpoint to receive email link: #4397
- [x] (10)—(11) implement `/on/npm/foo/` as project page when claimed: #4398
- [x] (12)—(16) implement `/on/npm/foo/edit` as project edit page when claimed: #4404

#### &c.

- [x] minor refactor: #4422
- [x] some UX cleanups: #4406
- [x] style the claim form: #4410 
- [x] more UX cleanups: #4411
- [x] **verify that we can give to a package**—the 💟  of the matter! ... #4423
- [x] on `verify.html`, show packages newly linked if germane: #4424
- [x] work PayPal into the workflow somehow: #4428
- [x] fix an egregious bug at low widths: #4435

### Todo

- [x] think through merging/takeover—anything special here? like, start verification workflow, then merge accounts, then finish verification.—looks like we already do [something](https://github.com/gratipay/gratipay.com/blob/2069/tests/py/test_take_over.py#L185-L206) &hellip; 9b7d1471693ed0876e7cf492fa7e71fa8eae421c
- [x] what about linking an existing project to an existing npm package? &hellip; sounds like an admin thing; bump for now, handle through email support
[...]

PR #4306, opened 2017-01-24
Add signin + copy to anon package view

whit537:
Implements first mock from https://github.com/gratipay/gratipay.com/issues/4297#issuecomment-273854209.

----------------------

screen shot 2017-01-19 at 1 14 43 pm

----------------------

screen shot 2017-01-24 at 5 40 10 pm


whit537:
Ready for review, @JessaWitzel @mattbk @nobodxbodon.

whit537:
Note that the base is #4305 and not `master`, so we should be good to merge even though the final state is incomplete.

www/on/npm/%package/index.html.spt
```diff
@@ -27,4 +27,15 @@ url = 'https://npmjs.com/package/' + package.name
 {% endblock %}
 
 {% block content %}
+{% if user.ANON %}
+<h2>{{ _( '{npm_package} has not been claimed on Gratipay.'
+        , npm_package='npm/' + package_name
```
nobodxbodon: Seems it's not highlighted as link in screenshot. But I don't want to block us on this.

mattbk: Is that because it's <h2>?

whit537: Ah, good call! Fixed in #4307.

PR #4309, opened 2017-01-25
UI to send confirmation email for unclaimed packages

whit537:
Part of https://github.com/gratipay/gratipay.com/pull/4305.


whit537:
The model for this is the email verification workflow.

whit537:
Oops. We'll see what Travis does with that `set_trace`. :-) Probably just time out ...

www/~/%username/emails/modify.json.spt
```diff
@@ -36,6 +37,11 @@ elif action == 'set-primary':
     participant.update_email(address)
 elif action == 'remove':
     participant.remove_email(address)
+elif action == 'add-email-and-claim-package':
+    package_id = request.body['package_id']
+    package = Package.from_id(package_id)
```
whit537: What if it's `None`?

whit537:
Sooooo ... I've been working on teaching the claim form about when there is only one address, but the implication of https://github.com/gratipay/gratipay.com/pull/4305#issuecomment-278429627 is that there will only ever be zero or one address(!).

whit537:
Yay for talking to users? :)

whit537:
Though I guess that's not necessarily true for all package managers, and since we're already mostly done with accounting for multiple emails (it would be work to change the schema at this point) then we may as well go for that. But on the other hand if we want to specify where the email is coming from, that will be different for each package manager.

whit537:
On the other hand we're trying to get this done once for npm. Future expansion is out of scope until we see traction from npm.

whit537:
Ready for review, @mattbk @nobodxbodon @JessaWitzel et al.

whit537:
screen shot 2017-02-08 at 5 56 01 pm

------------------

screen shot 2017-02-08 at 5 56 43 pm

----------------------

screen shot 2017-02-08 at 5 57 30 pm

www/on/npm/%package/index.html.spt
```diff
@@ -26,16 +27,43 @@ url = 'https://npmjs.com/package/' + package.name
...
+    <p>{{ _( 'Is this yours? You can claim it on Gratipay using any email address {a}on file{_a} in the maintainers field in the npm registry.'
+           , a=('<a href="https://registry.npmjs.com/' + package.name + '">')|safe
+           , _a='</a>'|safe
+            ) }}
+    {% if nemails == 0 %}
+    <p class="sorry">{{ _("Sorry, we didn't find any email addresses on file.") }}</p>
+    {% else %}
+    <input type="hidden" name="package_id" value="{{ package.id }}">
+    <ul>
+        {% for i, email in enumerate(package.emails) %}
```
nobodxbodon: Shall we [hide the bullets](http://stackoverflow.com/questions/1027354/need-an-unordered-list-without-any-bullets) in this email list?

tests/py/test_www_npm_package.py
```diff
@@ -4,10 +4,33 @@
...
+
+    def test_auth_gets_send_confirmation_page_from_unclaimed(self):
+        self.make_participant('bob', claimed_time='now')
+        body = self.client.GET('/on/npm/foo/', auth_as='bob').body
+        assert 'npm/foo</a> has not been claimed' in body
+        assert 'using any email address' in body
+        assert 'alice@example.com' in body
+
+    def test_auth_gets_multiple_options_if_present(self):
+        self.make_package('npm', 'bar', 'Bar', ['alice@example.com', 'alice@example.net'])
```
nobodxbodon: nit: use constant like 9c05276?

whit537: Done in 3586969370852adab37abfef8fb70db3f2ebfc66.

www/on/npm/%package/index.html.spt
whit537: Done in 455f06c67e56027839582c62ceeca8bab12b5351.

screen shot 2017-02-09 at 8 53 14 am

whit537:
Kicked Travis, green now. We good here?

PR #4308, opened 2017-01-25
Add a test for anon hitting unclaimed package

whit537:
Follow-up to #4306.

PR #4307, opened 2017-01-25
Add link to package on npm

whit537:
Also cleans up link styles left over from when we were going to render the entire README for packages.

PR #4323, opened 2017-02-03
Fix regression in test

whit537:
This snuck in when we merged #4308 after #4307 without a rebase.


whit537:
Ready for review/merge, @nobodxbodon @mattbk et al.

Issue #4327, opened 2017-02-09
Find new TLS certificate vendor


clone1018:
This morning I'm getting an SSL error for Gratipay.com
![image](https://cloud.githubusercontent.com/assets/226638/22790898/3c41047c-eead-11e6-9480-c2f8aef9c07f.png)

I'm running Google Chrome Version 57.0.2987.21 beta (64-bit)

clone1018:
From [Wikipedia](https://en.wikipedia.org/wiki/StartCom):

> On 24 October 2016, Mozilla announced on its security blog that, following its discovery of the purchase of StartCom by another Certificate Authority called WoSign during its investigation on numerous issues with that CA, and that both have failed to disclose this transaction,[12] Mozilla will stop trusting certificates that are issued after 21 October 2016 starting with Firefox 51.[13] On 1 November 2016, Google announced that it too would stop trusting certificates issued after 21 October 2016 starting with Chrome 56.[14] On 30 November 2016, Apple products will block certificates from WoSign and StartCom root CAs if the "Not Before" date is on or after 1 Dec 2016 00:00:00 GMT/UTC. [15]

whit537:
Good catch, @clone1018. Here is [Google's announcement](https://security.googleblog.com/2016/10/distrusting-wosign-and-startcom.html) (3+ months ago).

clone1018:
@whit537 One thing to remember, if migration is the solution CloudFlare does offer SSL

EdOverflow:
A dedicated SSL certificate on Cloudflare costs 5$ a month.

whit537:
Re: Cloudflare: https://github.com/gratipay/inside.gratipay.com/issues/957.

whit537:
Also: https://letsencrypt.org/.

whit537:
https://blog.heroku.com/announcing_heroku_free_ssl_beta_and_flexible_dyno_hours

clone1018:
I've used LetsEncrypt the it was very nice. I wonder if there's a stigma to using a free SSL provider for a payment website

whit537:
SNI should be fine from a browser support pov:

> ### Desktop Browsers
>
> - Firefox 2 and above
> - Internet Explorer 7 on Windows Vista and above
> - Windows Vista or OS X 10.6 with:
> - Chrome 5.0.342.0
> - Opera 14
> - Safari 4
>
> ### Mobile Browsers
>
> - Mobile Safari on iOS 4.0 and above
[...]

PR #4331, opened 2017-02-09
Rise to the ACME challenge

whit537:
StartSSL is hosed, so we need a new TLS cert vendor (https://github.com/gratipay/gratipay.com/issues/4327). Let's Encrypt suggests itself, but we need a way to pass verification challenges for multiple domains. Let's build a way!

### DIFF
--- a/gratipay/main.py
+++ b/gratipay/main.py
@@ -98,6 +98,7 @@ algorithm.functions = [
     algorithm['parse_environ_into_request'],
     algorithm['parse_body_into_request'],
 
+    utils.help_aspen_find_well_known,
     utils.use_tildes_for_participants,
     algorithm['redirect_to_base_url'],
     i18n.set_up_i18n,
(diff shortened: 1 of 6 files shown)


whit537:
Appears to be a hang-up with a (root?) path part starting with `.`. Investigating ...

whit537:
Aspen isn't matching the hidden directory, so it's falling into the [project &rarr; ~user fallthrough](https://github.com/gratipay/gratipay.com/blob/2046/www/%25team/%25sub.spt).

whit537:
screen shot 2017-02-09 at 8 29 44 pm

----------------------------

screen shot 2017-02-09 at 8 29 52 pm

----------------------------

screen shot 2017-02-09 at 8 29 59 pm

whit537:
Ready for review, @clone1018 et al.

clone1018:
Good solution! Waiting on travis.

PR #4333, opened 2017-02-10
Extend the ACME challenge system

whit537:
- mount read-only under /assets/
- display a count for admins on ./

Ready for review, @clone1018 et al.


clone1018:
Looks good, waiting on travis

Issue #4322, opened 2017-02-02
Localhost not loading in Firefox


kaguillera:
Just found this problem in Firefox while setting up Gratipay locally on @dmk246 laptop. For some reason the page never loads when you `make run` and try to open localhost:8537 in Firefox it hangs. We believe it is because `gratipay.report_uri.io`

whit537:
Indeed.

![screen shot 2017-02-16 at 1 34 35 pm](https://cloud.githubusercontent.com/assets/134455/23035321/ab821156-f44c-11e6-995b-49716ed4ec74.png)

whit537:
Why do we have a CSP violation in stock Gratipay?

![screen shot 2017-03-24 at 11 06 05 am](https://cloud.githubusercontent.com/assets/134455/24300167/edae5a34-1081-11e7-99b2-d4faea310637.png)

whit537:
Chrome also recognizes violations. Presumably their implementation precludes blocking on same.

![screen shot 2017-03-24 at 11 12 34 am](https://cloud.githubusercontent.com/assets/134455/24300428/cd8c3522-1082-11e7-8c38-ad8de8dc780d.png)

EdOverflow:
> The Content Security Policy directive 'upgrade-insecure-requests' is ignored when delivered in a report-only policy.

The `upgrade-insecure-requests` directive is ignored when sent via a `Content-Security-Policy-Report-Only` header: https://www.w3.org/TR/upgrade-insecure-requests/#delivery

> Unrecognized Content-Security-Policy directive 'reflected-xss'.

The `reflected-xss` directive has been deprecated in Chrome 56+: https://www.chromestatus.com/feature/5769374145183744

> Refused to apply inline style because it violates the following Content Security Policy directive: [...]

```html
<a class="github-ribbon" href="https://github.com/gratipay/gratipay.com">
   <img style="position: absolute; top: 0; right: 0; border: 0;"
      src="https://s3.amazonaws.com/github/ribbons/forkme_right_darkblue_121621.png"
      alt="Fork me on GitHub"> <!-- This line is the issue! -->
</a>
```

whit537:
Ah! So *that's* what @nobodxbodon was [on about](https://github.com/gratipay/gratipay.com/issues/4359#issuecomment-287519144). :D

EdOverflow:
Yep. We must remove the inline style attribute from the GitHub banner.

whit537:
[...]

PR #4335, opened 2017-02-14
Send confirmation email

whit537:
Part of #4305.

### Requirements

(7) and (8) from the [workflow](https://github.com/gratipay/gratipay.com/issues/4297#issuecomment-273900610):

> (7) receive email
> (8) click link

### Todo

- [x] fix these three failing tests
- [x] write Python API w/ tests & docs—should put mail on the queue
  - [x] ~~make sure we don't allow connecting the same package to two accounts~~<br>that happens on `finish-claim`, not `start-claim`
  - [x] make sure we're validating the address, can't send to just anyone
  - [x] extend `start_email_verification` to take packages
    - [x] what about packages in the except case?
  - [x] extend `add_email` to take packages
- [x] what happens to old `claims` when nonce is updated out from under it?
- [x] modify `modify.json.spt` w/ tests—should call into Python
  - [x] consider `LazyResponse`
- [x] write the email template
- [x] ~~should notify others listed as maintainers?~~ no, too much for now
- [x] can we inject via npm package name?
- [x] don't say we're connecting when we're not (ancillary accounts)
- [x] ~~should be able to start a verification for a different package w/ same email and have it not clobber the first (added? in addition?)~~ https://github.com/gratipay/gratipay.com/pull/4335#issuecomment-290554628


whit537:
Rebased, was f3e657ffef9b9f07855195467189a366ccbe69e2.

whit537:
Gosh, this would be easier to implement if we kept it to two steps:

- add email
- claim package

But that makes it harder for users.

whit537:
The natural thing for a user is to claim the package and add the email in one fell swoop.

whit537:
Our email infrastructure is pretty squishy. I started really getting into it yesterday, and I'm trying to decide how deep to go.

whit537:
Rebased.

whit537:
> I started really getting into it yesterday, and I'm trying to decide how deep to go.

Went pretty deep! 😁

whit537:
Alright, what the heck am I doing here?

whit537:
What's the end result here? What behavior?

/me reviews ticket and parent ...

whit537:
screen shot 2017-03-22 at 1 49 47 pm

whit537:
(I updated the ticket description with requirements.)

whit537:
Blorg. Right. Gotta think through adding email vs. mailing before adding. Current infrastructure assumes ... wait, does it? Assume we're only ever sending to `participant.email_address`? But then how does adding non-primary emails work?

whit537:
Ah ... is the `To` specific in the `context` in that case?

whit537:
[Yeah](https://github.com/gratipay/gratipay.com/blob/2065/gratipay/models/participant/email.py#L108), [that's](https://github.com/gratipay/gratipay.com/blob/2065/gratipay/email.py#L152) [it](https://github.com/gratipay/gratipay.com/blob/2065/gratipay/email.py#L170).

whit537:
Existing [`verification_notice`](https://github.com/gratipay/gratipay.com/blob/2065/emails/verification_notice.spt) should be good enough for the package claiming case as well.

whit537:
We can mostly use the existing `add_email` flow. The caveat is that we should send a mail even when the email already exists.

whit537:
I think even the existing `verification` template will work, we can just say "connect **npm/aspen and chad@zetaweb.com** to ~whit537" as appropriate.

whit537:
Gotta figure out how nonces are gonna work and how they will relate to packages. I think we may want a new table to track how nonces relate to packages, because nonces can either represent zero or one package, and down the road we're thinking they can also represent N packages in the case where someone wants to link multiple packages at once.

whit537:
But first let's get these test fixed ...

whit537:
Harumph. The trick here is to allow for sending a verification for package ownership even when the address itself is already verified, *and* not trip on our try/except/recurse log in `start_email_verification`.

whit537:
Topped up. Looking at #4359 &rarr; #4322 before proceeding further ...

whit537:
Returning from #4391 ...

whit537:
Rebased to pick up #4391.

whit537:
Alright, so I'm pretty sure the try/except/recurse wants to be tighter in, only around nonce creation vs. package linking. What else?

whit537:
Alright, so what are the conditions here? I'm to the point where I need to validate `email, *packages` when there's always an email and there may or may not be `packages`.

whit537:
We also need to think about package claiming across accounts.

|  | not claimed | half-claimed | already claimed |
|-------|--------|----------|----------|
| email |   |    |   |
| package (each) |   |   |   |

whit537:
If there aren't packages, we want to keep doing what we've been doing: block if the email is already verified either by the participant in question or someone else, or if the participant already has 10 emails.

If there _are_ packages, ... then what?

whit537:
Theeeennnnnn .......

whit537:
Then we want to block if:

- the email address is already verified for someone else
- the email address is not on file for all of the packages
- ...? is that it?

whit537:
I've confirmed that 2037a2f2cef199e493cb50818db80be30e6224ba properly extracts i18n messages, by running `make i18n` and inspecting `i18n/core.pot`.

whit537:
Okay! I've got this pretty close. I'm going to start working on the email copy. 😀

whit537:
Okay, phew! Email template in de1a0a32399173bc027b9242ab5da67197856a11. Let's see what it looks like live! 😮

After that there's one more edge case to consider:

> should be able to start a verification for a different package w/ same email and have it not clobber the first (added? in addition?)

whit537:
💃

screen shot 2017-03-30 at 5 19 38 pm

whit537:
screen shot 2017-03-30 at 5 21 46 pm

whit537:
However, I'm not getting the actual verification link on chad+foo@zetaweb.com. Hmm ... :thinking:

whit537:
Oh. Probably because I entered `chad+foo@example.com`. 😊

whit537:
💃 

screen shot 2017-03-30 at 5 29 34 pm

whit537:
screen shot 2017-03-30 at 5 30 11 pm

whit537:
> After that there's one more edge case to consider:
>
>> should be able to start a verification for a different package w/ same email and have it not clobber the first (added? in addition?)

Actually, I think clobbering is fine to start with here. We've got all the wiring we need to implement a UI for verifying multiple packages at once, which will reduce the incentive to trigger multiple separate verification flows in parallel. I'm comfortable erroring out the first flows in that case. Do we do that?

whit537:
Alright, started a test but it's bleeding into the next task, which is modifying the `verify.html` endpoint to receive package claims in addition to address verifications, so let's save that for the next PR.

```diff
diff --git a/tests/py/test_email.py b/tests/py/test_email.py
index 07ce431..292d764 100644
--- a/tests/py/test_email.py
+++ b/tests/py/test_email.py
@@ -286,6 +286,22 @@ class TestEndpoints(Alice):
         assert response.code == 400
         assert self.db.all('select package_id from claims order by package_id') == []
 
+    def test_starting_a_second_claim_clobbers_the_first(self):
+        foo = self.make_package()
+        start = lambda: self.hit_email_spt( 'start-verification'
+                                          , 'alice@example.com'
+                                          , package_ids=[foo.id]
+                                           )
+        start()
+        old_nonce = self.alice.get_email('alice@example.com').nonce
+        start()
+        new_nonce = self.alice.get_email('alice@example.com').nonce
+        one = self.verify_email('alice@example.com', old_nonce).body
+        two = self.verify_email('alice@example.com', new_nonce).body
+        import pdb; pdb.set_trace()
+        assert 'Bad Info' in one
+        assert 'Success!' in two
+
 
 class TestFunctions(Alice):
 
```

whit537:
And with that, I am submitting this for review!

Ping @rohitpaulk @nobodxbodon @mattbk @JessaWitzel @aandis @clone1018.

This is the next PR in our long march towards claiming npm packages (#4305). How does it look? :-)

whit537:
Rebased, was de1a0a32399173bc027b9242ab5da67197856a11.

kaguillera:
LGTM...off to the second reviewr

whit537:
screen shot 2017-04-12 at 5 32 25 pm

gratipay/models/participant/email.py
```diff
@@ -41,94 +41,172 @@ class Email(object):
...
-                                     )
+        return nonce
+
+
+    def start_package_claims(self, c, nonce, *packages):
+        """Takes a cursor, nonce and list of packages, inserts into ``claims``
+        and returns ``None``.
+        """
+        if not packages:
+            return
```
dowski: Is this a normal condition? If not, consider logging or throwing an exception. Seems odd to start a package claim on an empty list of packages.

gratipay/models/participant/email.py
```diff
@@ -41,94 +41,172 @@ class Email(object):
...
+        VALUES, values = [], []
```
dowski: Is this Gratipay convention, to use CAPS for format strings? Normally I'd expect that to be a constant.

gratipay/models/participant/email.py
```diff
@@ -41,94 +41,172 @@ class Email(object):
...
+        for p in packages:
+            VALUES.append('(%s, %s)')
+            values += [nonce, p.id]
+        VALUES = ', '.join(VALUES)
```
dowski: It seems tricky to switch types of a variable from line to line. Maybe that's the Java coder in me though. ;-)

tests/py/test_email.py
```diff
@@ -20,14 +22,35 @@ def setUp(self):
...
 class TestEndpoints(Alice):
 
-    def hit_email_spt(self, action, address, user='alice', should_fail=False):
+    def hit_email_spt(self, action, address, user='alice', package_ids=[], should_fail=False):
         f = self.client.PxST if should_fail else self.client.POST
-        data = {'action': action, 'address': address}
-        headers = {b'HTTP_ACCEPT_LANGUAGE': b'en'}
-        return f('/~alice/emails/modify.json', data, auth_as=user, **headers)
+
+        # Hack to work around Aspen test client limitations.
```
dowski: What limitation?

emails/verification.spt
```diff
@@ -1,16 +1,63 @@
 {{ _("Connect to {0} on Gratipay?", username) }}
 
 [---] text/html
-{{ _("We've received a request to connect {0} to the {1} account on Gratipay. Sound familiar?",
-     ('<b>%s</b>'|safe) % email,
-     ('<b><a href="https://gratipay.com/~{0}/">{0}</a></b>'|safe).format(username)) }}
+{% if new_email_verified %}
+{{ ngettext( "We've received a request to connect the {package_name} npm package to the "
+             "{username} account on Gratipay. Sound familiar?"
```
dowski: What if it doesn't sound familiar?

What if it does, for that matter? I don't really know what I would do if I received an email like this.

Am I supposed to click the package name?

gratipay/models/participant/email.py
```diff
@@ -41,94 +41,172 @@ class Email(object):
...
+        c.run('INSERT INTO claims (nonce, package_id) VALUES ' + VALUES, values)
```
dowski: I'm generally unfamiliar with the Gratipay DB environment. Does anything exist to clean up claims that lay around for a long time?

emails/verification.spt
whit537: Down on line 32 there's a "Yes, proceed!" button. Here's what it looks like:

screen shot 2017-04-13 at 8 41 55 am

dowski: Ah, woops - didn't see that! Thanks!

tests/py/test_email.py
whit537: Comment verbosified in 136591a2074cb902df7853de632b87cbff6986ef.

gratipay/models/participant/email.py
whit537: Addressed in 2c2a3a51116cf617e472431178f3cc436822a8bc. For large inputs the `', '.join` idiom is more performant than string concatenation due to Python implementation details, but our inputs here are not large (order of magnitude 100 packages max to be claimed at once).

gratipay/models/participant/email.py
whit537: Done in 2c2a3a51116cf617e472431178f3cc436822a8bc. 👍

gratipay/models/participant/email.py
dowski: FWIW, I didn't have a problem with .join(), it was just my internal type checker twitching at VALUES starting as a list of strings and then getting reassigned to a string.

gratipay/models/participant/email.py
```diff
@@ -177,16 +178,20 @@ def get_email_verification_nonce(self, c, email):
...
+            raise NoPackages()
+
+        # We want to make a single db call to insert all claims, so we need to
+        # do a little SQL construction. Do it in such a way that we still avoid
+        # Python string interpolation (~= SQLi vector).
+
+        extra_sql, values = '', []
         for p in packages:
-            VALUES.append('(%s, %s)')
+            extra_sql += ' (%s, %s)'
```
dowski: Looks like maybe this isn't going to be delimited by commas anymore?

" (%s, %s) (%s, %s) (%s, %s)"

Maybe that's OK for Postgres though?

whit537: Good point. The few tests I ran passed. Lessee here ...

whit537: Yay CI! :-)

```
    def execute(self, query, vars=None):
        self.Record = None
>       return super(NamedTupleCursor, self).execute(query, vars)
E       ProgrammingError: syntax error at or near "("
E       LINE 1: ...LUES ('208503b2-9283-4bc8-ab51-6ed60c28c9a3', 30) ('208503b2...
E  
```

https://travis-ci.org/gratipay/gratipay.com/builds/221772022#L2879-L2884

Fixing ...

whit537: You okay with 992698534d07424fd60f123bf1c7a040d54f5471?

gratipay/models/participant/email.py
whit537: No! Reticketed as https://github.com/gratipay/gratipay.com/issues/4408.

gratipay/models/participant/email.py
whit537: P.S. @kaguillera thinks this _is_ the Java coder in you. 😛

whit537:
With https://github.com/gratipay/inside.gratipay.com/issues/1038#issuecomment-293990454 you should be cleared to merge, @dowski. Got a green button?

whit537:
Woo-hoo! 💃

PR #4345, opened 2017-02-22
Start an Application object

whit537:
This creates a new `Application` object that subsumes `Website` and should eventually eat all of `wireup`(!). ~~Immediate next step is to make an `Email` mixin for it: #4346.~~ We need to make this available to payday and models before we can work on email: #4350.

### Todo

- [x] change base to `master`

PR #4346, opened 2017-02-22
Move some email API

whit537:
This decouples actual sending of email from the `Participant` object, which should only ever _queue_ email to be sent.

### Todo

- [ ] change base to master


whit537:
Alright, I need to split out tests against `gratipay.application.email` from those against `gratipay.models.participant.email`, and modify the latter to `get_last_email` from `email_queue` rather than intercepting calls to `send_email` via mock.

whit537:
Basically we should be able to isolate tests against the _mailer to a minimum, and run almost everything through the queue.

whit537:
These PRs are a little big but I'm not sure how to make them smaller.

www/~/%username/emails/modify.json.spt
```diff
@@ -27,11 +27,18 @@ if not participant.email_lang:
 
 msg = None
 if action in ('add-email', 'resend'):
-    r = participant.add_email(address, website.env.resend_verification_threshold)
-    if r:
-        msg = _("A verification email has been sent to {0}.", address)
+    try:
+        participant.add_email(address, website.env.resend_verification_threshold)
+    except EmailTaken:
```
whit537: Apparently we're not testing the exceptions here, because I just discovered down below #4348 that `EmailTaken` is never imported here.

gratipay/models/participant/email.py
```diff
@@ -206,8 +215,8 @@ def remove_email(self, address):
 
 
     def queue_email(self, spt_name, **context):
```
nobodxbodon: I wonder why queue_email and dequeue_email(s) are not in the same class?

whit537: Fair enough. Turns out this will take some further refactoring, though. Forthcoming ...

whit537:
Rebased, was fa3fa6ceaba5138f0729663b7319e59d736b9046.

gratipay/models/participant/email.py
whit537: Paved the way in https://github.com/gratipay/gratipay.com/pull/4350.

whit537:
Hmmm ... working on rebasing this onto the latest #4350, and it looks like we need to unravel and remove `gratipay.main` as a next step. Perhaps all the way back in #4345? I think I had some commits at some point on some local branch that moved `gratipay/main.py` to a `gunicorn_entrypoint.py` and then instantiated `Application` in `gratipay.testing.harness` vs. importing from `main`. To be continued ... !

whit537:
Rebased, was b305600f08477ec37d6eecbe4eea0f783dc24f76.

www/~/%username/emails/modify.json.spt
whit537: We're testing this now (not sure exactly where, but I saw failures related to this and fixed them.)

gratipay/models/participant/email.py
whit537: > some

That was an understatement. :-)

With #4350 and #4355, we now have queueing and flushing in the same class.

whit537:
Rebased on `master` post-#4355, was 260972325f1a465fe0b18434e2c26352f87bd55f.

whit537:
Self-merged per https://github.com/gratipay/gratipay.com/pull/4305#issuecomment-286796283 and https://github.com/gratipay/gratipay.com/pull/4360#issuecomment-287064268.

PR #4353, opened 2017-02-25
Show projects on profile like on index and search 

### LINKED ISSUE #4133: Team list for users with multiple teams is poorly spaced

mattbk:
This is a different direction than #4169, in order to keep things looking similar.

Will close #4133 as well.

todo
- [x] add logic back in about unreviewed or closed projects
- [x] fix tests
- [x] make the who-can-see-this logic work right again
- [x] check admin can see everything
- [x] fix width for viewing with sidebar
- [x] how to list projects on your own profile?
- [x] what's up with scrunching? https://github.com/gratipay/gratipay.com/pull/4353#issuecomment-286432362
- [x] find a better way to deal with long names
- [x] adapt to presence of sidebar
- [x] are we using the same template everywhere?
  - [x] homepage
  - [x] search
  - [x] profile
- [x] code okay?


mattbk:
Still need to get the spacing right.

screen shot 2017-02-24 at 8 44 02 pm

mattbk:
Width is figured out.
image

mattbk:
Ready for review @JessaWitzel @nobodxbodon et al.

templates/team-listing.html
```diff
@@ -1,18 +1,46 @@
 {% set unprivileged = not(participant == user.participant or user.ADMIN) %}
 {% set approved_open_teams = participant.get_teams(unprivileged, unprivileged) %}
-
-{% if (user.ADMIN or user.participant == participant) and approved_open_teams %}
+{% if (user.ADMIN or user.participant == participant) or approved_open_teams %}
```
whit537: Currently we show *all* teams to admins and owners, whether or not they're approved or open. We only constrain for non-admin/owner. I think we want to preserve that behavior.

templates/team-listing.html
```diff
@@ -1,18 +1,46 @@
...
+                        <tr>
+                            <td class="label">{{ _("Receiving") }}</td>
+                            <td>{{ format_currency(team.receiving, 'USD')
+                                    if team.receiving else '- &nbsp;'|safe }}</td>
+                            <td>{{ team.nreceiving_from if team.nreceiving_from else '-' }}</td>
+                        </tr>
+                    </table>
+                {% endif %}
+            </td>
+        </tr>
```
whit537: This is the third time we're copying and pasting this markup: it's time to factor it out into a snippet under `templates/` and use that DRY code from all three: homepage, search, and here.

templates/team-listing.html
```diff
@@ -1,18 +1,46 @@
...
-    <ul class="team memberships">
-        {% for team in approved_open_teams %}
-        <li><a href="/{{ team.slug }}/">{{ team.name }}</a></li>
-        {% endfor %}
-    </ul>
+    {% if user.participant == participant %}
+        <form action="/new" class="apply">
+            <button type="submit">{{ _("Fund Your Project") }}</button>
+        </form>
+    {% endif %}
```
whit537: Not a bad idea to include this cta here, but if so we should show it in the case where there are zero projects as well.

www/~/%username/index.html.spt
```diff
@@ -27,6 +27,18 @@ def with_others(obj):
...
+                  , "unreviewed" : _("Unreviewed")
+                  , "rejected": _("Rejected")
+                  , "featured": _("Featured")
+                   }
+
+status_icons = { "approved": "&#xe008;"
+               , "unreviewed": "&#xe009;"
+               , "rejected": "&#xe010;"
+               , "featured": "&#xe9d9;"
+                }
```
whit537: As above, can we factor these out for reuse in all three places?

scss/components/listing.scss
```diff
@@ -56,6 +56,10 @@ table.listing {
             &.room-for-numbers {
                 padding-right: 80px;
             }
+            &.profile {
+                max-width: 384px;
+            }
```
whit537: I'm not convinced this is the right solution. Hover doesn't clobber the left lower metadata on the homepage:

screen shot 2017-03-02 at 12 41 53 pm


but it does here:

screen shot 2017-03-02 at 12 42 02 pm

whit537:
Back to you, @mattbk.

templates/team-listing.html
mattbk: It should show only to the owner, even if zero. That's what I thought I did.

whit537: Okay, maybe you did, sorry. :)

mattbk:
I realize that the names I'm using aren't the best choices, but I'm pumped to get something like this working.

gratipay/utils/icons.py
```diff
@@ -0,0 +1,10 @@
+
+
+def statusicons():
```
whit537: Doesn't even need to be a function, can just be a top-level constant, in which case all-caps is customary.

```
STATUS_ICONS = {}
```

mattbk:
1. Set a variable to use in the for loop, same name each time
1. Put the for loop in the template 
1. Leave the table definition in the simplate

Played with a different way but scrapping it.

mattbk:
Come on, CSS...

mattbk:
I think I got it. It's not great with long names though:
image

mattbk:
Gave it a few more pixels:
image

whit537:
Rebased, was 2d140fe49ee20f2c9b1f4b83c438a05e3b7bfc09.

whit537:
Still way truncated compared to the homepage. Looking into it ...

screen shot 2017-03-14 at 10 09 14 am

# vs.

screen shot 2017-03-14 at 10 09 38 am

mattbk:
Doh! That's because I think I adjusted CSS based on it being a profile page and not worrying about logged in/admin or not.

templates/project-listing.html
```diff
@@ -0,0 +1,44 @@
+{% set current_page = request.path.raw.split('/')[1] %}
+{% for i, project in enumerate(project_list, start=1) %}
+    <tr>
+        <td class="item">
+            <img src="{{ project.get_image_url('small') }}">
+            <a class="name" href="/{{ project.slug }}/">{{ project.name }}</a>
+
+            <div class="details{% if project.status == 'approved' %} room-for-numbers{% endif %}{% if current_page == '~' %} profile{% endif %}">
```
mattbk: E.g., this needs to check for user.participant or user.ADMIN as well as current_page.

mattbk:
Long names still overlap. That's what changing the height in https://github.com/gratipay/gratipay.com/pull/4353/commits/bf0d122e3554eb8b2f8fa8ce5f6066cfcc960e3e#diff-442ad2cb7479f6fcf5df36e254dbf461L12 was for.

mattbk:
66px is the minimum, but there will still be overlap if something like this happens:

image

whit537:
I'm reluctant to change the height because of the effect on the subtle alignment between the elements. I've been looking at adding the ellipsis in the name as well as the details, or possibly putting a white glow on the numbers table that would cause a fade-out effect. We could also possibly modify the font size to fit better—the social profile username text is smaller, e.g., but then we're back into thinking more carefully about vertical alignment. Unresolved imo.

mattbk:
I'm okay with leaving it as it is. The only people who are going to see the sidebar-present version are the project owners and admins.

whit537:
I just noticed something else: the edge of the hover target is way out to the right of the listing item. This is true [on production](https://gratipay.com/search?q=open+source+solution), but could maybe be cleaned up while we're here.

---------

kgdvqvbbvp

whit537:
Can you live with 2c33e30b790c043af20b53b5cf7ad1e2aaa8335e, @mattbk?

# Before

screen shot 2017-03-15 at 11 41 47 am

# After

screen shot 2017-03-15 at 11 41 53 am

mattbk:
Sure. That's going to affect the front page too, btw.

whit537:
Right, another option would be to keep the larger name and instead fold the numbers table into the details row, like so:

screen shot 2017-03-15 at 11 52 52 am

whit537:
> That works for now, but isn't the idea to add back in the distributing numbers in the future?

Hrm ... yeah, I guess. Let's cross that bridge when we come to it? ☺️

With deemphasized detail links:

screen shot 2017-03-15 at 11 59 02 am

mattbk:
Also, 2c33e30 doesn't affect the baseline, it just makes collision less likely due to the name taking up less horizontal space.

whit537:
> it just makes collision less likely due to the name taking up less horizontal space.

Correct.

mattbk:
> With deemphasized detail links:

Let's try that one.

whit537:
>> it just makes collision less likely due to the name taking up less horizontal space.
>
>Correct.

Though that's moot if we move receiving/givers into `.details`.

whit537:
Alright, I'm ready here if you are, @mattbk.

PR #4350, opened 2017-02-23
Use Application in Payday

whit537:
Properly speaking, we should have zero global configuration. Everything should hang from a `Gratipay` object that we instantiate for gunicorn and in scripts. But for https://github.com/gratipay/gratipay.com/pull/4346#discussion_r102598587 we at least need to make the `application` object accessible from within `Payday`.

PR #4355, opened 2017-02-27
Further Application improvements

whit537:
Picking up from #4345:

- Clear out `gratipay/__init__.py` as part of moving away from global config.
- Make `.app` available on models.
- Simplify `website.version`.

### DIFF
--- a/gratipay/__init__.py
+++ b/gratipay/__init__.py
@@ -1,39 +1,4 @@
+# -*- coding: utf-8 -*-
 """This is the Python library behind gratipay.com.
 """
-import locale
-from decimal import Decimal
-
-
-try:  # XXX This can't be right.
-    locale.setlocale(locale.LC_ALL, "en_US.utf8")           # Heroku
-except locale.Error:
-    import sys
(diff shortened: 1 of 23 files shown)


whit537:
Rebased post-#4350, was 56a72ce3e05fed70ca943f76a125cc294a7378a4.

whit537:
Not sure why this is red. I'm double-checking that master is actually okay post-#4350.

whit537:
`master` is actually okay.

whit537:
There was a bug in the second commit on here. Not sure where it snuck in, whether it was there originally or entered via the rebase. Fixed now in f9b0bab8eff698ed1a667841239f5d99b551046e (right, Travis?).

whit537:
Alright, all three commits are back on here so we can see them at Travis.

whit537:
Self-merged per https://github.com/gratipay/gratipay.com/pull/4305#issuecomment-286796283 and https://github.com/gratipay/gratipay.com/pull/4360#issuecomment-287064268.

PR #4358, opened 2017-02-28
Add "Referrer-Policy" header

EdOverflow:
Set the `Referrer-Policy` to `strict-origin-when-cross-origin`. This
should boost our grade on
https://securityheaders.io/?q=gratipay.com&hide=on&followRedirects=on.

### DIFF
--- a/templates/base.html
+++ b/templates/base.html
@@ -4,7 +4,6 @@
     {% block head_early %}{% endblock %}
     <title>{% if title %}{{ title }} - {% endif %}{% if banner %}{{ banner }} - {% endif %}Gratipay</title>
     <meta name="viewport" content="width=device-width, initial-scale=1.0">
-    <meta name="referrer" content="no-referrer-when-downgrade">
     <link rel="stylesheet" href="{{ website.asset('gratipay.css') }}" type="text/css">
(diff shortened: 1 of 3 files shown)


templates/base.html
```diff
@@ -4,7 +4,6 @@
     {% block head_early %}{% endblock %}
     <title>{% if title %}{{ title }} - {% endif %}{% if banner %}{{ banner }} - {% endif %}Gratipay</title>
     <meta name="viewport" content="width=device-width, initial-scale=1.0">
-    <meta name="referrer" content="no-referrer-when-downgrade">
```
EdOverflow: Sorry.

whit537: This is a redo of #4357, it looks like?

whit537: - switch from [`no-referrer-when-downgrade`](https://w3c.github.io/webappsec-referrer-policy/#referrer-policy-no-referrer-when-downgrade) to [`strict-origin-when-cross-origin`](https://w3c.github.io/webappsec-referrer-policy/#referrer-policy-strict-origin-when-cross-origin)
- set header instead of using `meta`
- test it :)

Am I reading this right?

EdOverflow: Spot on! 🙂

PR #4361, opened 2017-03-02
Start recording ref in exchanges table

kaguillera:
This should fix the problem of not recording the ref to the external system. In this PR only the braintree ref id is being recorded. @whit537 suggested that the Paypal MassPay reference be done in another PR. 
Please note that this first commit should fail since I have not updated the `test_billing_exchanges.py` as yet.


kaguillera:
I do believe that this is ready for review. @mattbk @rohitpaulk @nobodxbodon et.al

gratipay/billing/exchanges.py
```diff
@@ -82,6 +83,7 @@ def create_card_hold(db, participant, amount):
             'options': { 'submit_for_settlement': False },
             'custom_fields': {'participant_id': participant.id}
         })
+        ref = result.transaction.id
```
nobodxbodon: A general question: can ref here and below be null?

kaguillera: yes it can be according to the current state of the db table. For recording and accounting purposes we don't want it not to be null but because of the current issue with the missing back data we need to allow nulls until we can back fill all the refs (if that is at all possible).

nobodxbodon: To make sure, after this PR, all the inserted records afterwards will have ref field as non-null?

gratipay/billing/exchanges.py
```diff
@@ -74,6 +74,7 @@ def create_card_hold(db, participant, amount):
 
     hold = None
     error = ""
+    ref = ""
```
whit537: Why do we need to set this here? For when the transaction fails and we don't have a ref? Shouldn't we just stick with  null in that case?

gratipay/billing/exchanges.py
whit537: Do they give us a URL and not just a transaction id? Eventually we're going to want to cross-link from Gratipay (admin) UI over to Braintree (etc.). Transaction id is a start but if we can know a URL at this point then let's just store that, ya?

whit537: > all the newly inserted records afterwards will have ref field as non-null?

I believe they will/should be null if the transaction fails. We still want to record our attempt at a transaction even if the API call fails.

mattbk: Where would I look to learn more about what we can pull from Braintree?

kaguillera: https://developers.braintreepayments.com/guides/overview

gratipay/billing/exchanges.py
kaguillera: I needed to set this (but it could be Null) because it kept giving the error that the variable was being referenced before being assigned

gratipay/billing/exchanges.py
kaguillera: @whit537 I believe that it should be null if it fails in the `try...except` block but it may still have a reference in Braintree if fails on there side after the `card_hold` has been created.

whit537: >  it may still have a reference in Braintree if fails on there side after the `card_hold` has been created.

Hmmm ... we should probably differentiate "failure" state as reported by Braintree from an "error" state due to, e.g., network failure during the API call.

kaguillera: we can always do that in the note field of the exchange record

mattbk: Can we just build a URL to the transaction? E.g., Stripe builds URLs like https://dashboard.stripe.com/payments/ch_19tIgqD5FJNWkW8j1uWePP8O with the charge ID.

kaguillera: After a ridiculous long time researching it...yes we can build a URL to the transaction using `transaction.id`
I believe it would look something like
`https://dashboard.braintreegateway.com/merchants/gratipay/transactions/1vt8z5a3`

where `1vt8z5a3` is the `transaction.id`

gratipay/billing/exchanges.py
kaguillera: After some consideration...yes it should be `Null` if the `try..except` fails

kaguillera:
This last commit just changes the ref from an empty string to `Null` if the create_card_hold fails in the try_except. Any other failures would be on Braintree and thus a ref (transaction.id) should exist and thus would server as a record for us as the properties of the transaction. 

It should be noted that all of the transaction history is attached to the one id on creation of the transaction in the Braintree system. Refund transactions is the only action that has a different id from the origin transaction for which it is made. There is a field in the refund transaction that links to the original transaction for which the refund is being made. IRL @whit537 suggested that eventually we may want to record the origin transaction ids for refunds that we do in its own field. Currently it is recorded in the `notes` field of the exchange record. (or atleast should be)

kaguillera:
ok ready for review/merge

mattbk:
Merge conflict!

kaguillera:
Working on the conflict files no sure why it is saying that.

kaguillera:
I just can't figure out the conflict

mattbk:
I'm not sure what I just did. Was following the command line instructions mentioned above.

kaguillera:
I was trying not to do that since the commands merge the branch into master (at least that is what I think it does) and I did not want to do that until the conflict was resolved.

...
 Guess not

!m @mattbk

mattbk:
Yeah, I stopped once I had merged master into this branch, and then pushed from there. I think it was only the first block of commands.
```
git fetch origin
git checkout -b record-ref-in-exchanges origin/record-ref-in-exchanges
git merge master
```
and then 
`git push origin record-ref-in-exchanges`

PR #4366, opened 2017-03-05
Implement SameSite cookies. 🍪

EdOverflow:
`SameSite` attribute set to `Lax`.

https://tools.ietf.org/html/draft-west-first-party-cookies-07


whit537:
Build failures?

EdOverflow:
@whit357 I am not entirely sure how to solve this. This is not a priority so maybe I should have just opened an issue for SameSite cookies.

whit537:
Looks like a `SyntaxError`:

```
    set_cookie(cookies, SESSION, token, expires=new_expires, samesite)
SyntaxError: non-keyword arg after keyword arg
```

Are you familiar with the distinction between keyword and non-keyword arguments in Python?

EdOverflow:
Oh, whoops! I see the issue now.

EdOverflow:
It turns out there isn't a `samesite` morsel object in https://docs.python.org/2/library/cookie.html#morsel-objects.

PR #4374, opened 2017-03-10
Move hyphen to end of allowed characters regex for ~user and project slugs

### LINKED ISSUE #4295: Disallow slashes and other naughty strings in project names
Per some security researchers, we get project names like this:
- https://github.com/gratipay/project-review/issues/313 /www.evil.com/user_ip.php?gratipay
- https://github.com/gratipay/project-review/issues/312 //www.google.com/?><x>
- https://github.com/gratipay/project-review/issues/311 //www.google.com/?

All of these break the URL, so the project page can't be reached and needs to be manually removed from the database.

Not a security issue, but is an annoyance and could be used for spam.

mattbk:
Closes #4295. This disallows characters like `/` and `?` from being used in slugs, which fubars URLs.


whit537:
Gonna need a test or four for this. Ideally the tests would be on a separate commit, pushed first so we can see Travis fail.

mattbk:
Similar to testing the error message, but asserting a different response?
```
    def test_error_message_for_invalid_team_name(self):
        self.make_participant('alice', claimed_time='now', email_address='alice@example.com', last_paypal_result='')
        data = dict(self.valid_data)
        data['name'] = '~Invalid:Name;'
        r = self.post_new(data, expected=400)
        assert self.db.one("SELECT COUNT(*) FROM teams") == 0
        assert "Sorry, your team name is invalid." in r.body
```

mattbk:
E.g., 
```
    def test_invalid_team_name(self):
        self.make_participant('alice', claimed_time='now', email_address='alice@example.com', last_paypal_result='')
        data = dict(self.valid_data)
        data['name'] = 'invalid?name/goes\here?'
        r = self.post_new(data, expected=400)
        assert self.db.one("SELECT COUNT(*) FROM teams") == 0
        assert "Sorry, your team name is invalid." in r.body
```

I know this tests for the output...

mattbk:
Temporarily reverted ee11dfc in dff5b4b. Added a test. It should fail.

mattbk:
Success? 
image

mattbk:
I moved the hyphen in the ~username regex as well, but from what I can tell it was already working correctly.

EdOverflow:
Looks good to me.

whit537:
> I moved the hyphen in the ~username regex as well, but from what I can tell it was already working correctly.

Turns out it is a `set` and not a regex. 🐻

mattbk:
I should read more closely.

PR #4383, opened 2017-03-17
Convert Throttle to a 400

whit537:
https://sentry.io/gratipay/gratipay-com/issues/236526695/

### DIFF
--- a/tests/py/test_email.py
+++ b/tests/py/test_email.py
@@ -98,6 +98,14 @@ class TestEndpoints(Alice):
         response = self.hit_email_spt('add-email', ('a'*243) + '@example.com', should_fail=True)
         assert response.code == 400
 
+    def test_post_too_quickly_is_400(self):
+        self.hit_email_spt('add-email', 'alice@example.com')
+        self.hit_email_spt('add-email', 'alice+a@example.com')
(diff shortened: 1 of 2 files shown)


whit537:
e58fd184129eaf284d17149e1097e2c1b89acbee is a naive fix. Ideally, though, we'd make `Throttled` a subclass of `Response` since it'll always only ever wanna be a 400 (what other `Throttled`s are lurking?). The complication is i18n of the error message, which we only have easy access to in simplate context.

whit537:
What other places could leak `Throttled`?

Changaco:
> The complication is i18n of the error message, which we only have easy access to in simplate context.

That's what [`LazyResponse`](https://github.com/gratipay/gratipay.com/blob/2063/gratipay/utils/__init__.py#L270) is for. I have lots of those [in Liberapay](https://github.com/gratipay/gratipay.com/blob/2063/gratipay/utils/__init__.py#L270).

whit537:
I reviewed all of our `email_queue.put` calls and confirmed that this is the only place we need to catch this.

whit537:
Alright, ready to review @rohitpaulk @mattbk @nobodxbodon et al.

Issue #4359, opened 2017-02-28
Travis often hangs on ttw tests


whit537:
Not infrequently, a build will fail at Travis due to the 10 minute timeout. Restarting the build generally fixes it. From what I've observed, the tests hang on the ttw tests. I think I've been able to observe this behavior locally as well, so perhaps this is debuggable. In the mean time ... restart build!

whit537:
Ah! We *do* have an AWS S3 dependency: the GitHub ribbon.

AWS is [having an issue](https://status.aws.amazon.com/) with S3 right now:

> Update at 10:33 AM PST: We're continuing to work to remediate the availability issues for Amazon S3 in US-EAST-1. AWS services and customer applications depending on S3 will continue to experience high error rates as we are actively working to remediate the errors in Amazon S3.

(Their status reporting is terrible, so I can't link to a specific incident report.)

whit537:
GitHub failed to post my screenshot of our S3 dependency, leading me to wonder whether they have an S3 dependency as well. ;-)

whit537:
Yeah, ttw tests proceed with normal slowness when I remove the ribbon.

whit537:
Sounds to me like another reason to have a knob to turn off report-uri.io in development/CI.

whit537:
This is really slowing me down.

nobodxbodon:
Would [host github ribbon on our own](https://github.com/lukaszx0/queues.io/commit/4a8756494888442e8ab10f96b052ab537d806b02) help?

whit537:
I don't think it's the ribbon that is slowing us up, I think it's report-uri.io. Need to confirm though ...

whit537:
The thing is that we *want* to test report uri, but maybe not all the time.

whit537:
But really it's not like we ever want to block page loading on report-uri.com. Is there something wrong with our implementation?

whit537:
Perhaps related to #4322?

whit537:
Looks like in #4390 this is not due to the CSP violation from the GitHub ribbon. That actually makes sense, since that was affecting Firefox only, and we use PhantomJS (WebKit ~= Chrome) at Travis.

whit537:
I usually run the ttw tests locally using Chrome. Can I repro w/ PhantomJS?

whit537:
Hmm ... different than hang. :thinking: 

```
Traceback (most recent call last):
[...]

PR #4390, opened 2017-03-24
Fix bugs in CSP implementation

### LINKED ISSUE #4322: Localhost not loading in Firefox

whit537:
Closes #4322 ~~and~~ ... but not #4359. :-/

### DIFF
--- a/gratipay/security/__init__.py
+++ b/gratipay/security/__init__.py
@@ -43,7 +43,8 @@ def add_headers_to_response(response):
 
     # https://www.w3.org/TR/referrer-policy/
     if 'Referrer-Policy' not in response.headers:
-        response.headers['Referrer-Policy'] = 'strict-origin-when-cross-origin'
+        response.headers['Referrer-Policy'] = \
+                                      'no-referrer-when-downgrade, strict-origin-when-cross-origin'
 
     # https://developer.mozilla.org/en-US/docs/Web/HTTP/CSP
     if 'content-security-policy-report-only' not in response.headers:
@@ -53,8 +54,6 @@ def add_headers_to_response(response):
             "style-src 'self' assets.gratipay.com downloads.gratipay.com cloud.typography.com;"
             "img-src *;"
             "font-src 'self' assets.gratipay.com cloud.typography.com data:;"
-            "upgrade-insecure-requests;"
             "block-all-mixed-content;"
-            "reflected-xss block;"
             "report-uri https://gratipay.report-uri.io/r/default/csp/reportOnly;"
         )
--- a/scss/components/github-ribbon.scss
+++ b/scss/components/github-ribbon.scss
@@ -1,0 +1,6 @@
+.github-ribbon {
+    position: absolute;
+    top: 0;
+    right: 0;
+    border: 0;
+}
--- a/templates/base.html
+++ b/templates/base.html
@@ -31,8 +31,7 @@
 
     <div id="wrapper">
         <a class="github-ribbon" href="https://github.com/gratipay/gratipay.com">
-            <img style="position: absolute; top: 0; right: 0; border: 0;"
-                 src="https://s3.amazonaws.com/github/ribbons/forkme_right_darkblue_121621.png"
+            <img src="https://s3.amazonaws.com/github/ribbons/forkme_right_darkblue_121621.png"
                  alt="Fork me on GitHub">
         </a>
         <div id="header" class="clearfix">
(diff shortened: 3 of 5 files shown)


whit537:
Of course, we don't want to require people to use https on localhost. What's our workaround?

gratipay/security/__init__.py
```diff
@@ -54,7 +54,6 @@ def add_headers_to_response(response):
             "img-src *;"
             "font-src 'self' assets.gratipay.com cloud.typography.com data:;"
             "upgrade-insecure-requests;"
-            "block-all-mixed-content;"
```
EdOverflow: Why remove `block-all-mixed-content` and not `upgrade-insecure-requests`?

whit537: Good question. I guess the latter gives us the flexibility to serve http on localhost, ya?

whit537:
While we're here, should we drop `reflected-xss`?

> Unrecognized Content-Security-Policy directive 'reflected-xss'.

gratipay/security/__init__.py
whit537: Confirmed working locally with 39901b710356357cd1fe1325ea643fe6eb5a67fc. 👍

EdOverflow: I believe the reason why you are not supposed to combine both is because `upgrade-insecure-requests` already blocks all mixed content.

whit537:
Neither Chrome 56 nor FF 52 accepts `reflected-xss`.

I'm also getting this in Chrome 56, but not FF 52:

> Failed to set referrer policy: The value 'strict-origin-when-cross-origin' is not one of 'no-referrer', 'no-referrer-when-downgrade', 'origin', 'origin-when-cross-origin', or 'unsafe-url'. The referrer policy has been left unchanged.

EdOverflow:
> While we're here, should we drop `reflected-xss`?
 
Yep, go ahead! We already have `X-XSS-Protection` in place.

whit537:
Okay! ttw tests confirmed working locally. If Travis runs green I think we're good?

EdOverflow:
> Failed to set referrer policy: The value 'strict-origin-when-cross-origin' is not one of 'no-referrer', 'no-referrer-when-downgrade', 'origin', 'origin-when-cross-origin', or 'unsafe-url'. The referrer policy has been left unchanged.

Chrome does not yet support the `strict-` prefix. We could add a `no-referrer` fallback.

EdOverflow:
~~~diff
if 'Referrer-Policy' not in response.headers:
-      response.headers['Referrer-Policy'] = 'strict-origin-when-cross-origin'
+     response.headers['Referrer-Policy'] = 'no-referrer, strict-origin-when-cross-origin'
~~~

whit537:
> We could add a `no-referrer` fallback.

Done in 7d8d4caceb6f3c2106094c3a0c964b2e522c386a.

whit537:
The `no-referrer` setting breaks our fonts, because Typography.com depends on that header for their DRM. We get a 403 w/o it.

EdOverflow:
In that case `no-referrer-when-downgrade`.

whit537:
In that case `no-referrer-when-downgrade`.

Done in d8068ef8702c15babc37126641930c2e1b4f6b85, confirmed working locally in Chrome & FF.

whit537:
I guess it's okay to load up Travis with little commits on this PR in order to really verify #4359. 😄

whit537:
Dang it, still hosed at Travis. Basically 50% failing still with timeouts. :-/

screen shot 2017-03-24 at 1 40 11 pm

whit537:
Alright, well, this at least fixes #4322, so I guess it's worth merging. I've restarted the latest Travis to work around #4359.

whit537:
Travis is Green.

whit537:
Squashed and rebased on master. Was d8068ef8702c15babc37126641930c2e1b4f6b85.

gratipay/security/__init__.py
```diff
@@ -43,7 +43,8 @@ def add_headers_to_response(response):
 
     # https://www.w3.org/TR/referrer-policy/
     if 'Referrer-Policy' not in response.headers:
-        response.headers['Referrer-Policy'] = 'strict-origin-when-cross-origin'
+        response.headers['Referrer-Policy'] = \
+                                      'no-referrer-when-downgrade, strict-origin-when-cross-origin'
```
whit537: Should be okay: http://caniuse.com/#feat=referrer-policy.

whit537:
Ready for review/merge, @EdOverflow @rohitpaulk et al.

PR #4400, opened 2017-04-06
Add rtlo_scrub()

EdOverflow:
[h1:210818](https://hackerone.com/reports/210818)

---

TODO:

- [x] Add test


whit537:
Closing as out of date.

tests/py/test_markdown.py
```diff
@@ -72,15 +72,16 @@ def test_scrubs_tricky_scripts(self):
         # http://stackoverflow.com/q/753052#comment24080274_4869782
         assert markdown.render_and_scrub('<script<script>>alert("Hi!")<</script>/script>') == \
                                                         '&gt;alert(\u201cHi!\u201d)&lt;/script&gt;'
-
```
whit537: Let's keep these newlines.

tests/py/test_markdown.py
```diff
@@ -72,15 +72,16 @@ def test_scrubs_tricky_scripts(self):
...
     def test_scrubs_comments(self):
         # http://stackoverflow.com/a/19730306
         assert markdown.render_and_scrub('<img<!-- --> src=x onerror=alert(1);//><!-- -->') == \
                                               'src=x onerror=alert(1);//&gt;&lt;!\u2013 \u2013&gt;'
-
     def test_renders_entity_references(self):
         assert markdown.render_and_scrub('&trade;') == '\u2122'
 
+    def test_scrubs_rtlo(self):
+        assert markdown.rtlo_scrub('ed.io/about&#8238;3p\u202Em.exe') == 'ed.io/about3pm.exe'
```
whit537: Working in test ... but we don't actually call rtlo_scrub from anywhere in production, do we?

EdOverflow: That is correct. I have just added it in production: https://github.com/gratipay/gratipay.com/pull/4400/commits/a8a169600562e986066371fd0c6b2d7f9d4fcfa8.

gratipay/utils/markdown.py
```diff
@@ -19,6 +19,13 @@ def render(markdown):
     ))
 
 
+def rtlo_scrub(markdown):
```
rohitpaulk: Can't this be done within `render_and_scrub`? Is there a case where we would want to use `render_and_scrub` without this functionality?

EdOverflow: I can, but if we somehow want to support RTLO languages in the future this may actually cause some issues, therefore to make it easy to spot and resolve I have made it into a separate function. Would you still prefer that I place everything into one function?

whit537: We may already have RTL user statements. What is the effect of this on those?

EdOverflow: I just experimented a bit and I see no issue. So my assumption was incorrect. I will make this into one function as @rohitpaulk suggested.

whit537:
I added a little cleanup commit in 7f3a689fe297e8a8087afa11716cc74b4a9f1e39. I'm good here if @rohitpaulk is.

EdOverflow:
Who would like to merge this? 😃

rohitpaulk:
Looks good to me.

PR #4404, opened 2017-04-10
Edit claimed packages

whit537:
Part of #4305, follows on #4398.

#### Todo

- [x] set `homepage` to URL on npmjs.com
- [x] set initial `product_or_service` to npm description
- [x] set initial image as wombat(?)
- [x] retain npm indicator over profile image for claimed packages
- [x] create review ticket, I guess? one ticket for all packages claiming?
- [x] regular project pages should work:
  - [x] project profile
  - [x] ~~distributing~~ didn't check; let's make an 🐴  out of you and me
  - [x] receiving
  - [x] listing on homepage
  - [x] listing in search results
  - [x] listing on participant profile
  - [x] edit—just can't change homepage
  - [x] review/reject
  - [x] close—package goes back to being unclaimed

PR #4405, opened 2017-04-11
Unclaim package (close project)

whit537:
Part of #4305, follows on #4404.


whit537:
Okay, this is ready.

whit537:
Actually, this is small enough that we can just fold it into #4404.

PR #4406, opened 2017-04-12
Some UX cleanups for package claiming

whit537:
Part of #4305, follows on #4422.

- [x] gray out the wombat so it feels more like a default
- [x] add an npm indicator in project listings
- [x] knock out https://github.com/gratipay/gratipay.com/issues/4394 please

PR #4410, opened 2017-04-17
Style the claim form

whit537:
Part of #4305, follows on #4406.

- [x] style the confirmation form so it's not so plain
- [x] highlight/preselect any already-verified email(s) on the confirmation form
  - [x] list emails in an order: primary, verified, linked-but-unverified, unlinked, linked to another account
  - [x] style the list


whit537:
Squashed and rebased on #4406, was 07d7e81ae65bed943f1b3e54ce65be4e5a4ba774.

whit537:
p7qgv8l0gs

whit537:
Rebased, was a3705b138d02207aa2fca33114c66858177572da.

gratipay/models/package/emails.py
```diff
@@ -0,0 +1,69 @@
...
+        for email in participant.get_emails():
+            if email.address == participant.email_address:
+                group = PRIMARY
+            elif email.verified:
+                group = VERIFIED
+            else:
+                group = UNVERIFIED
+            group_by_participant_email[email.address] = group
+
+        other_verified = self.db.all('''
```
dowski: This looks like another one of those reverse-mixins.

js/gratipay/dropdown.js
```diff
@@ -0,0 +1,117 @@
+Gratipay.Dropdown = function(selector) {
+    var $ul = $(selector);
+    var $labels = $('label', $ul);
+
+    // state for vertical position
+    var topFactor = 0;      // float between 0 and $labels.length-1
+    var maxTopFactor = $labels.length - 1;
```
dowski: What if $labels.length is 0?

js/gratipay/dropdown.js
```diff
@@ -0,0 +1,117 @@
...
+        if (scrollThrottle !== null) return;
+        window.setTimeout(function() {
+            var t = topFactor, by = e.deltaY / 128; // divisor arrived at experimentally
+            by = by * e.deltaFactor;
+            t = t - by;
+            if (t < 0) t = 0;
+            if (t > maxTopFactor) t = maxTopFactor;
+            moveTo(t);
+            scrollThrottle = null;
+        }, 12);
```
dowski: Why 12?

gratipay/models/package/emails.py
whit537: I guess let's discuss the general pattern at https://github.com/gratipay/gratipay.com/pull/4422#discussion_r113821151.

js/gratipay/dropdown.js
whit537: I believe that's prevented by the condition in the template at https://github.com/gratipay/gratipay.com/blob/371e17a8e089515cb4234e9b7c361de95df02fa1/www/on/npm/%25package/index.html.spt#L61, no? If there are no emails then there is no `.gratipay-dropdown` rendered and therefore we never get here.

js/gratipay/dropdown.js
whit537: Commented in 47ad781406bbc5d1bec5e12f5fc8bcb08a950177. S'okay?

js/gratipay/dropdown.js
dowski: Right!

PR #4412, opened 2017-04-19
Use hmac.compare_digest()

EdOverflow:


whit537:
Need to import hmac, looks like? And where is `force_bytes` from?

EdOverflow:
Whoops, I went a bit too fast there.

whit537:
[Interesting](https://hg.python.org/cpython/file/2.7/Lib/hmac.py#l8):

```python
from operator import _compare_digest as compare_digest
```

¯\\\_(ツ)\_/¯

PR #4422, opened 2017-04-26
Factor out a package.team mixin

whit537:
Part of #4305, follows on #4404.

Minor refactor. Ready for review.

### DIFF
--- a/gratipay/models/package/team.py
+++ b/gratipay/models/package/team.py
@@ -1,0 +1,79 @@
+# -*- coding: utf-8 -*-
+from __future__ import absolute_import, division, print_function, unicode_literals
+
+import uuid
+
+from gratipay.models.team import Team as _Team
+
+
+class Team(object):
+    """A :py:class:`~gratipay.models.package.Package` can have a
+    :py:class:`~gratipay.models.team.Team` associated with it.
+    """
+
+    @property
+    def team(self):
+        """A computed attribute, the :py:class:`~gratipay.models.team.Team`
+        linked to this package if there is one, otherwise ``None``. Makes a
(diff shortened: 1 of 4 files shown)


gratipay/models/package/team.py
```diff
@@ -0,0 +1,79 @@
...
+
+
+    def load_team(self, cursor):
+        """Given a database cursor, return a
+        :py:class:`~gratipay.models.team.Team` if there is one linked to this
+        package, or ``None`` if not.
+        """
+        return cursor.one( 'SELECT t.*::teams FROM teams t WHERE t.id='
+                           '(SELECT team_id FROM teams_to_packages tp WHERE tp.package_id=%s)'
+                         , (self.id,)
```
dowski: Might be my recent lack of exposure to Python mixin classes, but this actually feels a bit backward. You mix it in with Package, but instead of Package leveraging some of the functionality of the mixed-in Team, Team leverages things like `self.id` that must be in Package.

whit537: It's consumers that leverage the synergies contained herein, not necessarily the main Package object itself. The goal here is to keep files reasonably sized (well under 1,000 lines ideally), and the architecture is to have a main Package class (e.g.) that is essentially a container, with API provided by these mixins. But the API itself is called primarily by other objects.

Does that make any sense?

dowski: Yeah, that makes sense. It seems like the API of Package is a little more opaque that way. It more hides the complexity due to lines of code (if any) than reduces it. But it's no reason to hold up this PR - just my thoughts on the matter now. ;-)

Issue #4415, opened 2017-04-21
Robustify custom select


whit537:
I made a custom select widget under https://github.com/gratipay/gratipay.com/pull/4305. There's work yet to be done on it to make it behave like a proper select.

- [ ] scroll widget into view when activated at edge of viewport 
- [ ] arrow keys
- [ ] type to find
- [ ] scroll in touch - // http://stackoverflow.com/a/10803068
- [ ] ensure responsivity
- [ ] cross-browserize
- [ ] check on a phone, cf. https://github.com/gratipay/gratipay.com/pull/4432
- [ ] bounce physics at end of scroll?
- [ ] prevent text selection?
- [ ] fix off-by-one bug on large lists: https://github.com/gratipay/gratipay.com/pull/4411#issuecomment-296303534

PR #4432, opened 2017-04-28
Fix on a phone

whit537:
At the very least, text overflows poorly at low width.

screen shot 2017-04-26 at 8 34 15 pm

Check on an actual phone, though, to be sure it works well enough with touch.


whit537:
Okay! I've managed to use [ngrok](https://ngrok.com/) to hit localhost from a phone. The above width bug is an issue, yes, but there are even worse problems:

- I can't close the select by clicking outside of the select.
- I can't scroll just the select, only the entire page. This may not be a problem.

whit537:
The scrolling is *almost* okay. It clips the top at the top of the page, so if the select overflows the viewport you can't get to it.

Maybe we just need to do something to capture touch scroll events? Wasn't there something like that?

whit537:
I'm looking at:

- https://api.jquery.com/scroll/
- http://api.jquerymobile.com/scrollstart/

whit537:
If we're going to add another js dep then it's probably time to bundle vendor scripts into one to avoid multiple additional http requests.

whit537:
We've already got [the pattern](https://github.com/gratipay/gratipay.com/blob/2071/www/assets/gratipay.js.spt), just need to apply something similar to [these scripts](https://github.com/gratipay/gratipay.com/blob/2071/templates/base.html#L134-L137).

whit537:
https://coderwall.com/p/bdxjzg/tap-vs-click-death-by-ignorance

whit537:
That post says it was updated a few months ago, but [this post](https://developers.google.com/web/updates/2013/12/300ms-tap-delay-gone-away) suggests that the 300 ms click/tap delay is actually a thing of the past.

whit537:
I did double-check that we have `width=device-width` [set](https://github.com/gratipay/gratipay.com/blob/2071/templates/base.html#L6).

whit537:
[Apparently](http://stackoverflow.com/questions/3705937/document-click-not-working-correctly-on-iphone-jquery) `cursor: pointer` can render something clickable in mobile browsers but we don't want that off the select, only on the select.

whit537:
Is this out of date? I think we'd still be okay to use `vclick` for turning off the select. Maybe that moves the content underneath if it changes the size of the document (a long select will increase the size of the document)?

> Use `vclick` with caution on touch devices. Webkit based browsers synthesize `mousedown`, `mouseup`, and `click` events roughly 300ms after the `touchend` event is dispatched. The target of the synthesized mouse events are calculated at the time they are dispatched and are based on the location of the touch events and, in some cases, the implementation specific heuristics which leads to different target calculations on different devices and even different OS versions for the same device. This means the target element within the original touch events could be different from the target element within the synthesized mouse events.
>
> We recommend using `click` instead of `vclick` anytime the action being triggered has the possibility of changing the content underneath the point that was touched on screen. This includes page transitions and other behaviors such as collapse/expand that could result in the screen shifting or content being completely replaced.

whit537:
Alright, leaving open for now and bumping from #4305. Width bug fixed in #4435, that's enough for over there.

whit537:
(Part of the issue is that I don't have a phone to test on anymore this afternoon.)

whit537:
Closing in favor of #4415. Start a new PR now that this is off of #4305.

PR #4433, opened 2017-04-28
Make fake packages

whit537:
Can be merged after #4305 is in.


whit537:
Looks like the build failure is because of a package name collision. We should weed out duplicates there, I guess.

whit537:
Closing in favor of #4436.

Issue #4427, opened 2017-04-27
✈️  Give to package.json


whit537:
-------------------------------

✈️  **This is the flight deck for the [Give to package.json](https://github.com/orgs/gratipay/projects/14?fullscreen=true) project.** ✈️ 

-------------------------------

🗻  **That project is part of the [Make it easier for companies to fund open source](https://github.com/gratipay/inside.gratipay.com/issues/987) epic.** 🗻 

-------------------------------

The definition of done is when we can say "Give us $10,000 and a package.json, and we'll distribute it for you."

### Todo

- [x] make room on the `/on/npm/foo/` page for the claiming workflow—move the name to the banner h1, and replace the wombat with the description: #4304
- [x] new email subflooring: ~~#4343~~ ~~#4344~~ ~~#4345~~ ~~#4365~~ ~~#4350~~ ~~#4355~~ ~~#4346~~ ~~#4347~~ ~~#4348~~ ~~#4349~~ ~~#4383~~
- [x] pull emails from maintainers only, not author (https://github.com/gratipay/gratipay.com/pull/4305#issuecomment-278429627): https://github.com/gratipay/gratipay.com/pull/4438
  - [x] also dedupe email addresses: https://github.com/gratipay/gratipay.com/pull/4438
- [x] set up npm syncing—besides just needing to stay in sync, people are going to want to ~~update the email address on file in their package.json~~ ~~release a new version~~ *take some action at npm* and then "immediately" see the new email show up to continue the flow: https://github.com/gratipay/gratipay.com/pull/4438
  - [x] add sync lag to [health dashboard](http://inside.gratipay.com/appendices/health)
- [x] Implement claim/unclaim/reclaim packages: #4305.
- [x] bulk add from `/on/npm/`: #4488
- [x] discovery via package.json: #4507
- [ ] fix bugs
- [x] write blog post justifying amount recommendation
- [x] "Pay for Open Source" on the homepage: #4598
- [ ] update payment processing howtos: https://github.com/gratipay/inside.gratipay.com/pull/1154
- [ ] write howto for managing mailing list of company payers
- [ ] Takeoff on the [project board](https://github.com/orgs/gratipay/projects/14?fullscreen=true)

-------------------------------

✈️  **This is the flight deck for the [Give to package.json](https://github.com/orgs/gratipay/projects/14?fullscreen=true) project.** ✈️ 

-------------------------------

whit537:
[...]

PR #4438, opened 2017-05-03
Rebuild npm syncing

whit537:
Picks up from #4148 and https://github.com/gratipay/gratipay.com/issues/4427#issuecomment-298855735. Part of #4427.

We managed to load up a snapshot of npm back on #4148, but only barely enough to show stubby pages. Now that #4305 is pretty much ready to go, it's time to get some more robust npm syncing in place. Turns out the old API we were depending on is [going away](http://blog.npmjs.org/post/157615772423/deprecating-the-all-registry-endpoint), so we need to rebuild this subsystem around [npm's](https://github.com/npm/registry/blob/master/docs/follower.md) CouchDB [change stream](http://docs.couchdb.org/en/2.0.0/api/database/changes.html).

### Specs

- don't store author email address, only maintainer (https://github.com/gratipay/gratipay.com/pull/4305#issuecomment-278458790)
- deduplicate emails (shouldn't be a problem w/ only maintainers?)

### Todo

- [x] clear out old code
- [x] write new code
- [x] what about deletes?
- [x] remove `ijson` dep, add `couchdb`
- [x] moar tests for consume!
- [x] log distance behind registry?
- [x] fix on travis
- [x] check behavior when we catch up


whit537:
I'm pretty sure the initial load that took two days included README fetching, which we ripped out in #4211. I believe now we can load up from just the metadata, which should go much quicker.

whit537:
Curveball!

"[Deprecating the /-/all registry endpoint](http://blog.npmjs.org/post/157615772423/deprecating-the-all-registry-endpoint)"

whit537:
> If you are using the endpoint as a way to get a list of all packages, we encourage you to write a [registry follower](http://t.umblr.com/redirect?z=https%3A%2F%2Fgithub.com%2Fnpm%2Fregistry%2Fblob%2Fmaster%2Fdocs%2Ffollower.md&t=MDU0ZWJiODk1OTI5NmVhZmUzNWZjZTliNGJhNWQzNDM3N2ZkY2M1ZixLOTVXUDNmTw%3D%3D&b=t%3AnXsLs1P4AptPf1fBr_nFxw&p=http%3A%2F%2Fblog.npmjs.org%2Fpost%2F157615772423%2Fdeprecating-the-all-registry-endpoint&m=1) that watches the changes stream at replicate.npmjs.com for public packages. We provide sample code and libraries to support you.

whit537:
Pretty sure we want to use https://github.com/djc/couchdb-python.

whit537:
Probably this API?

https://pythonhosted.org/CouchDB/client.html#couchdb.client.Database.changes

whit537:
Underlying API: http://docs.couchdb.org/en/2.0.0/api/database/changes.html.

whit537:
Yeah, this is gonna be it. :-)

```python
#!/usr/bin/env python
# -*- coding: utf-8 -*-

from couchdb import Database


npm = Database('https://skimdb.npmjs.com/registry')
changes = npm.changes(limit=10)
import pdb; pdb.set_trace()
```

screen shot 2017-05-03 at 6 41 29 am

whit537:
https://blog.andyet.com/2015/04/06/postgres-pubsub-with-json/

whit537:
Alright, I think to start with we should go for a naive approach where we have a single process/thread that consumes the registry stream and inserts/updates in our database all at once. Decoupling fetch and update will be more complicated and should be done because we're getting too far behind otherwise. We could even probably build a quick dashboard to show far behind we are—or log over to Librato.

whit537:
Shelving:

```python
#!/usr/bin/env python
# -*- coding: utf-8 -*-

from couchdb import Database
from gratipay import wireup


def go(db):
    npm = Database('https://skimdb.npmjs.com/registry')
    changes = npm.changes(feed='continuous', include_docs=True)
    for change in changes:
        doc = change['doc']
        if 'name' not in doc:
            continue  # not a package, probably a design doc*
        name = doc['name']
        description = doc.get('description', '')
        emails = [e for e in [m.get('email') for m in doc.get('maintainers', [])] if e]

        try:
            db.run( "update packages set description=%s, emails=%s where package_manager='npm' and name=%s"
                  , (description, emails, name)
                   )
        except:
            db.run('insert into packages () values ()')





# * https://github.com/npm/registry/blob/aef8a275/docs/follower.md#clean-up
if __name__ == '__main__':
    env = wireup.env()
    db = wireup.db(env)
    go(db)
```

whit537:
This is basically a rewrite of this subsystem.

whit537:
This will actually be much better though. No more `ijson` dependency and also much closer to real-time. No more batch mode.

whit537:
Eep! Time to upgrade to Postgres 9.6 locally. 😊

```
psycopg2.ProgrammingError: syntax error at or near "ON"
LINE 6:     ON CONFLICT (package_manager, name) UPDATE
```

whit537:
Yesssss!!!

screen shot 2017-05-03 at 2 57 50 pm

whit537:
Okay! I'm going to run this locally and see how long it takes and how it behaves.

whit537:
Added some logging.

screen shot 2017-05-03 at 4 59 04 pm

whit537:
Started a long run ...

whit537:
```
pid-77299 thread-140735204086528 (MainThread) Picking up with npm sync at -1.
```

whit537:
I'm at about 80,000 after about 10(?) minutes.

whit537:
So maybe an hour for the whole thing?

whit537:
That's way better than two days, anyway. ☺️

whit537:
> deleted documents include the `"deleted": true` attribute

http://docs.couchdb.org/en/2.0.0/api/database/changes.html#polling

whit537:
Out of time for now.

```
pid-77350 thread-140735204086528 (MainThread) KeyboardInterrupt
pid-77350 thread-140735204086528 (MainThread) Encountered an error, will pick up with 517201 in 60 seconds (Ctrl-C to exit) ...
^C
real    17m47.575s
user    1m16.416s
sys     0m25.121s
[gratipay] $
```

```
Every 1.0s: echo 'select count(*) from packages' | psql gratipay                 Wed May  3 17:27:49 2017

Null display is "¤".
Line style is unicode.
Border style is 2.
┌────────┐
│ count  │
├────────┤
│ 100404 │
└────────┘
(1 row)
```

whit537:
Deletes should remove locally. Need to take care to unlink teams from packages when deleting packages. It's okay for the team to stick around, I think? It'll have a 404 homepage on npmjs.

whit537:
Chasing which value to use for logging trailing status.

https://skimdb.npmjs.com/registry

`update_seq`? `committed_update_seq`?

https://github.com/couchbase/sync_gateway/issues/1166#issuecomment-148770191 following http://docs.couchdb.org/en/2.0.0/json-structure.html define these as:

> The current number of updates made in the database.
> The number of committed updates.

Not very helpful. :-/

Better:

> So the difference between `update_seq` and `committed_update_seq` is delayed commits which is not a feature we support, its not any more useful for replication progress than our `update_seq` and theres no reason for us to implement it, so closing out

https://github.com/pouchdb/pouchdb/issues/4116#issuecomment-222127298

I guess we go with `update_seq` and see what we observe?

whit537:
I've asked on irc.freenode.net#npm about the difference between:

https://replicate.npmjs.com/
https://skimdb.npmjs.com/registry

whit537:
We _do_ get a doc for deletes.

```
(Pdb) pp change
{u'changes': [{u'rev': u'2-997ac5f43938c18e61b537c648a819ea'}],
 u'deleted': True,
 u'doc': {u'_deleted': True,
          u'_id': u'netlify-yo-styleguide',
          u'_rev': u'2-997ac5f43938c18e61b537c648a819ea'},
 u'id': u'netlify-yo-styleguide',
 u'seq': 45291}
(Pdb)
```

whit537:
Ah! But there's no name.

whit537:
Alright, so I guess we try deleting the thing and pass silently if it's not there to begin with?

whit537:
We don't have package linking yet. Handle that on #4305 once this is in master.

whit537:
Alright, now what's Travis on about?

whit537:
Errors in `test_team_takes.py` (at least), though that passes for me in isolation locally.

```
[gratipay] $ envtest tests/py/test_team_takes.py 
[...]
======================================= 38 passed in 9.55 seconds =======================================
```

whit537:
Hmm, I do get some errors locally but they look different.

```
  File "/Users/whit537/personal/gratipay/gratipay.com/env/lib/python2.7/site-packages/aspen/website.py", line 62, in respond
    , _return_after=return_after
  File "/Users/whit537/personal/gratipay/gratipay.com/env/lib/python2.7/site-packages/algorithm.py", line 288, in run
    new_state = function(**deps.as_kwargs)
  File "/Users/whit537/personal/gratipay/gratipay.com/env/lib/python2.7/site-packages/aspen/algorithms/website.py", line 115, in get_response_for_resource
    return {'response': resource.respond(state)}
  File "/Users/whit537/personal/gratipay/gratipay.com/env/lib/python2.7/site-packages/aspen/http/resource.py", line 59, in respond
    content_type, body = super(Dynamic, self).respond(accept, state)
  File "/Users/whit537/personal/gratipay/gratipay.com/env/lib/python2.7/site-packages/aspen/simplates/__init__.py", line 169, in respond
    exec(self.pages[1], context)
  File "/Users/whit537/personal/gratipay/gratipay.com/www/search.spt", line 48, in <module>
    ) _""", locals())
  File "/Users/whit537/personal/gratipay/gratipay.com/env/lib/python2.7/site-packages/postgres/__init__.py", line 548, in all
    return cursor.all(sql, parameters)
  File "/Users/whit537/personal/gratipay/gratipay.com/env/lib/python2.7/site-packages/postgres/cursors.py", line 145, in all
    self.execute(sql, parameters)
  File "/Users/whit537/personal/gratipay/gratipay.com/env/lib/python2.7/site-packages/psycopg2/extras.py", line 288, in execute
    return super(NamedTupleCursor, self).execute(query, vars)
ProgrammingError: function similarity(text, unknown) does not exist
LINE 3:                  SELECT p.*::participants p, similarity(user...
                                                     ^
HINT:  No function matches the given name and argument types. You might need to add explicit type casts.

========================== 9 failed, 831 passed, 34 xfailed in 156.47 seconds ===========================
[gratipay] $
```

whit537:
Is there some caching in play at Travis?

whit537:
Can we whittle down to a single test case that fails at Travis?

whit537:
Let's see what Travis does with cfbac72beb86ab78eef2b9758274f55a3ed1fc43 ...

whit537:
Okay! Still failing, and with the same error. 👍

Time to de🐛!

whit537:
9.6.1: https://travis-ci.org/gratipay/gratipay.com/builds/228820499#L1257

whit537:
Making another run at this. I'm going to investigate the nine local failures I'm seeing and get those squared. There's a chance that'll resolve Travis as well, and even if not it'll give me a clean baseline to diff against.

whit537:
Search works on production:

https://gratipay.com/search?q=test

Do I need some contrib module locally?

whit537:
`pg_trgm`

whit537:
Confirmed installed in production:

```
gratipay::DATABASE=> select * from pg_extension;
┌────────────────────┬──────────┬──────────────┬────────────────┬────────────┬───────────┬──────────────┐
│      extname       │ extowner │ extnamespace │ extrelocatable │ extversion │ extconfig │ extcondition │
├────────────────────┼──────────┼──────────────┼────────────────┼────────────┼───────────┼──────────────┤
│ plpgsql            │       10 │           11 │ f              │ 1.0        │ ¤         │ ¤            │
│ pg_stat_statements │       10 │        16407 │ t              │ 1.4        │ ¤         │ ¤            │
│ pg_trgm            │       10 │        16407 │ t              │ 1.3        │ ¤         │ ¤            │
└────────────────────┴──────────┴──────────────┴────────────────┴────────────┴───────────┴──────────────┘
(3 rows)

gratipay::DATABASE=>
```

https://www.postgresql.org/docs/9.6/static/catalog-pg-extension.html

whit537:
Extensions installed locally and tests passing. Okay!

whit537:
Let's get a clean Travis run ...

whit537:
Same 89 failures.

whit537:
Seems like there are two basic failures.

```
E       ProgrammingError: table row type and query-specified row type do not match
E       DETAIL:  Query provides a value for a dropped column at ordinal position 9.
E       CONTEXT:  SQL statement "UPDATE participants
E                          SET balance = balance + NEW.amount
E                        WHERE username = NEW.participant
E                          AND NEW.direction = 'to-participant'"
E       PL/pgSQL function process_payment() line 3 at SQL statement
```

and

```
E       IntegrityError: duplicate key value violates unique constraint "teams_pkey"
E       DETAIL:  Key (slug)=(TheEnterprise) already exists.
```

I was poking at the first earlier, seems tricky. Maybe the second is more transparent?

whit537:
Single test passes!

whit537:
Single file passes!

rohitpaulk:
```
E       ProgrammingError: table row type and query-specified row type do not match
E       DETAIL:  Query provides a value for a dropped column at ordinal position 9.
E       CONTEXT:  SQL statement "UPDATE participants
E                          SET balance = balance + NEW.amount
E                        WHERE username = NEW.participant
E                          AND NEW.direction = 'to-participant'"
E       PL/pgSQL function process_payment() line 3 at SQL statement
```

Quick google search reveals: 

https://dba.stackexchange.com/questions/153981/postgresql-9-6-column-dropping-and-side-effects-on-sql-functions-with-ctes

rohitpaulk:
I'm going to see if we can specify `9.6.2` on travis

rohitpaulk:
:( still failing

rohitpaulk:
`PostgreSQL 9.6.2 is not installed!`

whit537:
@rohitpaulk Can we get away with 9.5 on Travis? I don't think we are depending on any 9.6 features, that's just what Heroku installed as the latest. UPSERT was introduced in 9.5.

rohitpaulk:
> Can we get away with 9.5 on Travis?

Worth trying. Pushing..

rohitpaulk:
I'll try to install 9.6.2 on a separate PR, better if we run CI against the exact version on production.

rohitpaulk:
@whit537 Installing 9.5 worked!

whit537:
> I'll try to install 9.6.2 on a separate PR, better if we run CI against the exact version on production.

Okay. We should also ping Travis (IRC? email?) about upgrading in their infrastructure. Much easier for us that way.

whit537:
Metrics logging seems to be working. A couple more things to clean up and then ready for review ...

whit537:
I'm running the whole thing end-to-end locally to see what kind of behavior we can expect and flush out any last bugs.

whit537:
screen shot 2017-05-05 at 7 38 27 am

whit537:
Yeah, I'm good with this. Submitting for review, @rohitpaulk!

whit537:
Fixing Travis failure ...

gratipay/sync_npm.py
```diff
@@ -0,0 +1,84 @@
...
+            processed = process_doc(doc)
+            if not processed:
+                continue
+            cursor = connection.cursor()
+            op(cursor, processed)
+            cursor.run('UPDATE worker_coordination SET npm_last_seq=%(seq)s', change)
+            connection.commit()
+
+
+def delete(cursor, processed):
```
rohitpaulk: I think it'd be clearer if we renamed `processed` to `processed_doc`

gratipay/sync_npm.py
```diff
@@ -0,0 +1,84 @@
...
+    database, and write as we read.
+
+    """
+    last_seq = get_last_seq(db)
+    log("Picking up with npm sync at {}.".format(last_seq))
+    with db.get_connection() as connection:
+        for change in change_stream(last_seq):
+            if change.get('deleted'):
+                # Hack to work around conflation of design docs and packages in updates
+                op, doc = delete, {'name': change['id']}
```
rohitpaulk: I think this is a bit confusing. `delete` takes a dictionary, although it only needs one `string` as the argument. Also, we don't need to pass the fake doc (`{'name': change['id']}`) through `process_doc`. 

At the cost of a line or two more, I think this can be simplified.

rohitpaulk: Something along the lines of: 

screen shot 2017-05-05 at 5 37 48 pm

Raw version: 

Before: 

```python
with db.get_connection() as connection:
        for change in change_stream(last_seq):
            if change.get('deleted'):
                # Hack to work around conflation of design docs and packages in updates
                op, doc = delete, {'name': change['id']}
            else:
                op, doc = upsert, change['doc']
            processed = process_doc(doc)
            if not processed:
                continue
            cursor = connection.cursor()
            op(cursor, processed)
            cursor.run('UPDATE worker_coordination SET npm_last_seq=%(seq)s', change)
            connection.commit()


def delete(cursor, processed):
    cursor.run("DELETE FROM packages WHERE package_manager='npm' AND name=%(name)s", processed)


def upsert(cursor, processed):
    cursor.run('''
    INSERT INTO packages
                (package_manager, name, description, emails)
         VALUES ('npm', %(name)s, %(description)s, %(emails)s)
    ON CONFLICT (package_manager, name) DO UPDATE
            SET description=%(description)s, emails=%(emails)s
    ''', processed)
```

After: 

```python
with db.get_connection() as connection:
        for change in change_stream(last_seq):
            cursor = connection.cursor()
            if change.get('deleted'):
                # Hack to work around conflation of design docs and packages in updates
                delete(cursor, change['id'])
            else:
                upsert(cursor, process_doc(doc))
            cursor.run('UPDATE worker_coordination SET npm_last_seq=%(seq)s', change)
            connection.commit()


def delete(cursor, package_name):
    cursor.run("DELETE FROM packages WHERE package_manager='npm' AND name=%s", package_name)


def upsert(cursor, processed_doc):
    cursor.run('''
    INSERT INTO packages
                (package_manager, name, description, emails)
         VALUES ('npm', %(name)s, %(description)s, %(emails)s)
    ON CONFLICT (package_manager, name) DO UPDATE
            SET description=%(description)s, emails=%(emails)s
    ''', processed_doc)
```

Only downside I see here is that we're doing a little bit more work (calling `process_doc`, checking the `deleted` key) inside the transaction

gratipay/cli/sync_npm.py
```diff
@@ -3,40 +3,35 @@
...
     env = wireup.env()
-    args = parse_args(argv[1:])
     db = wireup.db(env)
-
-    subcommands[args.command](env, args, db)
+    while 1:
+        with sentry.teller(env):
+            consume_change_stream(production_change_stream, db)
+        try:
+            last_seq = get_last_seq(db)
```
rohitpaulk: Hmm, if we're calling `get_last_seq` here anyway - might make sense to simplify the function definition of `consume_change_stream` to accept the stream directly, and not a function that has to be called with `seq` to return the stream?

gratipay/utils/sentry.py
```diff
@@ -0,0 +1,31 @@
...
+            sys.stdout = sys.stderr  # work around aspen.log_dammit limitation; sigh
+            self.tell_sentry = wireup.make_sentry_teller(env, fallback)
+        finally:
+            sys.stdout = sys.__stdout__
+
+    def __enter__(self):
+        return self
+
+    def __exit__(self, exc_type, exc_value, traceback):
+        self.tell_sentry(exc_type, {})
```
rohitpaulk: Does this have all the details needed? Shouldn't we pass `traceback` and `exc_value`? (I'm not sure what they are, but `traceback` sure sounds important)

gratipay/utils/sentry.py
```diff
@@ -0,0 +1,31 @@
+# -*- coding: utf-8 -*-
+from __future__ import absolute_import, division, print_function, unicode_literals
+
+import sys
+import traceback
+
+from aspen import log
+from gratipay import wireup
```
rohitpaulk: `utils` importing `wireup`? 😛 That seems hacky. No neater way?

gratipay/sync_npm.py
whit537: Done in 631dfc9b7c8bd9b0c8851845e073a792c6d53810.

gratipay/sync_npm.py
whit537: That doesn't account for skipping docs with no `name` key. How about 631dfc9b7c8bd9b0c8851845e073a792c6d53810?

rohitpaulk: Ah, yes 631dfc9b7c8bd9b0c8851845e073a792c6d53810 looks good

gratipay/utils/sentry.py
whit537: Sentry accesses Python's global exception state directly during [`captureException`](https://github.com/gratipay/gratipay.com/blob/2072/gratipay/wireup.py#L163) (via `sys.exc_info`, presumably), so we don't have to pass it through these function calls.

rohitpaulk: Interesting

gratipay/utils/sentry.py
whit537: > No neater way?

Not on this PR. 😞 

I agree it's hacky. Eventually I would see rewiring Sentry along the lines of what was started in #4345 and other PRs listed under" new email subflooring" on #4427.

gratipay/cli/sync_npm.py
whit537: Done in af41409c5bea0f3cea3afbd804e64c83ef97c693.

gratipay/utils/sentry.py
whit537: (This PR is already 1.5x our 400-net-lines rule of thumb.)

rohitpaulk:
I'm good once travis is :)

whit537:
Travis is good! :-D

rohitpaulk:
`==`

whit537:
Some discussion of error reporting architecture in [slack](https://gratipay.slackarchive.io/gratipay/page-95/ts-1493988357169577).

rohitpaulk:
Merging and deploying...

whit537:
When you're done you could try adding an instrument to http://inside.gratipay.com/appendices/health for npm sync lag. Librato is in 1Password so it would be a good test for that as well. :)

rohitpaulk:
Okay, we've got an error..

screen shot 2017-05-05 at 7 02 28 pm

rohitpaulk:
Hmm, I had a `sync_npm` older lying around.. wonder where that came from

rohitpaulk:
I forgot to add the env var 😞 Gratipay was down for around 3 minutes, back up now

rohitpaulk:
Now to figure out how to run the syncer. Add it to the heroku procfile?

rohitpaulk:
I'm going to try to run as a one-off dyno first

whit537:
[Some deploy log in slack](https://gratipay.slackarchive.io/gratipay/page-96/ts-1493994782604712).

whit537:
> Hmm, I had a `sync_npm` folder lying around.. wonder where that came from

Maybe pyc files kept Git from removing it after the switch from `sync_npm/` to `sync_npm.py`?

PR #4444, opened 2017-05-05
Try to install pg 9.6.2 manually

rohitpaulk:


whit537:
The point is to work around a in 9.6.1 that we hit on https://github.com/gratipay/gratipay.com/pull/4438. We worked around it be dropping back to 9.5, but we're on 9.6 in production so better to have the same version in CI.

rohitpaulk:
https://github.com/gratipay/gratipay.com/issues/4445#issuecomment-300762419

whit537:
Nice!

Fwiw, I'd prefer we observe no self-merging unless urgent, to avoid a slippery slope.

rohitpaulk:
> Fwiw, I'd prefer we observe no self-merging unless urgent, to avoid a slippery slope.

👍 I saw this as less critical, since it only touched build scripts. Agreed though, will avoid self-merging as far as possible.

PR #4454, opened 2017-05-09
Fix regression in envvar checks

### LINKED ISSUE #4446: Add a check in deploy.sh for environment variables
When deploying #4438, I forgot to add the `CHECK_NPM_SYNC_EVERY` env var, and gratipay.com was down for around 3 minutes until I figured out what was wrong and fix it.

We should be able to detect this before deploying by adding a check to `deploy.sh`

whit537:
Closes #4446.

### DIFF
--- a/gratipay/wireup.py
+++ b/gratipay/wireup.py
@@ -441,6 +441,6 @@ def env():
     return env
 
 
-def __main__():
+if __name__ == '__main__':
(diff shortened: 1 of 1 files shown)


whit537:
Results demonstrated at https://github.com/gratipay/gratipay.com/issues/4446#issuecomment-300320924.

PR #4464, opened 2017-05-12
Ignore deleted events for inexistent packages

### LINKED ISSUE #4463: NPM sync is broken
https://gratipay.slack.com/archives/C36LJJF9V/p1494580201702422

rohitpaulk:
Fixes #4463

### DIFF
--- a/tests/py/test_sync_npm.py
+++ b/tests/py/test_sync_npm.py
@@ -70,6 +70,9 @@ class ConsumeChangeStreamTests(Harness):
         sync_npm.consume_change_stream(self.change_stream(docs), self.db)
         assert self.db.one('select * from packages') is None
 
+    def test_delete_tolerates_inexistent_packages(self):
+        docs = [{'deleted': True, 'id': 'foo'}]
+        sync_npm.consume_change_stream(self.change_stream(docs), self.db)
 
     def test_even_deletes_package_with_linked_team(self):
 
--- a/tests/py/test_sync_npm.py
+++ b/tests/py/test_sync_npm.py
@@ -35,14 +35,9 @@ class ProcessDocTests(Harness):
 class ConsumeChangeStreamTests(Harness):
 
     def change_stream(self, changes):
-        def change_stream(seq):
-            for i, change in enumerate(changes):
-                if i < seq: continue
-                change['seq'] = i
-                yield change
-        last_seq = sync_npm.get_last_seq(self.db)
-        return change_stream(last_seq)
-
+        for i, change in enumerate(changes):
+            change['seq'] = i
+            yield change
 
     def test_packages_starts_empty(self):
         assert self.db.all('select * from packages') == []
@@ -87,21 +82,6 @@ class ConsumeChangeStreamTests(Harness):
         docs = [{'deleted': True, 'id': 'foo'}]
         sync_npm.consume_change_stream(self.change_stream(docs), self.db)
         assert self.db.one('select * from packages') is None
-
-
-    def test_picks_up_with_last_seq(self):
-        docs = [ {'doc': {'name': 'foo', 'description': 'Foo.'}}
-               , {'doc': {'name': 'foo', 'description': 'See alice?',
-                  'maintainers': [{'email': 'alice'}]}}
-               , {'doc': {'name': 'foo', 'description': "No, I don't see alice!"}}
-                ]
-        self.db.run('update worker_coordination set npm_last_seq=2')
-        sync_npm.consume_change_stream(self.change_stream(docs), self.db)
-
-        package = self.db.one('select * from packages')
-        assert package.description == "No, I don't see alice!"
-        assert package.emails == []
-
 
     def test_sets_last_seq(self):
         docs = [{'doc': {'name': 'foo', 'description': 'Foo.'}}] * 13
--- a/gratipay/sync_npm.py
+++ b/gratipay/sync_npm.py
@@ -53,8 +53,9 @@ def consume_change_stream(stream, db):
             if change.get('deleted'):
                 package = Package.from_names(NPM, change['id'])
                 if not package:
-                    # We've received 'deleted' events on production that didn't
-                    # have a corresponding package in our database.
+                    # As a result of CouchDB's compaction algorithm, we might
+                    # receive 'deleted' events for docs even if we haven't seen
+                    # the corresponding events for when the doc was created
                     continue
                 op, kw = package.delete, {}
             else:
--- a/gratipay/sync_npm.py
+++ b/gratipay/sync_npm.py
@@ -52,7 +52,10 @@ def consume_change_stream(stream, db):
             # Decide what to do.
             if change.get('deleted'):
                 package = Package.from_names(NPM, change['id'])
-                assert package is not None  # right?
+                if not package:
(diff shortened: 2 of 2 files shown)


tests/py/test_sync_npm.py
```diff
@@ -85,21 +83,6 @@ def test_even_deletes_package_with_linked_team(self):
         sync_npm.consume_change_stream(self.change_stream(docs), self.db)
         assert self.db.one('select * from packages') is None
 
-
-    def test_picks_up_with_last_seq(self):
```
whit537: Is this covered by other tests in a way I'm not seeing?

clone1018:
Do we understand why a package was marked as deleted that we didn't have in our index?

whit537:
From [sentry](https://sentry.io/gratipay/gratipay-com/issues/271411814/) it looks like `adam-test` is the package in question. Could recrawl the changes stream from local to investigate.

tests/py/test_sync_npm.py
rohitpaulk: This test currently only tested our test function (`change_stream(self, changes)`). The job of picking up from `last_seq` now belongs to `production_change_stream`, which internally calls the CouchDB library, and hence we don't need to test it.

rohitpaulk:
> Could recrawl the changes stream from local to investigate.

Yes, we could do that. However, we haven't seen an error like this across 4 million records - is it really worth the effort? :)

Middle ground would be to merge this to get npm sync back up and running, and maybe log this error out and count logs a week later to decide whether we should investigate.

whit537:
We could also manually insert an `adam-test` package to then be deleted. We will need to learn whether this is something we're going to hit regularly or a one-off fluke.

The risk is that we'll be masking a bug by blanket-ignoring these discrepancies.

rohitpaulk:
> We could also manually insert an adam-test package to then be deleted.

Makes sense, I'm happy to go ahead with this

rohitpaulk:
I found an easier way to validate this - I just hit the API directly and filtered by `doc_id`. There indeed is only one doc for `adam-test`

screen shot 2017-05-13 at 7 07 48 am

Looks like a CouchDB bug? 😃

rohitpaulk:
http://docs.couchdb.org/en/2.0.0/api/database/changes.html doesn't say anything specific about this..

rohitpaulk:
My JIRA skills have FINALLY come to use outside of companies I work for! https://issues.apache.org/jira/browse/COUCHDB-3418

whit537:
Good sleuthing!

> Makes sense, I'm happy to go ahead with this

+1

rohitpaulk:
Fake package inserted. 

```
→ heroku pg:psql -a gratipay
--> Connecting to postgresql-octagonal-73752
psql (9.6.2)
SSL connection (protocol: TLSv1.2, cipher: ECDHE-RSA-AES256-GCM-SHA384, bits: 256, compression: off)
Type "help" for help.

gratipay::DATABASE=> INSERT INTO packages (package_manager, name, description, emails) VALUES ('npm', 'adam-test', 'Dummy description', '{}');
INSERT 0 1
```

rohitpaulk:
Hmm, it momentarily started to process a few docs - and then encountered an exception again. 

screen shot 2017-05-13 at 4 32 13 pm

Looking for the exception in Sentry...

rohitpaulk:
The initial change we got stuck on was sequence `1908160`, now we're stuck on `1908415`

rohitpaulk:
Hmm, I can't find the exception on Sentry! Tried using the exception reference number, that didn't work.

rohitpaulk:
Exceptions fixed in #4468

rohitpaulk:
This looks like yet another case of the same problem, this time the package name is `animatetypewriter`

rohitpaulk:
I inserted a fake package for that, and we're stuck again! 

This time, at `protobufjs5-fixpackedfields` (seq: `1908498`)

rohitpaulk:
I think we should go ahead with adding logs for deleted packages

tests/py/test_sync_npm.py
whit537: Gotcha. Before af41409c5bea0f3cea3afbd804e64c83ef97c693 this tested something other than our test function. Out of scope here but we could/should write a test or three for [`gratipay.cli.sync_npm.main`](https://github.com/gratipay/gratipay.com/blob/2080/gratipay/cli/sync_npm.py#L15-L40).

whit537:
Alright, so our hypothesis is that this is either a bug or an unexpected feature of CouchDB, yes?

I guess in either case there's not much we can do but work around it, unsettling as it may be. Since both Papertrail and apparently now Sentry (#4468) are somewhat unreliable, does it make more sense to log an event in our own `events` table instead using `add_event`?

rohitpaulk:
> Since both Papertrail and apparently now Sentry (#4468) are somewhat unreliable, does it make more sense to log an event in our own events table instead using add_event?

Yes, it does. Change incoming..

rohitpaulk:
I _want_ to use [`add_event`](https://github.com/gratipay/gratipay.com/blob/5228f5c8bf2f4c8eaec34d2931c263a25df6d8b9/gratipay/application.py#L60), but it is part of `gratipay.Application`, and passing that down to `sync_npm` doesn't seem like the best idea, seems better when dependencies are explicit. I could instantiate `gratipay.Application` and then pass the `db` and `add_event` function separately, but passing functions between modules leads to confusion as far as I've observed. I see two viable options: 

- Move `add_event` to `utils`
- Create an `EventStorage` object, with an `add` function. Initiate it in `gratipay.Application`, and pass it down to `sync_npm`

rohitpaulk:
Ah, wait.. I could just add it to the `Package` model as a class function.

rohitpaulk:
Done.

rohitpaulk:
Heard back from couchDB - this is intended. 

https://issues.apache.org/jira/browse/COUCHDB-3418

> Hi Paul,
> 
> This is expected. Database compression means that only the most recent revision required to successfully replicate the database is present. In the case of a deleted document, that means that only the "tombstone" (document ID + revision + "_deleted": true) is retained. This is done so that if the document is re-created later (with a revision 1 greater than the previous revision) we can successfully detect that this newer revision isn't, for instance, the original revision of that document. This is the same for all documents - older revisions will be cleaned away during database compression.

Also, from their docs: 

> Normal deletion of a document within CouchDB does not remove the document from the database, instead, the document is marked as _deleted=true (and a new revision is created). This is to ensure that deleted documents can be replicated to other databases as having been deleted. This also means that you can check the status of a document and identify that the document has been deleted by its absence.

That explains https://gratipay.slackarchive.io/gratipay/page-95/ts-1493996587423821 :)

whit537:
Awesome. So do we still need to log deletions, if this is expected behavior?

rohitpaulk:
> So do we still need to log deletions, if this is expected behavior?

Nope, we don't. Popping the [last commit](https://github.com/gratipay/gratipay.com/pull/4464/commits/08586b26414bf9414c843a5a74821b499f2f4d96) off...

rohitpaulk:
Amended the comment text: 

screen shot 2017-05-17 at 9 04 10 am

PR #4469, opened 2017-05-14
Improve fake data performance

rohitpaulk:
This speeds up fake data a bit, and extracts a few functions out to improve readability


rohitpaulk:
Ready for review

Issue #4313, opened 2017-01-31
Homepage project loading in tabs


JessaWitzel:
Currently the homepage loads all of the tabs and all of projects on those tabs at once. 

Make it so that the teams on a tab only load when you click on that tab or when you click directly to a link of that tab. "Make an AJAX endpoint to load projects by status."

JessaWitzel:
Reminding myself that tabs.js is the place to be.

whit537:
[slack](http://gratipay.slackarchive.io/gratipay/-/1485468572.003881/1485894090.003998/1485882378003954/) ftr

whit537:
I think as part of this we should address at least some of the page performance characteristics from https://github.com/gratipay/gratipay.com/issues/4262:

-----------------------------

Users notice if sites and apps don't perform well. These top-level metrics capture the most important perceived performance concerns.

**First meaningful paint:** _9689.3ms_
First meaningful paint measures when the primary content of a page is visible.
- [ ] Target: 1,600ms

**Perceptual Speed Index:** _5186_
Speed Index shows how quickly the contents of a page are visibly populated.
- [ ] Target: 1250

**First Visual Change:** _916ms_

**Last Visual Change:** _14079ms_

**Estimated Input Latency:** _1706ms_
The score above is an estimate of how long Gratipay takes to respond to user input, in milliseconds. There is a 90% probability that a user encounters this amount of latency, or less. 10% of the time a user can expect additional latency.
- [ ] Target: 50ms

**Time To Interactive:** _14186.8ms_
Time to Interactive identifies the time at which Gratipay appears to be ready enough to interact with.
- [ ] Target: 5,000ms

**Critical Request Chains:** _27_
The Critical Request Chains below show you what resources are required for first render of this page. 
**Longest request chain (shorter is better):** _3_
**Longest chain duration (shorter is better):** _8296.49ms_
**Longest chain transfer size (smaller is better):** _449.78KB_

```
/ (gratipay.com)
|-- css/fonts.css (cloud.typography.com)
|-- 244520/1BA9B830F035F9E44.css (downloads.gratipay.com) - 4556.72ms, 800.64KB
|-- /gratipay.css (assets.gratipay.com) - 1110.17ms, 466.45KB
|-- /AAEAAAABc3MwMQAIAAAAAQAAAAIABgAOAAYAAAABABAAAQAAAAEAKAABAAgAAQA... () - 6743.56ms, 449.78KB
[...]

PR #4470, opened 2017-05-15
Homepage server tabs

rohitpaulk:
https://github.com/gratipay/gratipay.com/issues/4313#issuecomment-301209391

Partly solves #4313, minus the AJAXness

### DIFF
--- a/gratipay/utils/__init__.py
+++ b/gratipay/utils/__init__.py
@@ -243,25 +243,24 @@ def to_javascript(obj):
 
 
 def get_featured_projects(db):
-    teams = db.all("""
+    npopular, nunpopular = db.one("""
 
-        SELECT teams.*::teams
-          FROM teams
-         WHERE not is_closed
-           AND is_approved
-      ORDER BY ctime DESC
+        WITH eligible_teams AS (
+            SELECT *
+              FROM teams
+             WHERE not is_closed
+               AND is_approved
+        )
 
-    """)
+        SELECT (SELECT COUNT(1)
+                  FROM eligible_teams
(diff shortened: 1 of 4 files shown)


rohitpaulk:
Preview (low quality gif)

homepage

mattbk:
Would storing the "Featured" projects and only refreshing daily or weekly save any time as well?

mattbk:
I can confirm that this works for me, but have not tested with a large number of teams.

rohitpaulk:
> Would storing the "Featured" projects and only refreshing daily or weekly save any time as well?

Yes, it would. I'm not very optimistic on the gains though. All we'll avoid is a DB call (which we haven't proven to be an expensive query or so). Out of the homepage request + the other 10 image requests being made, the proportion of time saved by caching the result of that one DB call doesn't seem like it'll be much.

Why I think it is sufficient to do just those two steps is: Every request to the homepage (or any other page that loads 200 teams) fires 200 other requests for team images (which in turn make DB calls), and itself takes a while to send data down the wire. The gains here are real, and proven on a similar environment.

A couple of other things I thought of but haven't proven that they'll be worth it: 

- Use ETag caching for team images
- Collapse all image requests into 1 (either by moving images to an external service that supports HTTP2 + caching, or by using a custom format and splitting on client side)

whit537:
Another option for images: put a CDN in front of it while using Gratipay.com as the origin, as we do with CSS/JS assets.

whit537:
[slack](https://gratipay.slackarchive.io/gratipay/page-100/ts-1494883039396305)

Issue #4479, opened 2017-05-18
Takeover with teams is 500


whit537:
Choking on setting takes to zero, wants a national identity on file.

https://sentry.io/gratipay/gratipay-com/issues/275904408/

Reticketed from https://github.com/gratipay/project-review/issues/426.

whit537:
Closing in favor of #4480.

PR #4480, opened 2017-05-18
Allow team owners to merge accounts

whit537:
Fix for https://github.com/gratipay/gratipay.com/issues/4479.

### DIFF
--- a/gratipay/models/participant/__init__.py
+++ b/gratipay/models/participant/__init__.py
@@ -341,6 +341,14 @@ class Participant(Model, Email, Identity, ExchangeRoutes):
         """Leave all teams by zeroing all takes.
         """
         for team in self.get_teams():
+            if team.owner == self.username and team.get_take_for(self) == 0:
+
+                # Team owners don't necessarily have takes, and we also don't
+                # hard-require team owners to have a a national identity on
+                # file. But we *do* require national identity before we will
+                # set a take--even if setting it to zero.
+
+                continue
             team.set_take_for(self, ZERO, recorder=self, cursor=cursor)
 
 
@@ -1253,6 +1261,13 @@ class Participant(Model, Email, Identity, ExchangeRoutes):
                 # ==========================
 
                 cursor.run(MERGE_EMAIL_ADDRESSES, dict(live=self.id, dead=other.id))
+
+                # Take over team ownership.
+                # =========================
+
+                cursor.run( "UPDATE teams SET owner=%s WHERE owner=%s"
+                          , (self.username, other.username)
+                           )
 
                 # Disconnect any remaining elsewhere account.
                 # ===========================================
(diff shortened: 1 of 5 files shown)


whit537:
Ready for review, @rohitpaulk. 🙇

gratipay/testing/harness.py
```diff
@@ -211,17 +211,19 @@ def make_admin(self, username='admin'):
...
         """Factory for packages.
         """
         self.db.run( 'INSERT INTO packages (package_manager, name, description, emails) '
                      'VALUES (%s, %s, %s, %s) RETURNING *'
                    , (package_manager, name, description, emails)
                     )
         package = Package.from_names(NPM, name)
-        if claimed:
+        if claimed_by:
+            if claimed_by == 'picard':
```
rohitpaulk: ... Why hardcode `picard` here? Couldn't we just check if the participant exists and create it if not (like the other similar functions)

whit537: You good with c9f3ecea620ec5488689ea7d08355e3bf5d3db1e?

whit537:
Rebased, was c9f3ecea620ec5488689ea7d08355e3bf5d3db1e.

gratipay/models/participant/__init__.py
```diff
@@ -341,6 +341,14 @@ def clear_takes(self, cursor):
...
         """
         for team in self.get_teams():
+            if team.owner == self.username and team.get_take_for(self) == 0:
+
+                # Team owners don't necessarily have takes, and we also don't
+                # hard-require team owners to have a national identity on
+                # file. But we *do* require national identity before we will
+                # set a take--even if setting it to zero.
+
+                continue
```
rohitpaulk: Hmm, this case only arises because we're pulling teams from `self.get_teams`, which considers both teams that a participant is an owner of and those where the participant is a member (code below). 

```python
teams = (cursor or self.db).all("""
            SELECT teams.*::teams FROM teams WHERE owner=%s

            UNION

            SELECT teams.*::teams FROM teams WHERE id IN (
                SELECT team_id FROM current_takes WHERE participant_id=%s
            )
        """, (self.username, self.id))
```

What we need... is to only pull the teams where the user has an active take. Might be worth splitting into a separate query?

rohitpaulk: Another option, is to check whether the user is a member using the [`p.member_of(team)`](https://github.com/gratipay/gratipay.com/blob/master/gratipay/models/participant/__init__.py#L796) function. It'll blow up the number of queries, but taking over a participant is a rare event - I wouldn't worry about it right now.

rohitpaulk: Going to take a shot at this..

rohitpaulk: @whit537 Does https://github.com/gratipay/gratipay.com/pull/4480/commits/d723c1ad11d6641bfa270f0a34f31bca3ace502e look good to you?

Issue #4416, opened 2017-04-21
Bulk-add from `/on/npm/` based on verified email addresses


whit537:
Reticketed from https://github.com/gratipay/gratipay.com/pull/4305.

API should mostly be there.

whit537:
This would also give us a landing page that we could use for Product Hunt under https://github.com/gratipay/inside.gratipay.com/issues/1045. Gonna see if I can knock something out right quick ...

whit537:
![screen shot 2017-05-18 at 11 37 09 am](https://cloud.githubusercontent.com/assets/134455/26210662/59b897be-3bbe-11e7-8841-7f4cc219727f.png)

whit537:
Shelving a diff against 6f25fe2597b681569fbc32cda9c2870f69b3d1ad.

```diff
diff --git a/www/on/npm/%package/index.html.spt b/www/on/npm/%package/index.html.spt
index 7066653..7f6c5a5 100644
--- a/www/on/npm/%package/index.html.spt
+++ b/www/on/npm/%package/index.html.spt
@@ -55,7 +55,7 @@ if user.participant:
     {{ _( 'Apply to accept payments for the {package_link} npm package:'
         , package_link=('<a href="' + package.remote_human_url + '">' + package_name + '</a>')|safe
          ) }}
- </p>
+</p>
 
 {% if user.ANON %}
     <div class="button-wrapper">
diff --git a/www/on/npm/index.html.spt b/www/on/npm/index.html.spt
index e751da7..85d083c 100644
--- a/www/on/npm/index.html.spt
+++ b/www/on/npm/index.html.spt
@@ -1,9 +1,29 @@
 [---]
-title = "Search npm"
+banner = "npm"
+page_id = "package"
 suppress_sidebar = True
-website.redirect('/')  # nothing here yet
+npackages, Npackages = website.db.one('''
+    select (select count(*) from teams_to_packages) as n, (select count(*) from packages) as t
+''')
 [---]
 {% extends "templates/base.html" %}
 {% block content %}
-<input type="text" name="q"> <button class="selected">Search</button>
+
+<p class="description important-thing-at-the-top">
+    {{ _( "{n} out of {N} npm packages are on Gratipay."
+        , n=format_number(npackages)
+        , N=format_number(Npackages)
+         ) }}
+</p>
+
+<p class="instructions">
+    {{ _('Apply to accept payments for your npm packages:') }}
+</p>
+
+{% if user.ANON %}
[...]

PR #4488, opened 2017-05-26
Add bulk claiming at /on/npm/

whit537:
Picking up from #4416.

### Todo

- [x] move package API to a mixin: #4495
- [x] insert into `emails` during `make_participant`: #4495
- [x] add `get_packages_for_claiming`
- [x] modify `get_packages_for_claiming` to sort/group as desired
- [x] link to the email settings page
- [x] set the avatar as the npm logo
- [x] disable apply when there are no claimable packages
- [x] indicate when a package is already claimed, differentiating claimed-by-self and -by-other
- [x] fix test suite
- [x] wire up to `emails/modify.json`, send to address displayed
- [x] verify that we test what happens when someone tries to claim an already claimed package
- [x] ~~reuse open review ticket where possible~~—bumping to #4505
- [x] tie into review process better

### DIFF
--- a/www/on/npm/index.html.spt
+++ b/www/on/npm/index.html.spt
@@ -2,8 +2,9 @@ from gratipay.utils import icons
 [---]
 banner = manager = "npm"
 suppress_sidebar = True
-npackages, Npackages = website.db.one('''
-    select (select count(*) from teams_to_packages) as n, (select count(*) from packages) as t
+npm_stats = website.db.one('''
+    select (select count(*) from teams_to_packages) as claimed_packages
+         , (select count(*) from packages) as total_packages
 ''')
 if user.participant:
     packages_for_claiming = user.participant.get_packages_for_claiming(manager)
@@ -122,8 +123,8 @@ if user.participant:
 {% endif %}
 <p class="fine-print">
     {{ _( "{n} out of all {N} npm packages are on Gratipay."
-        , n=format_number(npackages)
-        , N=format_number(Npackages)
+        , n=format_number(npm_stats.claimed_packages)
+        , N=format_number(npm_stats.total_packages)
          ) }}
 </p>
 {% endblock %}
--- a/www/on/npm/index.html.spt
+++ b/www/on/npm/index.html.spt
@@ -1,9 +1,129 @@
+from gratipay.utils import icons
 [---]
-title = "Search npm"
+banner = manager = "npm"
 suppress_sidebar = True
-website.redirect('/')  # nothing here yet
+npackages, Npackages = website.db.one('''
+    select (select count(*) from teams_to_packages) as n, (select count(*) from packages) as t
+''')
+if user.participant:
+    packages_for_claiming = user.participant.get_packages_for_claiming(manager)
+    any_claimable = any([rec.claimed_by is None for rec in packages_for_claiming])
 [---]
 {% extends "templates/base.html" %}
+
+{% block banner %}
+<a class="elsewhere" href="https://www.npmjs.com/">
+    <div class="avatar">
+        <img class="avatar" src="{{ website.asset('package-default-large.png') }}" />
+        <img class="platform" src="{{ website.asset('npm-n.png') }}" />
+    </div>
+    {{ super() }}
+</a>
+{% endblock %}
+
+{% block scripts %}
+<script>
+    $(document).ready(function() {
+        Gratipay.packages.initBulk();
+    });
+</script>
+{{ super() }}
+{% endblock %}
+
 {% block content %}
-<input type="text" name="q"> <button class="selected">Search</button>
+
+<p class="description important-thing-at-the-top">
+    {{ _('Free as in money.') }}
+</p>
+
+<p class="instructions">
+    {{ _('Apply to accept payments for your npm packages:') }}
+</p>
+
+{% if user.ANON %}
+    <div class="important-button">
+        {{ sign_in_using(button_class='large') }}
+    </div>
+{% else %}
+
+    {% if not packages_for_claiming %}
+    <p class="sorry">{{ _("No packages found.") }}</p>
+    {% else %}
+    <table class="listing">
+        {% for i, rec in enumerate(packages_for_claiming, start=1) %}
+        <tr>
+            <td class="item i{{i}}{% if rec.claimed_by %} disabled{% endif %}"
+                data-email="{{ rec.email_address }}" data-package-id="{{ rec.package.id }}">
+                <img class="avatar" src="{{ website.asset('package-default-small.png') }}">
+                <div class="package-manager"><img src="{{ website.asset('npm-n.png') }}"></div>
+                <a class="listing-name" href="{{ rec.package.url_path }}">
+                    {{ rec.package.name }}
+                </a>
+
+                <div class="listing-details">
+                    <span class="i">{{ i }}</span>
+                    {% if rec.claimed_by %}
+                    <span class="owner">&middot;
+                    <span class="status-icon failure">
+                        {{ icons.STATUS_ICONS['failure']|safe }}</span>
+                        {% if rec.claimed_by == user.participant %}
+                        {{ _( "already claimed by {a}you{_a}"
+                            , a=('<a class="owner" href="{}">'|safe
+                                 ).format(rec.claimed_by.url_path)
+                            , _a='</a>'|safe
+                             ) }}
+                        {% else %}
+                        {{ _( "already claimed by {a}{owner}{_a}"
+                            , owner='~'+rec.claimed_by.username
+                            , a=('<a class="owner" href="{}">'|safe
+                                 ).format(rec.claimed_by.url_path)
+                            , _a='</a>'|safe
+                             ) }}
+                        {% endif %}
+                    </span>
+                    {% else %}
+                    <span class="owner">
+                        &middot;
+                        {% if rec.email_address_is_primary %}
+                        <span class="status-icon feature">
+                            {{ icons.STATUS_ICONS['feature']|safe }}</span>
+                        {% else %}
+                        <span class="status-icon success">
+                            {{ icons.STATUS_ICONS['success']|safe }}</span>
+                        {% endif %}
+                        {{ rec.email_address }}
+                    </span>
+                    {% endif %}
+                    <span class="description">&middot;
+                        {{ rec.package.description }}
+                    </span>
+                </div>
+            </td>
+        </tr>
+        {% endfor %}
+    </table>
+    {% endif %}
+
+    <div class="important-button">
+        <button type="submit" class="apply selected large"
+            {% if not any_claimable %}disabled{% endif %}>
+            {{ _('Apply to accept payments') }}
+        </button>
+    </div>
+
+    <p class="fine-print">
+        {{ _( "{a}Link an email{_a} from npm to see related packages."
+            , a=('<a href="/about/me/emails/">'|safe
+                 if packages_for_claiming else
+                 '<a href="/about/me/emails/" class="highlight">'|safe)
+            , _a='</a>'|safe
+             ) }}
+    </p>
+{% endif %}
+<p class="fine-print">
+    {{ _( "{n} out of all {N} npm packages are on Gratipay."
+        , n=format_number(npackages)
+        , N=format_number(Npackages)
+         ) }}
+</p>
 {% endblock %}
--- a/www/~/%username/emails/modify.json.spt
+++ b/www/~/%username/emails/modify.json.spt
@@ -48,8 +48,15 @@ if action in ('add-email', 'resend', 'start-verification'):
 
     participant.start_email_verification(address, *packages)
     if show_address_in_message:
+
+        # When reverifying an already-verified email (package claiming is a
+        # special case of this), then don't worry about content spoofing,
+
         msg = _("Check {email_address} for a verification link.", email_address=address)
     else:
+
+        # ... but otherwise, do: https://hackerone.com/reports/117187.
+
         msg = _("Check your inbox for a verification link.")
 elif action == 'set-primary':
     participant.set_primary_email(address)
--- a/www/~/%username/emails/modify.json.spt
+++ b/www/~/%username/emails/modify.json.spt
@@ -17,6 +17,7 @@ participant = get_participant(state, restrict=True)
 
 action = request.body['action']
 address = request.body['address']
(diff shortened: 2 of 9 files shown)


whit537:
Working up a ttw test.

screen shot 2017-05-31 at 5 07 01 am

whit537:
I'm working on this commit that actually adds the email address during `make_participant` (instead of just blindly inserting it). It's a simple change (fc36e0c5840f2854019075c2714949f67212eed1) but has follow-on effects for the rest of the test suite. I'm down to one more failure, and it's weird. It's in `test_teams.py`, and it's an interaction between several tests, with the failure manifesting in `test_error_message_for_slug_collision`. It's a failure to overwrite a VCR file—but! it only shows up if five other tests are present. Of the six I've got with it right now (a3872a50ed7f971bf8d37acc32795c04a97872a0) if any of the other five are gone the test passes. If all other six are present then it fails. 🐨

whit537:
Heck with it. Let's just rerecord the VCR fixture. 😉

whit537:
Moved off this PR into #4495.

whit537:
screen shot 2017-06-02 at 11 23 46 am

whit537:
Rebased on master, was 4587ba8a3e2989d385d5a6c91a941e5e09378ce5.

whit537:
Bringing over some local notes:

```
think through what happens when I put someone else's email in there                                           
    they get a link that's like "Claim your packages on Gratipay!"                                            
    they think, "Um, okay, why not?"                                                                          
        they think that we sent it to them                                                                    
    they click                                                                                                
    they are now sharing an account with an attacker                                                          
                                                                                                              
    I think we need to confirm again on landing                                                               
        "You are linking to thus and such account"                                                            
                                                                                                              
    this is a general problem with email verification, not specific here                                      
                                                                                                              
    At the very least we need to have the message in the email be "Sound familiar?"                           
        Try to communicate that they supposedly initiated the action
```

whit537:
screen shot 2017-06-05 at 11 33 37 am

whit537:
screen shot 2017-06-06 at 5 18 16 pm

whit537:
Ready for review, @rohitpaulk! Commits are a bit sprawling, pushing the limit on max lines delta. Lmk if I should factor out a PR for ease.

whit537:
Rebased, was c95af1e58150771db85f6772def8a8df07950f5b.

mattbk:
It all seems to work for me, but (as a new user), what do I do once I'm here? What's the next step?

image

mattbk:
I like the automatic PayPal route creation if a person doesn't have the PayPal route set up yet. :+1:

image

whit537:
Rebased (was 5d1e8095a152757341b50b3a4ea66f30b0650533) and added 51172012b4c8e07d21df90d010803bde8d598554:

screen shot 2017-06-12 at 10 52 42 am

whit537:
e18babf4afafae14f94bc930efcc6a3f02d2dd37:

```
------=_Part_1298180_670291361.1497279406886
Content-Type: text/plain; charset=UTF-8
Content-Transfer-Encoding: 7bit

Greetings!

Thanks for applying to use Gratipay! The next step is for us to review your application, which happens on this public GitHub issue:

some-github-issue

You can watch and participate in our review there, and we will email you again when we finish (usually in about a week). Thanks again!

Something not right? Reply to this email for help.

----

Sent by Gratipay, LLC, https://gratipay.com/
716 Park Road, Ambridge, PA, 15003, USA
------=_Part_1298180_670291361.1497279406886
```

whit537:
Better now, @mattbk? I ended up simplifying the `project-review.spt` template to allow reuse when there are multiple teams (by not referring to teams at all).

whit537:
https://travis-ci.org/gratipay/gratipay.com/builds/242054280 hit #4511. Restarting ...

whit537:
Travis found some regressions. I think 585f16e83f38bdec094d7f514ae069b62ca717a1 is clean?

whit537:
Module 05a1295195fdff005758c9778aaa9ad7de572dbe (i.e., unrelated test imp that gets squished in the next PR, #4507), yes.

whit537:
Rebased onto #4511 (which now includes 05a1295195fdff005758c9778aaa9ad7de572dbe), let's get that in first. Was 585f16e83f38bdec094d7f514ae069b62ca717a1.

mattbk:
If we're only sending one email, looks good to me. #4505 seems like the logical next step.

whit537:
Ya, that and https://github.com/gratipay/gratipay.com/issues/4512.

rohitpaulk:
Rebased. Was 60d0e85f7a6ca669c74697c13b8ba2dd9ea653cb, now 486158949bce0b9d30c2b3880304a6d74b2c04ca

gratipay/models/participant/packages.py
```diff
@@ -6,6 +6,61 @@
...
+    def get_packages_for_claiming(self, manager):
+
+        """Return a list of packages on the named ``manager`` for which the
+        participant has verified an email address on Gratipay, along with the
+        current package owner on Gratipay (if any).
+
+        :param string manager: the name of the package manager on which to look
+            for potential packages
+
+        :return: a list of (:py:class:`~gratipay.models.package.Package`,
```
rohitpaulk: This doc looks outdated - looks like we're returning tuples with 4 elements, and not 3. The `is_primary` element seems to be missing.

gratipay/project_review_process.py
```diff
@@ -31,12 +42,31 @@ def create_issue(self, *teams):
...
+        assert len(owner_usernames) == 1, owner_usernames
         body.extend(['', '(This application will remain open for at least a week.)'])
         data = json.dumps({'title': title, 'body': '\n'.join(body)})
-        return self._poster.post(data)
+        review_url = self._poster.post(data)
+
+        self.db.run("UPDATE teams SET review_url=%s WHERE id = ANY(%s)", (review_url, team_ids))
+        [team.set_attributes(review_url=review_url) for team in teams]
+
+        owner = Participant.from_username(owner_usernames.pop())
```
rohitpaulk: Might be better to use IDs instead of usernames for this (asserting we only have one unique value in the list + constructing a participant object)?

rohitpaulk: Oh, we don't have IDs on the teams table. Ignore.

gratipay/project_review_process.py
```diff
@@ -73,6 +105,8 @@ def __init__(self, fp=sys.stdout):
         self.fp = fp
 
     def post(self, data):
+        """POST data to nowhere and return a URL of lies.
```
rohitpaulk: 😃

gratipay/project_review_process.py
whit537: Yeah, #835.

rohitpaulk:
This is huge, going to try to split out some stuff...

rohitpaulk:
`get_packages_for_claiming` was modified in three commits - I extracted those changes out into #4513. Going to clean up the docs (https://github.com/gratipay/gratipay.com/pull/4488#discussion_r121636690) in that PR.

whit537:
#4515 factored out, was a42f0023dc76052f268ad5493537b4b3403b3fa5.

whit537:
Also factored out #4516 and #4517. More manageable now, @rohitpaulk? :-)

whit537:
Rebased, was 6fc28cc0829b86460224cdd9dfcc480327133032.

www/on/npm/index.html.spt
```diff
@@ -1,9 +1,129 @@
...
+</p>
+
+{% if user.ANON %}
+    <div class="important-button">
+        {{ sign_in_using(button_class='large') }}
+    </div>
+{% else %}
+
+    {% if not packages_for_claiming %}
+    <p class="sorry">{{ _("No packages found.") }}</p>
```
rohitpaulk: Might be worth adding a note saying that we searched using the email addresses a, b, and c - and to update on npm if needed (just like the page for a single package)

rohitpaulk: Oh wait, that is present below. Ignore.

www/on/npm/index.html.spt
```diff
@@ -1,9 +1,129 @@
+from gratipay.utils import icons
 [---]
-title = "Search npm"
+banner = manager = "npm"
 suppress_sidebar = True
-website.redirect('/')  # nothing here yet
+npackages, Npackages = website.db.one('''
```
rohitpaulk: I think the naming here could be better. These values are only being pulled so that we can show the `x packages out of y are on Gratipay` - might be better to keep them in a single dict so that a confusion doesn't arise between `npackages`, `Npackages` and `packages_for_claiming`. 

How about - ```npm_stats = website.db.one('''select () as total_packages, () as claimed_packages''', return_as=dict)```, followed by ```n=npm_stats.total_packages, N=npm_stats.claimed_packages``` (in the template)?

www/~/%username/emails/modify.json.spt
```diff
@@ -46,7 +47,10 @@ if action in ('add-email', 'resend', 'start-verification'):
         raise Response(400)
 
     participant.start_email_verification(address, *packages)
-    msg = _("Check your inbox for a verification link.")
+    if show_address_in_message:
```
rohitpaulk: Hmm, we always have an `address` - why not show it in the link?

rohitpaulk: i.e. Why do we have to say `Check your inbox` when we have an `address` available

rohitpaulk: Is it because we send multiple emails to addresses?

rohitpaulk: I only see two emails being sent in `start_email_verification` - one is a verification notice to the primary email, the other is the actual verification link to the address being verified.

whit537: Because it's a content spoofing vector. I thought we had a HackerOne ticket about this but it's not jumping out at me on http://inside.gratipay.com/appendices/disclosures.

rohitpaulk: Hmm, we allow reverifying email addresses? I see `new_email_verified=email in verified_emails` n `start_email_verification`. Digging deeper..

whit537: It's safe to show when the user is in control, but better not to for e.g. verifications, where a third party can initiate a message to any address.

rohitpaulk: This is confusing 😕 `verification.spt` is used both to verify emails and to verify linking packages to emails. The pattern of `if` conditions suggests that there isn't much in common either - I think it is best to separate these out.

rohitpaulk: > I think it is best to separate these out.

i.e. use separate functions (and templates) for verifying an email vs verifying linking an email to a package

rohitpaulk: > Because it's a content spoofing vector. I thought we had a HackerOne ticket about this but it's not jumping out at me on http://inside.gratipay.com/appendices/disclosures.

I think it'd be good to add a comment in the simplate that mentions this

EdOverflow: > Because it's a content spoofing vector. I thought we had a HackerOne ticket about this but it's not jumping out at me on http://inside.gratipay.com/appendices/disclosures.

It was around the time of [**h1:117187**](https://hackerone.com/reports/117187) and [h1:156542](https://hackerone.com/reports/156542).

whit537: Good find, @EdOverflow.

> I think it'd be good to add a comment in the simplate that mentions this

Comment in c3677d4943a4be57dd3f4ecb0aee16187f3c30c7, with link to H1.

I've also kicked off disclosure for those two. Seems they fell through the cracks. ☺️

whit537: > This is confusing 😕 

D'oh! 😞

> `verification.spt` is used both to verify emails and to verify linking packages to emails.

Yeah, it seemed to me that it was best to have one code path for both, since we want linking packages to emails to _also_ verify the email. That is, we don't want to make the user go through two verification flows (one for the address itself, a second for the packages) when from their point of view one should be sufficient. Am I thinking about that right?

> The pattern of `if` conditions suggests that there isn't much in common either

Are you referring to the `if action == 'start-verification'` block? That's the extra bit for the package claiming case, where we load the packages in question. The flows rejoin just after, in the call to [`start_email_verification`](https://github.com/gratipay/gratipay.com/blob/c3677d4943a4be57dd3f4ecb0aee16187f3c30c7/gratipay/models/participant/email.py#L57-L100), which is where the bulk of the common wiring lives.

> I think it is best to separate these out.
> i.e. use separate functions (and templates) for verifying an email vs verifying linking an email to a package

I agree it's pretty dense (especially when we get to [`finish_email_verification`](https://github.com/gratipay/gratipay.com/blob/c3677d4943a4be57dd3f4ecb0aee16187f3c30c7/gratipay/models/participant/email.py#L218-L254)), but I think it's necessary to support the simpler, single-step user experience during package claiming. Do you see a way to refactor while preserving the single-step flow?

www/on/npm/index.html.spt
whit537: Done in abfef42b687522c60b4061c6fa362fa6266e9bc0!

www/~/%username/emails/modify.json.spt
rohitpaulk: > Yeah, it seemed to me that it was best to have one code path for both, since we want linking packages to emails to also verify the email. That is, we don't want to make the user go through two verification flows (one for the address itself, a second for the packages) when from their point of view one should be sufficient. Am I thinking about that right?

I see how this makes sense. In that case, do we absolutely *need* to re-verify emails when claiming packages? If a user has a verified email, can't we just allow them to claim the package directly?

whit537:
Rebased, was abfef42b687522c60b4061c6fa362fa6266e9bc0.

www/~/%username/emails/modify.json.spt
whit537: Decided in [slack](https://gratipay.slackarchive.io/gratipay/page-100/ts-1497614829276345) to reticket dropping the re-verification.

rohitpaulk: ^ Discussed on [Slack](https://gratipay.slackarchive.io/gratipay/page-100/ts-1497615566425964).

whit537: Reticketed as #4520.

PR #4509, opened 2017-06-08
Adapt dashboard to CSP

whit537:
We implemented a content security policy in #4244 ff. It doesn't allow inline scripts or styles, nor `data:` URIs in src attributes.


rohitpaulk:
LGTM, waiting for travis

PR #4518, opened 2017-06-15
Start recording ref when recording an exchange in masspay

kaguillera:
As per @dmk246 suggestion [here](https://github.com/gratipay/gratipay.com/pull/3912#issuecomment-283485576) #4361 only took care of exchanges that recorded charges to participants. Masspay which is used for payouts were still recording NULL refs since there was no way to record a ref via `record_an_exchange.spt` 
This PR is to fix this issue.


kaguillera:
Ready for review @whit537 @rohitpaulk

rohitpaulk:
LGTM.

PR #4521, opened 2017-06-16
Don't restart count across ready/unready packages

whit537:
How about a single count across ready and unready packages.


whit537:
Eh, @rohitpaulk @mattbk et al.?

whit537:
screen shot 2017-06-16 at 12 14 01 pm

PR #4529, opened 2017-06-21
Simple fix for Chrome autofill

mattbk:
Replaces #4228, which I managed to mess up somehow when merging master into it.

> I looked at this again. I think it's ready for review once green. Essentially it doesn't pass the card number through our internal validation, which lets Chrome begin filling in the card number immediately. It also adds all the right tags to the fields for Chrome autocomplete to work.

https://github.com/gratipay/gratipay.com/pull/4228#issuecomment-310162804


mattbk:
Tests pass locally.

mattbk:
Darn it, I must be screwing something up locally.

PR #4534, opened 2017-06-26
Fix bug in package claiming when address verified

whit537:
When the address is already verified and the user starts a package claim twice, we aren't clearing out old nonces properly.

### DIFF
--- a/gratipay/models/participant/email.py
+++ b/gratipay/models/participant/email.py
@@ -163,6 +163,7 @@ class Email(object):
         existing = c.one( 'SELECT * FROM emails WHERE address=%s AND participant_id=%s'
                         , (email, self.id)
                          )  # can't use eafp here because of cursor error handling
+                            # XXX I forget what eafp is. :(
 
         if existing is None:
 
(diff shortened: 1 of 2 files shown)


tests/py/test_email.py
```diff
@@ -592,12 +592,26 @@ def test_claiming_package_with_an_address_not_on_file_is_a_no_go(self):
...
+        _start = lambda: self.alice.start_email_verification('alice@example.com', foo)
+        _nonce = lambda: self.db.one('select nonce from claims')
+        _start()
+        nonce = _nonce()
+        _start()
+        result = self.alice.finish_email_verification('alice@example.com', nonce)
+        assert result == (_email.VERIFICATION_FAILED, None, None)
+        assert nonce != _nonce()
+
+    def test_restarting_verification_also_clears_old_claims_when_address_preverified(self):
```
whit537: This is the new failing test.

whit537:
Rebased, was 11dabce20246d6ba5d04e1f85034609911d4df68.

whit537:
Ready for review, @rohitpaulk, et al.!

gratipay/models/participant/email.py
```diff
@@ -225,22 +226,51 @@ def finish_email_verification(self, email, nonce):
...
-                assert record.nonce is None  # and therefore, order of conditions matters
+
+                # We don't have that email address on file. Maybe it used to be
+                # on file but was explicitly removed (they followed an old link
+                # after removing in the UI?), or maybe it was never on file in
+                # the first place (they munged the querystring?).
+
+                return _fail
+
+            if record.nonce is None:
```
rohitpaulk: Why should we rely on the `nonce` being empty to deduce that the address was marked as verified, when we could just check `record.verified`? 🤔

whit537: The two conditions are not interchangeable, as [demonstrated](https://travis-ci.org/gratipay/gratipay.com/builds/252616241) by d44c59c6c4706c9983ff285943017ea39b9afca6. In other words, `nonce is None` is not just a proxy for whether it's marked as verified. Comment updated in 07e170329aa0ca49715b8d59580ff9cf6c6b9a31 to hopefully clarify.

Issue #1052, opened 2013-06-17
provide Medium-style email authentication


whit537:
We should provide [Medium-style email authentication](https://github.com/gratipay/gratipay.com/issues/1052#issuecomment-125301724).
### Old

We should provide password authentication. You should be able to use either your username and/or an email address (which one? only primary? configurable per-address?). This ticket started out being about signing up with email, but the actual important/hard part is the password authentication.

We first discussed email sign-up on #89, but that became more about linking email accounts to existing Gittip accounts rather than using email to sign up in the first place. Here's a ticket about signing up with email in the first place, in response to a request I received in private email:

> Is there a way to give money without signing up for Twitter, etc.?

We currently rely on social network profiles to detect fraud. It does limit our reach, however.
#### Notify:
- [ ] https://gratipay.freshdesk.com/helpdesk/tickets/2358
- [ ] https://gratipay.freshdesk.com/helpdesk/tickets/2749
- [ ] https://gratipay.freshdesk.com/helpdesk/tickets/3213
- [ ] https://gratipay.freshdesk.com/helpdesk/tickets/6349
##

whit537:
Some recent thinking on design for this:

http://blog.codinghorror.com/the-god-login/

rummik:
@whit537 I'd also recommend trying to help break the trend that [xkcd 936](http://xkcd.com/936/) illustrates by giving an example using multiple words, and possibly by pushing the term 'passphrase'.

Also, that was a good read :P  Not sure why I hadn't thought of making login attempts progressively slower as more failures occur, since that's something that's been done by Unix and Unix-likes for decades.

slaivyn:
+1 and thx for "the God login", very interesting!

whit537:
+1 https://gratipay.freshdesk.com/helpdesk/tickets/1667

whit537:
[Something to consider](https://twitter.com/homakov/status/577269918296633344) when we get around to implementing this:

> Attack on Rails apps: sign up as victim@gmail.com, copy _app_sess, wait for the victim to set a new password, then use saved session

seanlinsley:
Hah, did you see my RT of that? :stuck_out_tongue_winking_eye: 

The vulnerability being that Rails doesn't invalidate session tokens stored as cookies, and that the most popular authentication library for Rails (Devise) doesn't require email confirmation by default.

[...]

PR #4539, opened 2017-07-08
Add email auth

rohitpaulk:
https://github.com/gratipay/gratipay.com/issues/1052.

This PR allows users to sign-in or create an account on Gratipay using email.

### Screenshots

**Email HTML**

screen shot 2017-07-13 at 4 15 51 pm

**Email text**

screen shot 2017-07-13 at 4 03 42 pm

**Signup flow (GIF)**

gif

(http://recordit.co/ZJwXHLYiHF)

**Signin flow (GIF)**

(http://recordit.co/SX6hYCQI7j)

### Non-critical items to reticket

- [ ] Add cron to clean out old `email_auth_nonces`
- [ ] When we show 'link expired' upon verification, we only say 'Generate a new one'. Provide a button inline instead (or open the sign-in modal with the error message)

### Todo

- [ ] `-` instead of `_` in URLs—https://github.com/gratipay/gratipay.com/pull/4539#issuecomment-315895113: #4550
- [ ] tighten up UI on sign-in—https://github.com/gratipay/gratipay.com/pull/4539#issuecomment-315913082
- [ ] tighten up UI on `auth/verify`
- [ ] reclaim old account if using the same email? — https://github.com/gratipay/gratipay.com/pull/4539#issuecomment-315897100
- [ ] success notification on sign in—https://github.com/gratipay/gratipay.com/pull/4539#issuecomment-315897483
- [ ] better harmonize `Participant.with_*`—https://github.com/gratipay/gratipay.com/pull/4539#discussion_r127858756
- [ ] can we integrate emails and accounts elsewhere?—https://github.com/gratipay/gratipay.com/pull/4539#issuecomment-315921605
- [ ] tighten up some comments/copy: #4551


rohitpaulk:
Ready for review. I've taken care to structure the commits such that this PR is easy to break down into pieces if needed.

gratipay/exceptions.py
```diff
@@ -7,23 +7,23 @@
...
+class UsernameContainsInvalidCharacters(ProblemWithUsername):
     msg = "The username '{}' contains invalid characters."
 
-class UsernameIsRestricted(ProblemChangingUsername):
+class UsernameIsRestricted(ProblemWithUsername):
     msg = "The username '{}' is restricted."
 
-class UsernameAlreadyTaken(ProblemChangingUsername):
+class UsernameAlreadyTaken(ProblemWithUsername):
     msg = "The username '{}' is already taken."
```
EdOverflow: This can lead to username enumeration. Although usernames are public on Gratipay, I still think it is best practice to prevent username enumeration on the account creation endpoint.

rohitpaulk: @EdOverflow Is there a way around this, that doesn't sacrifice UX?

rohitpaulk: The way I see it, username enumeration is not a risk for us because usernames are public (as you mentioned). Whether they are exposed via this endpoint or a normal `/~username` endpoint shouldn't make a difference

EdOverflow: @rohitpaulk I have thought it through and agree with you. Lets leave it as it currently is.

whit537:
### To Review

- [ ] run it soup to nuts
- [ ] `_check_orphans`
- [ ] `with_email_and_username`, `from_email` — actually useful?
- [ ] proof-read comments and copy
  - [ ] "sign-in" &rarr; "sign in"
- [ ] `security/authentication/email` &rarr; `models/participant/email`
- [ ] `website_helpers`?
- [ ] button CSS
- [ ] test coverage
- [ ] `www/auth`
- [ ] merging accounts
- [ ]

whit537:
I think we should standardize on `-` instead of `_` in URLs.

```
$ find . -name \*-\*|wc -l
      60
$ find . -name \*_\*|wc -l
       8
$
```

whit537:
I think maybe we should allow reclaiming an existing account when signing up with the same email, as we do with accounts elsewhere.

screen shot 2017-07-17 at 2 56 16 pm

whit537:
I should get a notification when I follow a link from email to sign in.

whit537:
Picking up from #4543 &hellip;

> IMO that worked pretty well on #4488.

Sorry, I meant #4305. #4488 was a single PR (∆ = 359), whereas #4305 was an integration PR for a larger chunk of work (∆ = 3,628). We merged a couple dozen PRs into #4305 before merging the whole thing back to master. In my view it was a successful way of building out a fairly complex feature. I don't think this PR will be as big, but I like the idea of making PRs against it before merging to master.

What's behind your aversion to this method? Burned in the past? ... ?

mattbk:
Not sure I like how the login screen just sits there after you submit an email, and how the message at the top fades away so quickly. Maybe we should close the modal so people can look at projects while they wait for the email?

image

whit537:
For my part, a) I've never seriously used feature flags, but b) I'm disinclined towards them because I think they are overkill for our current level of complexity. There's only two-ish of us working on Gratipay right now. The main disadvantage with the integration branch approach is keeping it in sync as master moves (yes?), and that's not very hard at our current velocity. Feature flags are tech debt in our codebase itself. Integration branches essentially delegate code branching to git, which imo seems like a good place for it. ☺️ 

One way or another, we do need to find a development pattern that allows us to move quickly (by digesting small chunks at a time, ∆ <= 400), while also not pushing out incomplete features. I caught a quote from Mårten Mickos from HackerOne / MySQL that I like, along the lines of "MVP is only for the first iteration, after that you should only ship finished features." Room for interpretation in what constitutes "finished," of course, but it's often going to require ∆ > 400, I think.

whit537:
@mattbk Noted in todo in ticket description as "tighten up UI on sign-in"

whit537:
Re: feature flags:

https://martinfowler.com/articles/feature-toggles.html
https://martinfowler.com/bliki/FeatureToggle.html

Many caveats and best practices in there, e.g.:

> Savvy teams view the Feature Toggles in their codebase as inventory which comes with a carrying cost and seek to keep that inventory as low as possible. In order to keep the number of feature toggles manageable a team must be proactive in removing feature toggles that are no longer needed. Some teams have a rule of always adding a toggle removal task onto the team's backlog whenever a Release Toggle is first introduced. Other teams put "expiration dates" on their toggles. Some go as far as creating "time bombs" which will fail a test (or even refuse to start an application!) if a toggle is still around after its expiration date. We can also apply a Lean approach to reducing inventory, placing a limit on the number of toggles a system is allowed to have at any one time. Once that limit is reached if someone wants to add a new toggle they will first need to do the work to remove an existing toggle.

whit537:
I see so many similarities between email and accounts elsewhere. We talked about this before (should dig up): why don't we expand the accounts elsewhere machinery to encompass emails?

gratipay/models/participant/__init__.py
```diff
@@ -81,6 +81,28 @@ def with_random_username(cls):
         return cls.from_username(username)
 
     @classmethod
+    def with_email_and_username(cls, email, username):
+        """Return a new participant with given email and username
+        """
+        username = username and username.strip()
```
whit537: Why not just `username = username.strip()`

whit537: And why not push that down into `_validate`?

gratipay/models/participant/__init__.py
```diff
@@ -81,6 +81,28 @@ def with_random_username(cls):
         return cls.from_username(username)
 
     @classmethod
+    def with_email_and_username(cls, email, username):
```
whit537: I'd prefer to see this use the existing `username.safely_reserve` infrastructure for consistency and code reduction.

scss/components/modal.scss
```diff
@@ -46,7 +46,7 @@
     text-align: center;
     display: none;
     background: transparentize($light-brown, 0.3);
-    z-index: 1000;
+    z-index: 900;
```
whit537: I assume this is for a reason, ya?

mattbk:
Running into issues on my Macbook, but still running fine on the Ubuntu VM.

whit537:
Rebased, was 114f36c00de3eea050adc59680005df709052f79.

whit537:
Coming back here from https://github.com/gratipay/gratipay.com/pull/4604 ... I am thinking about taking the stub participant approach discussed at ... somewhere? We already have plumbing for routes and exchanges and email addresses, seems wrong to duplicate those to enable anon to make payments. Better to stub out a participant, send a normal email verification, and then populate the return page with a "Confirm Payment" section similar to how we could do package claims on the receiver side.

whit537:
> I am thinking about taking the stub participant approach discussed at ... somewhere?

https://github.com/gratipay/gratipay.com/issues/4547#issuecomment-315657487

rohitpaulk:
Closing as stale

Issue #4520, opened 2017-06-16
Don't require reverification for package claiming


whit537:
@rohitpaulk proposes at https://github.com/gratipay/gratipay.com/pull/4488#discussion_r122426725 and in [slack](https://gratipay.slackarchive.io/gratipay/page-100/ts-1497614829276345) that we drop the email dance when an email address is already verified, that we just send a notification ex post facto instead.

### Todo (from #4584)

- [ ] update claim action so it happens immediately (w/ notification to self)
- [ ] update bulk claiming so it happens immediately (w/ notification to self)
- [ ] implement verify action
- [ ] rip out claim-by-email wiring entirely

whit537:
Notification to claimer would line up with notifying other maintainers as well: #4425.

rohitpaulk:
There are two reasons I think this'll be a good idea: 

a) Easier for the user, one less step 
b) Decouples email verification from package claiming verification/notification - keeps our code simpler.

whit537:
The potential drawback is security concerns around claiming packages. This would be offset by an after-the-fact notification but not quite as tight. Do we care?

whit537:
> The potential drawback is security concerns around claiming packages.

It turns out through discussion culminating at https://github.com/gratipay/gratipay.com/issues/4557#issuecomment-324654121 that limiting the messages we send to unknown email accounts to just verifications is a better protection against phishing, because such verification spam is more of a known quantity to the general population and ignoring such messages is an already-learned behavior. Assigning any further agency to unsuspecting users is significantly more unsettling.

It would still be safer to require email confirmation for sensitive actions like linking packages (cf. #3969), but I agree that we shouldn't conflate claiming and verifying for unverified addresses.

> There are two reasons I think this'll be a good idea:

So security is a third reason.

whit537:
> So security is a third reason.

Sorry, that's not actually fully relevant here. This ticket is about when someone has _already_ verified an email address. The security concern has to do with when an email address is unverified.

PR #4543, opened 2017-07-11
Add email sign in - Part 1

rohitpaulk:
Part of https://github.com/gratipay/gratipay.com/issues/1052. https://github.com/gratipay/gratipay.com/issues/1052#issuecomment-309198597, specifically.

This adds a table to store email auth nonces + a few helper functions to create, verify and invalidate records in that table.

Controllers etc. in next PR (#4539). This can be merged and deployed independently of those changes.

**Non-critical items to be reticketed:** 

- [ ] Add cron to clean out old `email_auth_nonces`


rohitpaulk:
Ready for review

gratipay/models/participant/email.py
```diff
@@ -54,6 +54,20 @@ class Email(object):
 
     """
 
+    @classmethod
+    def from_email(cls, email_address):
```
whit537: I would expect this be in the __init__.py with the other constructors. No?

gratipay/security/authentication/website_helpers.py
```diff
@@ -86,3 +80,8 @@ def add_auth_to_response(response, request=None, user=ANON):
     if SESSION in request.headers.cookie:
         if not user.ANON:
             user.keep_signed_in(response.headers.cookie)
+
+def start_user_as_anon():
+    """Make sure we always have a user object, regardless of exceptions during authentication.
+    """
+    return {'user': ANON}
```
whit537: 👍

sql/branch.sql
```diff
@@ -0,0 +1,16 @@
+BEGIN;
+    CREATE TYPE email_auth_intent AS ENUM ('sign-in');
```
whit537: What are you thinking the other intents will be?

sql/branch.sql
```diff
@@ -0,0 +1,16 @@
...
+
+    CREATE TABLE email_auth_nonces
+    ( id                    serial                      PRIMARY KEY
+    , email_address         text                        NOT NULL
```
whit537: Doesn't this duplicate `emails.address`?

sql/branch.sql
```diff
@@ -0,0 +1,16 @@
...
+    , email_id              integer                     REFERENCES emails
+                                                          ON UPDATE CASCADE
+                                                          ON DELETE CASCADE
+    , intent                email_auth_intent           NOT NULL
+    , nonce                 text                        NOT NULL
+    , ctime                 timestamp with time zone    NOT NULL DEFAULT CURRENT_TIMESTAMP
+    , UNIQUE (email_address, nonce)
```
whit537: Since we're using UUID for the nonce, `UNIQUE(nonce)` should be enough, no?

whit537:
Made some drive-by comments on first skim. Will go deeper on next pass ...

gratipay/models/participant/email.py
rohitpaulk: Hmm, yh probably better placed there.

sql/branch.sql
rohitpaulk: `sign-up`

rohitpaulk: The purpose of storing this is that for `sign-in`, we have to make sure that an email reference (to the `emails` table) is maintained. When an email is deleted, this record should be deleted too. (Which reminds me, got to add a test for that) This prevents the case where a user mistakenly adds an email address, and an attacker verifies it and uses it to log in. (Not sure how likely it is, but plausible)

sql/branch.sql
rohitpaulk: We won't always have a reference to the `emails` table, this is only for `sign-in`. (check the `sign_in_has_valid_email` constraint below).

sql/branch.sql
rohitpaulk: Yes, should be. Will amend

rohitpaulk: Fixed.

gratipay/models/participant/email.py
rohitpaulk: Fixed.

sql/branch.sql
rohitpaulk: > Which reminds me, got to add a test for that

Added.

rohitpaulk: > The purpose of storing this is that for sign-in, we have to make sure that an email reference (to the emails table) is maintained. When an email is deleted, this record should be deleted too. (Which reminds me, got to add a test for that) This prevents the case where a user mistakenly adds an email address, and an attacker verifies it and uses it to log in soon after the user has deleted it. (Not sure how likely it is, but plausible)

Just realized that we don't need this. When the email link is hit, we'll have to verify that the email (and linked participant) exists anyway. Removing.

rohitpaulk: Removed.

sql/branch.sql
rohitpaulk: Note: Removed the reference to `emails`, not needed

whit537:
@rohitpaulk What do you think of 1e5ada688a5c7a389a5d22fc9aeba322d2462034? The API seemed parallel enough to include there instead of under `gratipay/security`.

whit537:
Honestly I'd be more comfortable integrating this and downstream work in a feature branch before merging to master. IMO that worked pretty well on #4488. It gives us a chance to iterate on the API more freely than if we are deploying to master at each step (it also avoids the feature-flag complication). We may decide to vary the API or even the database schema based on downstream considerations. You willing to give that a try, @rohitpaulk?

whit537:
If so, we can simply leave this open as our integration branch/PR, and merge #4539 and additional work into here before finally landing.

rohitpaulk:
> What do you think of 1e5ada6? The API seemed parallel enough to include there instead of under gratipay/security.

Hmm, that might be a problem when implementing sign up. The decoupling was intentional, we won't always have a `Participant` object available when dealing with nonces.

rohitpaulk:
> Honestly I'd be more comfortable integrating this and downstream work in a feature branch before merging to master. IMO that worked pretty well on #4488. It gives us a chance to iterate on the API more freely than if we are deploying to master at each step (it also avoids the feature-flag complication). We may decide to vary the API or even the database schema based on downstream considerations. You willing to give that a try, @rohitpaulk?

#4488 was merged in a hurry before sustain, not the best example IMO. Making changes to the API was hard then too (even though we used layered PRs and didn't merge stuff early), and we decided to reticket [a major change](https://github.com/gratipay/gratipay.com/issues/4520). 

I understand how this can be difficult to review though. Half of the work that builds upon this is up in my head and not documented anywhere. I'll work on sign-up a bit before resurfacing this for review. This change is currently at ~500 lines modified, including both PRs. I'll aim for a cap of ~1000 lines modified. If it goes past that, let's find a way to merge a safe enough portion in early? It is difficult to maintain multiple PRs and >1000 lines outstanding.

Closing this PR, keeping the other open.

whit537:
Okay, let's see how that works.

PR #4546, opened 2017-07-14
Cleanup schema.sql

rohitpaulk:
This PR cleans up the schema definitions in `sql/schema.sql`. Every commit was verified to maintain parity with production, using `make schema-diff`.


rohitpaulk:
I considered squashing commits, but thought that it'll be easier to follow git history if we retain the small & targeted changes

whit537:
I don't like that this leaves the PR URLs in comments. Those only made sense when the file was structured as an append-only log. As it stands now, I'm not sure how the file *is* structured anymore.

Can we dig up the last go-round on this for ideas? I think that's when we sprouted `make schema-diff`.

whit537:
I want to read this file and have it tell a story. What story is it telling now?

rohitpaulk:
> I don't like that this leaves the PR URLs in comments. Those only made sense when the file was structured as an append-only log. As it stands now, I'm not sure how the file is structured anymore.

Agreed, I'll remove the PR URLs. 

> I want to read this file and have it tell a story. What story is it telling now?

Once PR URLs are removed, the story should correspond to exactly what the file is named - `schema.sql`. Just that the bottom will slowly grow as an append-only log for a while, until we compact it again.

rohitpaulk:
> Agreed, I'll remove the PR URLs.

Done.

whit537:
!m @rohitpaulk

Issue #4545, opened 2017-07-13
Send email messages immediately


rohitpaulk:
The current email_queue implementation we have polls every 60 seconds for pending emails. This works fine for cases where delivery time isn't noticeable/important (payday emails, for example). 

In scenarios where users expect an email immediately (verification link when adding an email, [email sign-in links](https://github.com/gratipay/gratipay.com/pull/4543)), a delay anywhere between 0-60 secs will be observed. We can make this better by adding event-based notifications in addition to the current polling mechanism. i.e. every time we insert an email into the database, we fire an event. The email_queue listens for that event and processes accordingly. [Postgres supports `LISTEN`/`NOTIFY`](https://www.postgresql.org/docs/9.0/static/sql-notify.html).

rohitpaulk:
Note: this will require some thinking about throttling. Currently, we throttle based on how many emails are in the queue already. If queue processing becomes almost instantaneous, then throttling will become ineffective.

whit537:
@Changaco had input on this when we implemented throttling. He implemented an alternative on Liberapay, I don't remember the details off hand but we could dig that up.

Changaco:
Actually I still haven't implemented a throttling mechanism in Liberapay: <https://github.com/liberapay/liberapay.com/issues/61>.

Issue #4547, opened 2017-07-15
Add support for queuing emails without a participant linked


rohitpaulk:
This is a requirement for [sign-up via email](https://github.com/gratipay/gratipay.com/issues/1052). 

Our `email_queue` currently only supports queuing emails linked to a participant. Extend this to support queuing emails with only an email address.

If a linked participant is present, stick to throttling based on `participant_id`. If only email_address is provided, throttle based on `email_address`.

whit537:
How about a stub participant like we do for accounts elsewhere?

rohitpaulk:
Hmm, that _could_ work. I'm on the edge here though, there are pros and cons to this approach. 

**Upside**

Schema + code is simplified in some places - we don't end up implementing the stuff that is common, can reuse some features.

**Downside**

We'll need special handling in a lot of places to differentiate claimed participants from unclaimed participants. It seems easier for bugs to creep in when we're not explicitly stating that the participant could be half-baked. For example, the changes we'll need to make if we go with this approach: 
  - `email_queue` will now have the handle the case of an `unclaimed` participant differently. We should be careful to not link the random username anywhere
  - We'll need to create a random participant the first time, and then pick the same participant every time after so that throttling based on `participant` works as intended. 

I see how this approach made perfect sense under Gittip/Gratipay 1 - A lot of the core features related to users who hadn't signed up yet (pledging etc). I'm not sure if we should continue with that pattern for auth, I do see benefits in being explicit about the inexistence of an actual user record to deal with. 

This posts explains my thoughts better: http://rdingwall.com/2009/11/20/the-trouble-with-soft-delete/ ('soft create' is what we're dealing with)

rohitpaulk:
@whit537 (pinging so that this appears in your notifications) thoughts?

whit537:
From http://rdingwall.com/2009/11/20/the-trouble-with-soft-delete/:

> Primary transactional tables should only contain data that is valid and active right now.

I think stub participants are valid and active. We are actively initiating package claims for them, no? And over on https://github.com/gratipay/gratipay.com/pull/4598 we want to start placing credit card holds for them as well.

whit537:
We want to send them email: https://github.com/gratipay/gratipay.com/pull/4548. That seems active.

whit537:
Honestly I'm still drawn to the idea of using `elsewhere` for email:

[...]

Issue #4425, opened 2017-04-26
Notify other maintainers on package claim


whit537:
Reticketed from #4305.

> This is complicated because our current infrastructure assumes that an email address is linked to a participant. Either we need to create stub participants for all relevant emails (we do something similar for social accounts, but then I guess we want to be able to "promote" the stubs? does that get us into #1052?) or undo that assumption so we can mail without having a participant.

whit537:
+1 from @rohitpaulk at https://github.com/gratipay/gratipay.com/pull/4579#issuecomment-324401774:

> The email body should clearly mention that the Gratipay user is a verified owner of that specific email address.

Issue #2993, opened 2014-12-09
catch mail bounces and mark as such in db/ui


whit537:
Reticketing from #2752.
##

whit537:
http://docs.aws.amazon.com/ses/latest/DeveloperGuide/sending-email-with-ses.html
https://aws.amazon.com/blogs/ses/goodbye-blacklist-introducing-the-suppression-list/

PR #4548, opened 2017-07-15
Support queueing emails without participants

rohitpaulk:
&larr; "Ready to verify" (#4584)

-------------

#4547 

- ~~Extracted out a few tests from `test_email.py` into `test_participant_emails.py`. Now, the tests in `test_email.py` only test stuff in `gratipay/email.py`~~ &rarr; #4570
- Moved `email` from within `context` to a new field in the DB. This helps in visibility + simplifies DB  queries and constraints
- Added support for queuing emails without participants

### Todo

- [ ] handle unsubscribes


whit537:
Rebased, was 3d2430c3d55d39facecd202d6aedf70a8a7a3d4f.

whit537:
Brainstorm: can we decouple participant entirely from the email queue? Is that a cleaner architecture?
I'm also considering adding a couple other improvements here:

- changing the email queue so we don't delete messages
- not sending more than one verification email at a time (to address #4557)
- possibly tweaking the copy on the verification message (also for #4557)

whit537:
We need this for https://github.com/gratipay/gratipay.com/pull/4569 in addition to #4539.

whit537:
Participant is related to throttling. I think we might be able to simplify throttling to only be about verified emails, rather than the generic `user_initiated` strategy we currently employ. Right?

whit537:
Yeah, from a review of `ack queue\.put` the only place we call `put` without passing in `_user_initiated=False` (thereby bypassing throttling) is [during verification](https://github.com/gratipay/gratipay.com/blob/2099/gratipay/models/participant/email.py#L83-L90).

whit537:
That means we can resolve #4545, too!

whit537:
Alright, let's try to keep this PR focused though ...

whit537:
Refactoring throttling will make it easier to extricate participant from `gratipay.email`.

whit537:
I've almost got a throttling rewrite done, which I plan to make as a separate PR ahead of this one. But now I think that we actually want _both_: throttling based on "N per time", and only allowing one open verification at a time. The reason is that we still need to allow resending verification messages, and you can be sure we'll get turkies flooding some poor maintainer with N00 verification resends if we don't throttle that. I guess we could also do that as N resends per address? Anyway, the point is that to implement (some variant) of both protections we'll need to upgrade `email_queue` to an `email_messages` table where we store email messages in perpetuity, so that we can query on recent past. I guess that's a third PR ahead of _both_ of these.

whit537:
Throttling redo started in #4571. Turning to permanentizing email messages table ...

whit537:
Email saving in #4572.

whit537:
Throttling redo redone in #4579.

In addition to better protection against verification spam/phishing, the other abuse protection we need before we can start mailing just any ol' email address is a blacklist/unsubscribe.

whit537:
We should probably treat unsubscribes as locks against claims via that email address across all packages.

whit537:
Moving out of review, we need unsubscribe as part of this. We can't start sending messages to random email addresses without an unsubscribe/blacklist mechanism.

whit537:
Rebased on #4584, was 4bc67bc18c489b621dfac5a0cc6f29d55ecfb7fe.

whit537:
I'm seeing confusion in the `email_messages.participant` field. It conflates the sender and the recipient of messages. It's the recipient in cases where no `email` address is specified in the context, and it's also treated as the recipient for purposes of templating (language, username, etc.[?]). We treat `participant` as the sender for purposes of throttling.

So far in 27d4c8f61c5a7699e51ec0544f0d9bd07c74fa56 on this PR we are bumping `context.email` up to `email_messages.email_address`, but the conflation of sender and receiver remains: the `put` API accepts _either_ a participant (`to`) _or_ an email address (`email`). This sort of works on the recipient side because we can just fall back to sane-ish defaults for templating context, but when get to the sender side we fall into a bit of an absurdity, [possibly throttling](https://github.com/gratipay/gratipay.com/blob/27d4c8f61c5a7699e51ec0544f0d9bd07c74fa56/gratipay/email.py#L116-L131) on what can only be the recipient (`email`).

I think we need to update the `put` API to clearly distinguish the sender and the recipient of messages:

- [ ] Replace the `email_messages.user_initiated` boolean with a `sender` field referencing `participants`. `NULL` means it's a system-initiated message.
- [ ] Add a required `to` field to `email_messages` that holds the recipient email address. Make it the caller's responsibility to provide all context variables required by the template rather than assuming participant and attempting defaults.

I'm not sure yet how to think about unsubscribe/blacklisting, but the above should be cleaned up one way or another, maybe in its own PR.

whit537:
I think blacklisting should be its own table, `email_blacklist`. When we get a spam report under #2993 it should go on the blacklist. When a recipient clicks unsubscribe on our pledge notifications under #4425 or #4569 it should go on the blacklist. We shouldn't any non-verification emails to "unbound" email addresses (not linked to a participant who initiated the message) that are on the blacklist. I'm not sure we should even send verification messages, in which case they'd have to email us from that address in order to unblacklist it.

whit537:
This'll require a complex migration since we want to extract email address out of the `context` field which is a Python pickle and therefore opaque to Postgres. I've implemented deploy hooks in #4590 to help with that.

whit537:
Rebased onto #4584, was 27d4c8f61c5a7699e51ec0544f0d9bd07c74fa56.

whit537:
Interesting! I want to set up fixture using the old API, meaning old code old slug. But no `before.py`!

whit537:
Not to self: account for the case where we want to send payday charge notifications but the participant doesn't have an email address on file (cf. https://github.com/gratipay/inside.gratipay.com/issues/1149#issuecomment-326535433)

whit537:
Rebased, was c61f4d76023b4ac521fb5a48fd54b497530486c2.

whit537:
Alright, upon closer inspection, `email_messages.participant` is actually the `recipient`, and so by also using that for throttling we are assuming that the recipient is always also the sender. Hmm ... 🤔

whit537:
And `user_initiated` is how we mark when the receiver is also the sender.

whit537:
We have two presenting use cases for sending email messages to addresses unassociated with a participant:

1. email sign-up (#4547), and
1. pledging (#4569).

Under (2) these are going to be system-generated messages, so we don't need to worry about throttling (well, we need to worry a lot about it, but it's on us to throttle ourselves! :D).

Under (1) we do want to throttle on the receiver (mailbombing is annoying) but even moreso on the sender (mass-mailbombing is worse, and mass-phishing is worser). That'll probably require logging IPs, out of scope on this PR.

whit537:
Rebased yet again to go back to @rohitpaulk's original 4bc67bc18c489b621dfac5a0cc6f29d55ecfb7fe w/ updates for #4590 #4579 #4584. Was 83bacdfec0bdcb4d8c1136094fa7c4501371351f.

whit537:
Half-baked notes on my desktop from when I was offline:

> The only place where we include the email address in the context (overriding any `participant.email_address`) is verification.
>
> Hah! Pass in recipient_address and dereference a participant based on that (`select * from email_addresses where address=%s and verified;`). If null then double-check that we're sending a verification message, we shouldn't need username, etc. at that point and we can default to eng.
>
> Dang, no. Email lang should come from the participant that initiated the message.
>
> Pass in lang to put.
>
> | sender | receiver | e.g.              | lang     |
> |--------|----------|-------------------|----------|
> | auth   | auth     | user messaging    | receiver |
> | anon   | auth     | n/a               | |
> | sys    | auth | | |
> | sys    | | | |
>
>  - for sender:auth sender-receiver (charged) it should 
>  - for anon sender-receiver (sign-in) it should come from http lang
> - for sys sender/anon receiver (pledging) it should be `en`
> - for auth sender-receiver (charged) it should

whit537:
We need this for https://github.com/gratipay/gratipay.com/pull/4598.

whit537:
Crap. Wrong base. :-/

whit537:
So now my options are:

- self-merge #4579 and #4584
- wait for someone else (@rohitpaulk? @dowski?) to merge those
- make a new PR with these commits against master

whit537:
I meant @kaguillera, not @dowski. Honest! 🐭

rohitpaulk:
Yh umm.. I'm confused - what did you intend to merge this into? `master`?

whit537:
Yes.

whit537:
[slack](https://gratipay.slackarchive.io/gratipay/page-100/ts-1504877781000439)

PR #4552, opened 2017-07-18
Whitelist the widget CSS for CSP

whit537:
This should resolve the worst of the CSP reports in https://github.com/gratipay/inside.gratipay.com/issues/1118.

### DIFF
--- a/gratipay/security/__init__.py
+++ b/gratipay/security/__init__.py
@@ -53,7 +53,7 @@ def add_headers_to_response(website, response):
             "default-src 'self';"
             "script-src 'self' assets.gratipay.com 'unsafe-inline';"
             "style-src 'self' assets.gratipay.com downloads.gratipay.com cloud.typography.com"
-            "          'sha256-WLocK7HeCKzQLS0M+PGS++5IhyfFsOA5N4ZCeTcltoo=';"
+            "          'sha256-WLocK7HeCKzQLS0M+PGS++5IhyfFsOA5N4ZCeTcltoo=';" # CSS on widget.html
             "img-src *;"
             "font-src 'self' assets.gratipay.com cloud.typography.com data:;"
             "block-all-mixed-content;"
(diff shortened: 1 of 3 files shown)


whit537:
Taking out of review. There are some other CSP violations on widgets.html that we should deal with on this PR as well.

whit537:
Actually, putting back in review because I think this takes care of the worst of it and I'm not instantly seeing how to make others fit in here (not able to trigger `script-src` right away).

gratipay/security/__init__.py
```diff
@@ -52,7 +52,8 @@ def add_headers_to_response(website, response):
         response.headers['content-security-policy-report-only'] = (
             "default-src 'self';"
             "script-src 'self' assets.gratipay.com 'unsafe-inline';"
-            "style-src 'self' assets.gratipay.com downloads.gratipay.com cloud.typography.com;"
+            "style-src 'self' assets.gratipay.com downloads.gratipay.com cloud.typography.com"
+            "          'sha256-WLocK7HeCKzQLS0M+PGS++5IhyfFsOA5N4ZCeTcltoo=';"
```
rohitpaulk: What is this value? 🤔 Will it change in the future?

whit537: It's a hash of the [`<style>`](https://github.com/gratipay/gratipay.com/blob/2097/www/~/%25username/widget.html.spt#L12-L104), which Google Chrome shows me in the developer console as part of its CSP error message. I copied and pasted it from there. It will need to be updated anytime that content changes (it hasn't [changed](https://github.com/gratipay/gratipay.com/commits/2097/www/~/%25username/widget.html.spt) in over two years). Commented in b5390d96d3d43806c5e4ccd0d116fcf5dac26108.

EdOverflow: In that case, couldn't we manually add a [Subresource Integrity](https://developer.mozilla.org/en-US/docs/Web/Security/Subresource_Integrity) hash to the widget?

rohitpaulk: @EdOverflow Wouldn't that mean that the users who embed the widget would have to change the value everytime we update our content?

EdOverflow: @rohitpaulk Yes, they would get a warning in the console, which would let them know that the hash needs updating.

PR #4555, opened 2017-07-26
Switch from SentEmailHarness to QueuedEmailHarness in notify_participant test

### LINKED ISSUE #4544: Switch from SentEmailHarness to QueuedEmailHarness in notify_participant tests
We have two test helpers for testing emails - [`SentEmailHarness`](https://github.com/gratipay/gratipay.com/blob/231af7959bdd85c33b0cb15a91170e9523f5f620/gratipay/testing/email.py#L51) and [`QueuedEmailHarness`](https://github.com/gratipay/gratipay.com/blob/231af7959bdd85c33b0cb15a91170e9523f5f620/gratipay/testing/email.py#L39). The former mocks the actual mailer call, whereas the latter picks values from the database queue (bypassing the mailer call). 

We don't use the `SentEmailHarness` almost everywhere in our test suite, except for 

1. [FlushEmailQueue](https://github.com/gratipay/gratipay.com/blob/231af7959bdd85c33b0cb15a91170e9523f5f620/tests/py/test_email.py#L398), which is valid - since we're testing the queue itself.
2. [TestNotifyParticipants](https://github.com/gratipay/gratipay.com/blob/231af7959bdd85c33b0cb15a91170e9523f5f620/tests/py/test_billing_payday.py#L619) - we shouldn't have to do this, `QueuedEmailHarness` should work just fine since `notify_participants` writes to the email queue

Misara:
Closes #4544.

### DIFF
--- a/tests/py/test_billing_payday.py
+++ b/tests/py/test_billing_payday.py
@@ -12,7 +12,7 @@ from gratipay.exceptions import NegativeBalance
 from gratipay.models.participant import Participant
 from gratipay.testing import Foobar, D,P
 from gratipay.testing.billing import BillingHarness, PaydayMixin
-from gratipay.testing.email import SentEmailHarness
+from gratipay.testing.email import QueuedEmailHarness
 
 
 class TestPayday(BillingHarness):
@@ -616,7 +616,7 @@ class TestTakes(BillingHarness):
         assert P('picard').balance  == D('200.00')
 
 
-class TestNotifyParticipants(SentEmailHarness, PaydayMixin):
+class TestNotifyParticipants(QueuedEmailHarness, PaydayMixin):
 
     def test_it_notifies_participants(self):
         kalel = self.make_participant('kalel', claimed_time='now', is_suspicious=False,


whit537:
Thanks for jumping in, @Misara! Do you have the test suite running locally? Are you able to reproduce the test failure we're seeing at Travis?

Misara:
Yes, I can see the test failure. This assert is passing:
`assert emails.spt_name == 'charge_'+status`
but these three are failing:
`
assert self.get_last_email()['to'] == 'kalel <kalel@example.net>'
assert 'Gratiteam' in self.get_last_email()['body_text']
assert 'Gratiteam' in self.get_last_email()['body_html']
`
I've tried to identify the issue, but I'm afraid I don't understand the test suite well enough to fix it. Sorry for messing around with it.

rohitpaulk:
@Misara - I think the issue here is with the [call to `flush()`](https://github.com/gratipay/gratipay.com/blob/231af7959bdd85c33b0cb15a91170e9523f5f620/tests/py/test_billing_payday.py#L636). `flush()` deletes items from the email queue in our DB, and tries to invoke the mailer. We shouldn't need that if using the `QueuedEmailHarness` - since our tests should pick values directly from the DB. In fact, calling `flush()` could be the reason that `get_last_email()` ends up as `None` - coz it doesn't exist in the DB anymore. Can you try removing the `flush()` call and see what happens? :)

If you'd like to run that specific test locally, you can use this command: `PYTEST="test_billing_payday.py -k notify -vvv"make pytest`.

Misara:
I actually tried that, the tests still failed. The self.get_last_email() call does not seem to return anything. 
Thanks for the command to run the specific test, that sure saves a lot of time

rohitpaulk:
Note: This was done in https://github.com/gratipay/gratipay.com/pull/4576, closing.

Issue #4557, opened 2017-07-31
Mitigate npm spam/phishing


mattbk:
Whether it's an accident or not, seeing this coming through Freshdesk.

> Someone [...] is trying to use my email to claim the [...] package on gratipay.com and I'm getting a bunch of request spam. Please deal with this loophole.  https://gratipay.freshdesk.com/helpdesk/tickets/9219

> What is this asking me to do? I'm unsure who this [...] is.
https://gratipay.freshdesk.com/helpdesk/tickets/9220

rohitpaulk:
We do have rate limiting already, I think it is 3 emails per minute or similar.

rohitpaulk:
Well, that _could_ end up in 1440 emails per day - so might be worth revising our algorithm.

whit537:
Alright, what's the right thing to do here?

whit537:
Since we're working with public emails that haven't been verified with us yet, there's no way we can close this off entirely without neutering the entire feature.

whit537:
I've marked the account doing the spamming as suspicious.

mattbk:
Considering that you need to be logged in to see the emails and click apply, I think limiting per account wouldn't be outside a realistic possibility. 

Or do like Slack does and require that you can only have n unconfirmed emails. So if you click "apply" on three projects that aren't yours, you won't be able to on the fourth one.

whit537:
> Or do like Slack does and require that you can only have n unconfirmed emails. So if you click "apply" on three projects that aren't yours, you won't be able to on the fourth one.

I like this idea.

whit537:
I'm trying to understand the damage here. How many mails did Spammer send?

mattbk:
I asked at https://gratipay.freshdesk.com/helpdesk/tickets/9219.

mattbk:
Three to the person above, and he is not happy about it.

whit537:
I've replied:

> Hi, []. Chad here, we met at Sustain. Sorry for the spam. :-(
>
>The reason this happened is because your email address at npm is public in the npm API, and we use that information to verify ownership of npm packages in order to make it easy for npm package owners to apply to receive funding via Gratipay. I had lunch with the npm folks the day after Sustain, and they confirmed that until they implement OAuth this is the best method to use.
>
>What this means is that a user who is signed in to Gratipay can see your email address on this page:
>
>https://gratipay.com/on/npm/[]/
>
[...]

PR #4569, opened 2017-08-14
Pay for package.json in aggregate

whit537:
So far under #4427 we've deployed _discovery_ via `package.json`, but the payment mechanic is the same one we've always had: recurring weekly payments managed individually for each and every package. This PR adds a new payment mechanic (our first in years—ever?): givers can set a single dollar amount to pay for their entire `package.json` in aggregate.

### Todo

- [ ] replace listing with "How much?" prompt
- [ ] add Review page with receipt-y thing + credit card form
- [ ] add cron to process distributions daily
- [ ] add distribution customization UI
- [ ] email receipt
- [ ] email invitation to maintainers
- [ ] email distribution report

### DIFF
--- a/tests/py/test_www_npm_package.py
+++ b/tests/py/test_www_npm_package.py
@@ -62,5 +62,5 @@ class Bulk(Harness):
 
     def test_anon_gets_payment_flow(self):
         body = self.client.GET('/on/npm/').body
-        assert 'Paste a package.json' in body
+        assert 'Paste your own package.json' in body
         assert '0 out of all 1 npm package' in body
--- a/tests/ttw/test_paying_for_package_json.py
+++ b/tests/ttw/test_paying_for_package_json.py
@@ -8,7 +8,7 @@ class Tests(BrowserHarness):
 
     def assertDiscovery(self):
         instructions = self.css('.instructions').text
-        assert instructions == 'Paste a package.json to find packages to pay for:'
+        assert instructions == 'Paste your own package.json:'
 
     def test_anon_gets_discovery_page_by_default(self):
         self.visit('/on/npm/')
--- a/tests/ttw/test_paying_for_package_json.py
+++ b/tests/ttw/test_paying_for_package_json.py
@@ -20,7 +20,7 @@ class Tests(BrowserHarness):
         self.visit('/on/npm/')
         self.assertDiscovery()
 
-    def test_pasting_a_package_json_works(self):
+    def test_paying_for_package_json_works(self):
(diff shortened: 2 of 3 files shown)


whit537:
Email auth would potentially allow us to charge people as "guests" without making them make an account. That moves us even further outside our current flow of funds, however (e.g., what are the fraud implications?). And I think the user experience would actually be fairly close in either case, because the email prompt would come at the end of the flow regardless (you don't want to put it anywhere else because then they have to go to email and come back—a high-drag maneuver). In sketching up the flow I see:

1. Post `package.json` just like we have now.
1. Instead of listing of all packages (grouped by acceptance of funds), prompt for a dollar amount with only a summary of packages discovered ("10 ready to accept payments, 5 not ready, 85 unclear").
1. Prompt for sign-in.
1. Review screen with credit card form on it.
1. After payment, prompt for email address if missing ("To receive a receipt by email and a report in a week ...").

sketch

whit537:
Here's a write-up of what's sketched there:

1. giver posts a package.json
1. giver enters a dollar amount and a distribution (default or custom)
1. giver signs up/in
1. giver reviews payment instruction, enters cc details if necessary
1. on post:
    1. store payment instruction
    1. charge giver immediately and increase their balance
    1. email receipt to giver
    1. if all specified packages are claimed-or-locked: distribute immediately
    1. else: email maintainers of unclaimed-or-locked packages and wait

Daily cron task:

1. review payment instructions
1. for each payment instruction that has been open for a week:
    1. distribute funds based to opted-in receivers
    1. email a final report to giver

whit537:
From slack:

> The basic idea for bulk pledging is that we immediately charge the company the $10,000 (or whatever) and hold onto it for a week while we automatically email the maintainers of unclaimed packages in the set of packages in their package.json. After a week we distribute the $10,000 based on who has claimed their packages by then.
> It’s a new and different flow from what we’ve ever offered before (weekly recurring).
> It’ll take some work to get there, and cleaning up our current npm integration moves us in the right direction.
> The idea is to _scale_.
> Email is the key.
> But we have to tighten up the ship or it’ll come apart at the seams once it achieves any real velocity.
> If we’re already getting phishing against npm [#4557], imagine what we’ll get once we start promoting (via the npm weekly, etc.).

whit537:
Bringing back pledging means bringing back locking/hard-opt-out, which plays into phishing mitigation: https://github.com/gratipay/gratipay.com/issues/4557#issuecomment-324474385.

whit537:
Closing in favor of https://github.com/gratipay/gratipay.com/pull/4598.

PR #4570, opened 2017-08-14
Extract out test_participant_emails.py

whit537:
Split out from #4548 for easy merge and cleaner diff over there.


whit537:
!m @rohitpaulk

Issue #4284, opened 2017-01-09
Process bounces and complaints automatically


whit537:
Reticketing from #4232/#4282. Email bounces and complaints should post back to gratipay.com so we can mark the address as unverified (or otherwise deal with the situation).

whit537:
http://docs.aws.amazon.com/ses/latest/DeveloperGuide/sending-email-with-ses.html
https://aws.amazon.com/blogs/ses/goodbye-blacklist-introducing-the-suppression-list/

whit537:
Eep! Dupe of #2993! 😮

PR #4571, opened 2017-08-14
Limit pending email verifications

### LINKED ISSUE #4545: Send email messages immediately

### LINKED ISSUE #4557: Mitigate npm spam/phishing

whit537:
&larr; Persist email verifications (#4573) — Support queuing emails without participants (#4548) &rarr;

------------------

Fixes #4557.

### Todo

- [ ] Test/think through having more than one unverified email (two cases: merge/take-over, and pre-existing conditions)
- [ ] Review envvars


gratipay/models/participant/email.py
```diff
@@ -106,6 +103,16 @@ def validate_email_verification_request(self, c, email, *packages):
         if not all(email in p.emails for p in packages):
             raise EmailNotOnFile()
 
+        unverified_email = c.one("""
+            SELECT address
```
rohitpaulk: A user could repeatedly add an email + remove it to circumvent this restriction, no? The throttling strategy we use will have to rely on previous state (as you've mentioned in the ticket description)

whit537: We need to start also thinking about not deleting email addresses, because when we start handling bounces and complaints (#4284) we'll want to retain that info for a given email address even if one participant removes it and another (or the same) adds it again. Hmm ...

whit537:
Alright, after wrestling to rebase this onto #4572 I think the way to architect this is to rename `email_addresses` (formerly `emails`) to `email_verifications`, and then make a new `email_addresses` table that holds just `address` as primary key and a nullable `verification_id`.

When a participant adds an email address we file the address in `email_addresses` (if needed) with a null verification, and we start a new verification in `email_verifications` (can have multiple verifications open across users for the same address, but only one per participant per address).

When a participant completes a verification we link to it in `email_addresses.verification_id`. We leave any other pending verifications open in `email_verifications`. We never delete from `email_verifications`.

When a participant removes an email address we null out the verification in `email_addresses`. We keep the record in the table because 

We link to `email_addresses` from `email_messages`.

We link to `email_messages` from `email_verifications`.

whit537:
Something like that, anyway. :)

whit537:
I've kicked out #4573 to persist email verifications. Once done will revisit here to see about constraining based on past verifications.

whit537:
Working up a ttw test so we can get the user requirements here.

screen shot 2017-08-18 at 11 17 43 am

whit537:
Alright this is a mess. First of all the email listing should be styled like our other listings (homepage, search page, social profiles, etc.).

screen shot 2017-08-22 at 2 49 02 pm

whit537:
And after this PR we don't even want UI to add another email address when there is a pending verification.

whit537:
But how does this interact with bulk npm claiming? We shouldn't allow interaction there _either_ if there is a pending verification.

whit537:
But of course we need to protect against multiple verifications in the Python/API layer and not just in the JavaScript/UI.

whit537:
And what about the case where someone legitimately enters a wrong address? They shouldn't be able to remove pending verifications either (UI and API). That's bound to generate some confusion. "I entered the wrong email and now I can't remove it or submit my correct email." It'll have to be a support request with an admin intervention to undo their conditin.

mattbk:
Are you trying to prevent people from ~~spamming~~ email bombing by adding secondary email addresses that point to someone else's address? Or is the verification system so monolithic that there's no way to throttle just one type of verification request?

whit537:
screen shot 2017-08-22 at 3 25 08 pm

gratipay/models/participant/email.py
whit537: > A user could repeatedly add an email + remove it to circumvent this restriction, no? 

Eureka! Turns out it makes the most UX sense to disallow removing emails when verification is pending, to properly communicate the reality that the pending verification still exists. But, if we truly disallow removing emails when verification is pending, then we'll have closed the repeatedly-add-and-remove loophole. Violà!

Proceeding on that basis in #4579 ...

whit537:
> Are you trying to prevent people from ~~spamming~~ email bombing by adding secondary email addresses that point to someone else's address? Or is the verification system so monolithic that there's no way to throttle just one type of verification request?

I'm not sure how to understand these as mutually exclusive. Did our recent slack chat (archiving [stuck](https://twitter.com/Gratipay/status/900315434674683904) 😞 ) clarify this?

whit537:
Closing in favor of https://github.com/gratipay/gratipay.com/pull/4579.

gratipay/models/participant/email.py
rohitpaulk: Umm, what if a person mistakenly enters a wrong email address?

rohitpaulk: > Turns out it makes the most UX sense to disallow removing emails when verification is pending, to properly communicate the reality that the pending verification still exists.

This is also a security risk, as far as I can tell. If we don't allow canceling verification requests - a mistake in entering their email could result in a malicious user taking over their account.

rohitpaulk: For example, github: 

screen shot 2017-08-23 at 7 49 26 pm

Issue #4442, opened 2017-05-03
Clean up exchanges


kaguillera:
-------------------------------

✈️  **This is the flight deck for the [Clean up exchanges](https://github.com/orgs/gratipay/projects/20?fullscreen=true) project.** ✈️ 

-------------------------------

Due to bad schema design we have missing values for some fields in our `exchanges` and `exchange_routes` tables. We use either `unknown` or `NULL` to represent this missing info (we started using `unknown` in https://github.com/gratipay/gratipay.com/pull/3975 in order to make querying easier).

| table | field | value |
|:------|:------|:-------|
| `exchanges` | `ref` | `NULL` |
| `exchanges` | `status` | `unknown` |
| `exchange_routes` | `network` | `unknown` |

## Todo

- [ ] determine whether `network` is corrupt—https://github.com/gratipay/gratipay.com/issues/4442#issuecomment-305864193
- [x] determine whether we are still leaking
  - [x] `ref`—https://github.com/gratipay/gratipay.com/issues/4442#issuecomment-305859075
  - [x] `status`—https://github.com/gratipay/gratipay.com/issues/4442#issuecomment-305868068
  - [x] `network` and `address`—https://github.com/gratipay/gratipay.com/issues/4442#issuecomment-305864193
- [x] add `ref` to record an exchange -- #4518  
- [x] plug last remaining leak: `ref` under MassPay —https://github.com/gratipay/gratipay.com/pull/3912#issuecomment-305855944
-- #4518 
- [ ] backfill old records
  - [ ] `ref`
  - [ ] `status`
  - [ ] `network` and `address`
- [ ] constrain field for non-{`NULL`,`unknown`}
  - [ ] `ref`
  - [ ] `status`
  - [ ] `network`

## Original

Confused. So far.
- There is no payday log for payday 0, payday 1, payday 2
- The payday log for payday run on 2012-06-22 is no where to be found
- There are exchange records in the exchange table for all of the above paydays assuming that 
       - Payday 0 was run on 2012-06-01
       - Payday 1 was run on 2012-06-08
       - Payday 2 was run on 2012-06-15 which according to `gittip-3.log` (payday logs it was)
       - Payday 3 was run on 2012-06-22 but no log exist but there are records of exchanges in the `exchanges` table
       - Payday 4 was run on 2012-06-29 but is actually recorded in `gittip-4.log`
[...]

PR #4572, opened 2017-08-15
Persist email messages

whit537:
&larr; ~Factor out email tests (#4570)~ — Persist email verifications (#4573) &rarr;

------------------

Previously the email queue was ephemeral. We would delete messages from it after successfully firing them off to SES. That prevents us from doing several things:

- properly handling bounces and complaints (#4284)
- sending instantly while also throttling (#4545)
- limiting verification emails properly (#4557)

This PR:

- changes `email_queue` to `email_messages`, a table that we never delete from
- changes `emails` to `email_addresses` to disambiguate

### Todo

- [x] store `MessageId`


whit537:
Ready for review, @rohitpaulk! :-)

whit537:
As we start persisting messages, we should store the `MessageId` [returned](http://docs.aws.amazon.com/ses/latest/APIReference/API_SendEmail.html#API_SendEmail_ResponseElements) from SES for successful submissions. That'll be crucial to bounce handling (#4284), and in general is a Good Idea&trade; (cf. #4442).

whit537:
Already, `MessageId` stored in 36c54650cdc73b79ae58dd988259fca6fea6efea.

rohitpaulk:
Looks good to me. I'm a bit confused as to why `log_every` has to be patched for the test, but that isn't really important compared to the purpose of this patch itself. 

Waiting for Travis.

whit537:
> I'm a bit confused as to why `log_every` has to be patched for the test

Good guestion! That might be noise. 😶

PR #4573, opened 2017-08-16
Persist email verifications

whit537:
&larr; ~~Persist email messages~~ (#4572) — Limit pending email verifications (#4571) &rarr;

---------------------

### Todo

- [ ] Modify schema.
- [ ] Change `DELETE` into `INSERT`.
- [ ] Change `SELECT` to pull latest only.


whit537:
Alright, the goal on this PR is to keep the Python API exactly the same, but have a record of email verifications in the database that is persistent across a) email removal/re-add, *and* b) account close/re-open. Here is the relevant API I find:

```python

class Emails(object):               # gratipay.models.package.emails
    def classify_emails_for_participant(self, participant): pass

class Participant(object):          # gratipay.models.participant
    def clear_personal_information(self, cursor): pass

class Email(object):                # gratipay.models.participant.email
    def start_email_verification(self, email, *packages): pass
    def validate_email_verification_request(self, c, email, *packages): pass
    def get_email_verification_link(self, c, email, *packages): pass
    def get_email_verification_nonce(self, c, email): pass
    def set_primary_email(self, email, cursor=None): pass
    def _set_primary_email(self, email, cursor): pass
    def finish_email_verification(self, email, nonce): pass
    def save_email_address(self, cursor, address): pass
    def get_email(self, address, cursor=None, and_lock=False): pass
    def get_emails(self, cursor=None): pass
    def get_verified_email_addresses(self, cursor=None): pass
    def remove_email(self, address): pass
    def set_email_lang(self, accept_lang): pass

class Packages(object):             # gratipay.models.participant.packages
    def get_packages_for_claiming(self, manager): pass
    def start_package_claims(self, c, nonce, *packages): pass
    def get_packages_claiming(self, cursor, nonce): pass
    def finish_package_claims(self, cursor, nonce, *packages): pass
```

whit537:
Here are the places we delete from `email_addresses` (where verifications are stored as of #4572):

#### `clear_personal_information`

https://github.com/gratipay/gratipay.com/blob/e45ba2b48c92b23e25d328642dcdd7d12980eafa/gratipay/models/participant/__init__.py#L365

#### `take_over`

https://github.com/gratipay/gratipay.com/blob/e45ba2b48c92b23e25d328642dcdd7d12980eafa/gratipay/models/participant/__init__.py#L1113-L1115

#### `remove_email`

https://github.com/gratipay/gratipay.com/blob/e45ba2b48c92b23e25d328642dcdd7d12980eafa/gratipay/models/participant/email.py#L363-L364

#### `save_email_address`

https://github.com/gratipay/gratipay.com/blob/e45ba2b48c92b23e25d328642dcdd7d12980eafa/gratipay/models/participant/email.py#L310-L313

whit537:
So basically we need to remove those `DELETE`s and then fix up the reads to only pull the "correct" verification.

whit537:
Here are the readers:

```python

class Emails(object):               # gratipay.models.package.emails
    def classify_emails_for_participant(self, participant): pass

class Email(object):                # gratipay.models.participant.email
    def validate_email_verification_request(self, c, email, *packages): pass
    def get_email_verification_nonce(self, c, email): pass
    def save_email_address(self, cursor, address): pass
    def get_email(self, address, cursor=None, and_lock=False): pass
    def get_emails(self, cursor=None): pass

class Packages(object):             # gratipay.models.participant.packages
    def get_packages_for_claiming(self, manager): pass
```

whit537:
This is reminiscent of the schema design we use on `payment_instructions`, but there we can signal "no payment instruction" with a zero `amount`. What's the parallel here?

whit537:
Maybe a null `verification_start`?

```sql
CREATE TABLE email_addresses
( id                    serial                      PRIMARY KEY
, address               text                        NOT NULL
, verified              boolean                     DEFAULT NULL
                                                      CONSTRAINT verified_cant_be_false                                                                          -- Only use TRUE and NULL, so that the
                                                        -- unique constraint below functions
                                                        -- properly.
                                                        CHECK (verified IS NOT FALSE)
, nonce                 text
, verification_start    timestamp with time zone    NOT NULL
                                                      DEFAULT CURRENT_TIMESTAMP
, verification_end      timestamp with time zone
, participant_id        bigint                      NOT NULL
                                                      REFERENCES participants(id)
                                                      ON UPDATE RESTRICT
                                                      ON DELETE RESTRICT

, UNIQUE (address, verified) -- A verified email address can't be linked to multiple
                             -- participants. However, an *un*verified address *can*
                             -- be linked to multiple participants. We implement this
                             -- by using NULL instead of FALSE for the unverified
                             -- state, hence the check constraint on verified.
, UNIQUE (participant_id, address)
, CONSTRAINT emails_nonce_key UNIQUE (nonce)
 );
```

whit537:
I think we can get rid of `verified` and just use `verification_end is not null`. I also think we could rename `verification_{start,end}` to `ts_{start,end}` to match `paydays`.

whit537:
... but that also sounds like probable scope creep.

whit537:
Okay! Time to hack! :dark_sunglasses:

whit537:
Over at https://github.com/gratipay/gratipay.com/pull/4571#discussion_r134643620 I think I discovered that we should be able to do this with minimal schema changes. Closing in favor of #4579.

PR #4576, opened 2017-08-17
TestNotifyParticipants uses QueuedEmailHarness 

redixhumayun:
The TestNotifyParticipants class uses QueuedEmailHarness now, and the test passes. The problem was the `self.app.email_queue.flush()` call.

If it was in there, the db was flushed, and the QueuedEmailHarness class' call to `rec = self.db.one('SELECT * FROM email_queue ORDER BY id DESC LIMIT 1')` returned nothing. 

However, removing the flush call, resulted in the db being populated with two rows, and this provided the `TooMany()` error, because the db had two rows instead of one. 

I just moved `self.app.email_queue.flush()` call below the assert statements.


rohitpaulk:
:+1:

@redixhumayun There are a couple of linter errors left (unused imports, quick fixes), I'm ready to merge this once they are fixed.

screen shot 2017-08-17 at 9 38 42 pm

https://travis-ci.org/gratipay/gratipay.com/builds/265620733?utm_source=github_status&utm_medium=notification

PR #4577, opened 2017-08-18
Confirm linking accounts elsewhere

whit537:
It's possible in convoluted circumstances for attackers to trick participants into linking an account elsewhere belonging to the attacker, to the participant's Gratipay account. We should protect against this by adding an additional confirmation step, as Phabricator does:

phabricator

Fix for https://hackerone.com/reports/170684.

### DIFF
--- a/www/on/%platform/associate.spt
+++ b/www/on/%platform/associate.spt
@@ -13,70 +13,85 @@ platform = getattr(website.platforms, request.path['platform'], None)
 if platform is None:
     raise Response(404)
 
-# Get the query id from the querystring
-query_id = platform.get_query_id(request.qs)
+if request.method == 'GET':
 
-# Check that we have a cookie that matches the query id (CSRF prevention)
-cookie_name = (platform.name+'_'+query_id).encode('ascii')
-try:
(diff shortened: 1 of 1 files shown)


whit537:
We already have a confirm page for merging accounts. We want to add another confirm page *before* that one.

whit537:
We have two actions, `opt-in` and `connect`. What's the difference?

whit537:
https://github.com/gratipay/gratipay.com/blob/2101/www/on/%25platform/associate.spt#L53

https://github.com/gratipay/gratipay.com/blob/2101/www/on/%25platform/associate.spt#L66

whit537:
`opt-in` is initial sign-in, `connect` is adding another account once you're already signed in. We certainly need the additional confirmation in the `connect` case. Do we also need it in the `opt-in` case?

whit537:
Yes. Creating a new account with a malicious elsewhere attached is no better than linking a malicious elsewhere to an existing account.

whit537:
Let's see if we can stick with `./associate` as the callback to avoid having to update at all providers.

whit537:
https://github.com/gratipay/inside.gratipay.com/issues/1196.

PR #4579, opened 2017-08-22
Limit pending verifications in emails UI

whit537:
&larr; Implement deploy hooks (#4590) — Redux of #4571 — "Ready to verify" (#4584) &rarr;

-------------

We want to limit pending email address verification attempts to one, to prevent abuse (#4557). We need to do this both in the UI and in the API/db. The former is in some ways easier than the latter so let's try starting with that?

### Todo

  - [x] style listing
    - [x] figure out what to do with action buttons
    - [x] bring in status icons
  - [x] constrain to available actions only
  - [x] get all action buttons working again
  - [x] bring back add email (as button next to header?)
  - [x] ~~confirm on remove~~ &rarr; #4581
  - [x] change base back to `master` after #4590


whit537:
screen shot 2017-08-22 at 5 37 57 pm

whit537:
Whoa! Monster mouse! 🐭  👹 

y7qk328paa

whit537:
I realized at https://github.com/gratipay/gratipay.com/pull/4571#discussion_r134643620 that we can implement verification throttling with no additional schema changes by protecting against removal in the app layer. That's enough for here, though handling unsubscribes will still be a chore ...

whit537:
Squashed and rebased, was 1b1ee1e0f956742849bdc4d608dc2bf6455288ac.

rohitpaulk:
@whit537: Comments added at https://github.com/gratipay/gratipay.com/pull/4571#discussion_r134764987

whit537:
Okay! First chunk of work done in 153469be7f65d191fe2cbedf82a8cbb7d90cf088. Since we're already at ~370 ∆ I will probably scope this just to the email listing, and implement the package page and API changes on other PRs.

-----------------

zlgmqwct3p

whit537:
Picking up from https://github.com/gratipay/gratipay.com/pull/4571#discussion_r134764987 ...

> Umm, what if a person mistakenly enters a wrong email address?

Then they have to contact support. And that raises a good point: admins should be able to remove pending email verifications. If we get enough pressure on support then we consider re-enabling this as self-serve.

> This is also a security risk, as far as I can tell. If we don't allow canceling verification requests - a mistake in entering a user's email could result in a malicious user taking over their account.

Sorry ... how so?

> For example, github: 
>
>screen shot 2017-08-23 at 7 49 26 pm

👍

rohitpaulk:
> Sorry ... how so?

I enter a wrong email address. Person verifies the email, uses it to sign-in?

whit537:
> I enter a wrong email address. Person verifies the email, uses it to sign-in?

Okay, but how do we mitigate that? That's the achilles heel of email for auth, isn't it? :-/

rohitpaulk:
> Okay, but how do we mitigate that? 

Allow the user to remove the email immediately, by themselves.

whit537:
> Allow the user to remove the email immediately, by themselves.

Okay, but then how do we protect against the *other* security risk: remove-and-add to work around verification throttling?

whit537:
Remember that verification links expire after 24 h anyway. We could tighten that to five minutes once we have instant send (#4545).

rohitpaulk:
> Okay, but then how do we protect against the other security risk: remove-and-add to work around verification throttling?

Can't we maintain persistent history of sent emails separately, and use that info to throttle? Canceling of an email verification link doesn't have to necessary delete *ALL* history of an email every being sent.

whit537:
Sorry, throttling isn't the best word. The way we're headed here is to do both:

- Throttle as we have done (though it'll look a little different with #4545).
- Limit the number of pending verifications.

whit537:
I think we should tighten the verification window to five minutes: that will protect against "honest mistakes" turning into inadvertent phishing.

I think we should only allow one pending verification at a time with no option to remove. Our use case is different than GitHub's because we have _all npm maintainer emails_ in our system. We're playing with fire. 😳 🔥

rohitpaulk:
Hmm, I see what you're getting at.

rohitpaulk:
> Our use case is different than GitHub's because we have all npm maintainer emails in our system

I understand the 'limit pending verifications' concern. However, what does 'having all npm maintainer emails in our system' have to do with user email verification? Is this because it is tied to claiming packages?

whit537:
> Is this because it is tied to claiming packages?

Yes. On my next PR I'll be modifying the UI on pages like https://gratipay.com/on/npm/express to limit the ability to initiate package claims when there are existing verifications (~= package claims) outstanding.

whit537:
We've had three users already report spam/phishing to us (#4557), so this isn't theoretical.

rohitpaulk:
I still think that we should decouple email verification and linking packages. What I propose is: 

- Only allow users that own a verified email that is listed in npm to 'claim' a package. 
- Once we do have a verified email that corresponds to an npm package, allow user to claim the package directly, and send a notice (not verification, just a notice) out to all maintainers.

Now that the email only says 'link abcd@gmail.com to the abcd account on Gratipay' - it is less dangerous. As all services do, we'll just add a note saying 'If this wasn't you, no action is required' at the bottom. Generic sign-up email spam is understandable, and won't freak users out like saying 'May we receive money on your behalf'.

whit537:
Okay, so can we hash out the user flow on https://gratipay.com/on/npm/express, then? Do we display all maintainer emails and only allow the user to initiate a message to addresses they've already verified? Do we only display email addresses they've already verified? Do we offer to send an email verification right from the package page? Do we make them click over to the Emails page to add the address there?

rohitpaulk:
> Okay, so can we hash out the user flow on https://gratipay.com/on/npm/express, then? Do we display all maintainer emails and only allow the user to initiate a message to addresses they've already verified? Do we only display email addresses they've already verified? Do we offer to send an email verification right from the package page? Do we make them click over to the Emails page to add the address there?

Basics, just to get things working: 

- If user is a maintainer, show 'Claim' button directly. Make it clear that the user is allowed to claim this package because they own that specific address, and also show the list of maintainers that we'll send emails to if they claim the package. 
- If the user is not a maintainer, show all emails, say that you have to add one of these (link to emails page) as a verified email address on Gratipay. Once verified, come back to this page and you'll be able to claim a package. 

Streamlining (building upon the above): 

- When an email is verified, automatically fish out unclaimed packages related to that address, and take them to the claim page? (or a bulk claim page?)
- Add 'link/verify' buttons to the emails on the 'claim' page, without requiring the user to go to the emails page and come back.

whit537:
> send a notice (not verification, just a notice) out to all maintainers.

That's #4425.

whit537:
Actually, tightening scope a little more, even. Bumping confirm on remove to #4581 since we're over 400 ∆ here and we don't have confirm-on-remove as it stands and I want to do that in a modal along with confirm of send on package pages.

This is ready for review/merge, @rohitpaulk et al.! :-)

whit537:
Rebased, was 3b6b14ef5cea58d2a9b8f798163d8e693ff2b6f3.

whit537:
Rebased onto #4590, was 4f3393252ccbc69db79773e28e2eee87178b3c81.

whit537:
Rebased back onto `master`, was a6b431561383a4e75dc7bfd7b93b6a5171179cea.

www/~/%username/emails/index.spt
```diff
@@ -1,12 +1,18 @@
...
 banner = '~' + participant.username
-title = _("Emails")
+title = _("Email")
 
 emails = participant.get_emails()
+groups = [[], [], []]  # primary, verified, unverified
+for email in emails:
+    i = 0 if email.address == participant.email_address else 1 if email.verified else 2
+    groups[i].append(email)
+emails = sum(groups, [])
```
rohitpaulk: 😃 Is this common usage? 

Perils of not having `flatten` in the std lib? :P

whit537: Yes, that's how to spell `flatten` in Python. 🐭

www/~/%username/emails/index.spt
```diff
@@ -1,12 +1,18 @@
...
 
 participant = get_participant(state, restrict=True)
 banner = '~' + participant.username
-title = _("Emails")
+title = _("Email")
 
 emails = participant.get_emails()
+groups = [[], [], []]  # primary, verified, unverified
+for email in emails:
+    i = 0 if email.address == participant.email_address else 1 if email.verified else 2
```
rohitpaulk: We're essentially doing a `sort_by(func)` action here, can't we use https://docs.python.org/3/library/functions.html#sorted with the `key` param? 

i.e. something like: 
```python
def get_sort_key(email):
  if email.address == participant.email_address:
    return 0
  elif email.verified:
     return 1
  else:
    return 2

sorted_emails = sorted(emails, key=get_sort_key)
```

rohitpaulk: Looks less cryptic to me than the awkward spelling of `flattern`, and the inlined inline `if`

tests/ttw/test_email_verification.py
```diff
@@ -0,0 +1,160 @@
+# -*- coding: utf-8 -*-
+from __future__ import absolute_import, division, print_function, unicode_literals
+
+import re
+from gratipay.testing import BrowserHarness, QueuedEmailHarness
+
+
+class Row(object):
```
rohitpaulk: Neat!

rohitpaulk:
Rebased, previous head was @c3452ab8472658a42f2acdefa5cd865bbb645f33.

www/~/%username/emails/index.spt
rohitpaulk: Done in https://github.com/gratipay/gratipay.com/pull/4579/commits/11a93a4b3943a1ed5574d6cb1e11f46d20e65c4c

rohitpaulk:
@whit537 This looks good to me. Ready to merge?

whit537:
@rohitpaulk Yes!

PR #4581, opened 2017-08-24
Confirm some things

whit537:
Follow-on from #4579.


whit537:
Closing per https://github.com/gratipay/inside.gratipay.com/issues/1196.

PR #4588, opened 2017-08-25
Refresh the README ... and Makefile

whit537:
- brush up the welcome header
- add a codebase overview
- move some sections around
- update the initial output and screenshot
- solidify `make test`
- drop bountysource badge

PR #4590, opened 2017-08-31
Implement deploy hooks

whit537:
Limit pending verifications (#4579) &rarr;

----------

Currently we do database migrations in a `sql/branch.sql` file that we then append to `schema.sql`. This is okay for basic migrations, but for more complex migrations (e.g. #4548) we need multiple SQL steps with a Python step in between. This PR achieves this with a new "deploy hooks" mechanism, files in a new `deploy` directory:

- `before.sql`
- `after.py`
- `after.sql`

Furthermore, it adds infrastructure for testing deploy hooks, so that we can better trust our migrations (more necessary as they become more complex).

With this increased flexibility, we should no longer need maintenance mode during deployment, so that's dropped here.

Furthermore, this seems like a good time to stop managing `schema.sql` manually with diffing against production after the fact to avoid drift. Why not just dump production back to `schema.sql` after deployment?


whit537:
Ping @rohitpaulk @clone1018.

whit537:
Blech, blach. Running a script as a subprocess that imports gratipay.wireup ends up removing the compiled asset files, which causes other tests to fail.

whit537:
Reticketed as #4592 and hacked around in 118b310d494afe691173445a237f01e5b6df26ce.

whit537:
Alright, we're green here finally. Rebased, was ae229fdc94d4451db59bcc81c5af8350163719cb.

whit537:
∆ for 5194bbffafb21e3f4313b744b8fa17e9cb4c111f is 2,852 — that's the schema.sql dump from production.

∆ for 08b8a3efc25d878a8e2c4ad1e0e86a623434d0f7 is only 277.

whit537:
Even though we don't need this until #4548 (three hops away), I think we should roll this out now because it also follows on the Makefile cleanup in #4588, and there's bound to be some regressions for the first few deploys using the modified system. Better to start shaking those out asap.

deploy.sh
```diff
@@ -60,61 +65,42 @@ fi
...
-[ "$run_sql" = "before" ] && heroku pg:psql -a gratipay <sql/branch.sql
+# Deploy to Heroku, with hooks
+cd deploy
+[ -e before.sql ] && heroku pg:psql -a gratipay < before.sql
+[ -e before.py ] && heroku run -a gratipay python < before.py
 git push --force heroku master
-[ "$maintenance" = "yes" ] && heroku maintenance:off -a gratipay
-[ "$run_sql" = "after" ] && heroku pg:psql -a gratipay <sql/branch.sql
-rm -f sql/branch.sql
+[ -e after.py ] && heroku run -a gratipay python < after.py
```
whit537: This is the wrong way to do this (cf. https://github.com/gratipay/inside.gratipay.com/issues/1149#issuecomment-326552930). Since the `after.py` script exists on the dyno we should just run it directly there:

> `[ -e after.py ] && heroku run -a gratipay python deploy/after.py`

deploy.sh
```diff
@@ -60,61 +65,42 @@ fi
...
 git tag $version
 
 
-# Deploy to Heroku
-[ "$maintenance" = "yes" ] && heroku maintenance:on -a gratipay
-[ "$run_sql" = "before" ] && heroku pg:psql -a gratipay <sql/branch.sql
+# Deploy to Heroku, with hooks
+cd deploy
+[ -e before.sql ] && heroku pg:psql -a gratipay < before.sql
+[ -e before.py ] && heroku run -a gratipay python < before.py
```
whit537: 🙈

kaguillera:
LGTM

PR #4597, opened 2017-09-01
Fix syntax error in updated deploy.sh

whit537:
Follow-on from #4590.


kaguillera:
Once it is green with travis I will merge

PR #4603, opened 2017-09-07
Tweak a couple tiny things

whit537:


whit537:
Another slow pitch, @mattbk.

PR #4620, opened 2017-09-10
Fix h1-109161.

EdOverflow:
https://hackerone.com/reports/109161


whit537:
Deployed. We good @EdOverflow?

EdOverflow:
Yep, all good now. Thank you, @whit537.

PR #4626, opened 2017-09-12
Relicense as MIT

whit537:
https://github.com/gratipay/inside.gratipay.com/issues/1161

### DIFF
--- a/COPYING
+++ b/COPYING
@@ -1,151 +1,19 @@
-Gratipay is in the public domain, via the "CC0 1.0 Universal - Public Domain
-Dedication" from the Creative Commons.
+Copyright 2012-present Gratipay, LLC
 
+Permission is hereby granted, free of charge, to any person obtaining a copy of
+this software and associated documentation files (the "Software"), to deal in
+the Software without restriction, including without limitation the rights to
+use, copy, modify, merge, publish, distribute, sublicense, and/or sell copies
+of the Software, and to permit persons to whom the Software is furnished to do
+so, subject to the following conditions:
 
-## Summary of "CC0 1.0 Universal - Public Domain Dedication"
+The above copyright notice and this permission notice shall be included in all
+copies or substantial portions of the Software.
 
-http://creativecommons.org/publicdomain/zero/1.0/
-
-No Copyright
- 
-The person who associated a work with this deed has dedicated the work to the
-public domain by waiving all of his or her rights to the work worldwide under
-copyright law, including all related and neighboring rights, to the extent
-allowed by law.
-
-You can copy, modify, distribute and perform the work, even for commercial
-purposes, all without asking permission. See Other Information below.
-
-
-Other Information
-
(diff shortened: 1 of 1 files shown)


whit537:
You cool with this @clone1018?

clone1018:
I like the MIT license. Are there any special processes going from public domain to MIT?

whit537:
> Are there any special processes going from public domain to MIT?

According to @massonpj in irl convo there are not.

Changaco:
This might not have gone over well if Gratipay was a library or app that other projects depend on and which is packaged by third parties, but since it isn't you probably won't hear anyone complaining.

mattbk:
Is that because the MIT license requires derivatives to also be MIT license? I'm new at this stuff.

Changaco:
@mattbk No, the MIT license isn't a [copyleft](https://en.wikipedia.org/wiki/Copyleft) license.

mattbk:
Thanks @Changaco, you made me read it again, slowly. :grin:  :+1:

whit537:
Rebased. Good to merge, @clone1018?

whit537:
💃

massonpj:
Hi all. Thanks @whit537 for introducing this. I am happy to help if anything is needed. We'd like to support your work, but would only be able to if the projects carry an OSI approved license, not public domain (https://opensource.org/faq#public-domain) or CC0 (https://opensource.org/faq#cc-zero).

whit537:
Thanks, @massonpj.

If you're referring to our [own code](https://github.com/gratipay/), then we're almost there: this PR is merged and relicenses our main repo, `gratipay.com`. We have [two more minor repos outstanding](https://github.com/gratipay/inside.gratipay.com/issues/1161#issuecomment-328992725), should be easy now that this PR is in.

If you're referring to projects that we would support with funds collected via the new form on [our homepage](https://gratipay.com/), then that's fine, we can definitely enforce that restriction there.

If you're referring to the [almost 800 projects](https://gratipay.com/browse/projects?status=approved) that are already on Gratipay under our old model (weekly micro-payments direct to the project), we might need to drop to a phone call to sort that out. Should we schedule a call? Maybe we drop to email to schedule? chad@gratipay.com

massonpj:
Hey @whit537,

Great to hear you're migrating gratipay code to an OSI approved license--we love that--and thus makes it easy for us to support the project as a whole. 

We'll also be happy to promote/support your new forum for projects that carry an OSI approved license. 

Regarding the others from the old model, yeah we'd have to delineate those, and quarantine them from any OSI branding, promotion, support. But it's not a problem at all. If we can work with Microsoft and Adobe to support their open source efforts, while avoiding their proprietary stuff, we can make it work here too. 

We do not want to be a burden/hassle to work with, but yeah let's get on a call. I will email you to set that up.

whit537:
Sounds good, @massonpj. Got your mail and looking forward to chatting on Tuesday!

Issue #4633, opened 2017-09-14
Paging links broken on browse


whit537:
They still go to `/`.

https://gratipay.com/browse/projects

Regression from #4598.

rohitpaulk:
Resolved in #4635

PR #4635, opened 2017-09-15
Fix project paging (#4633)

whit537:
Quick fix. Eh, @mattbk @rohitpaulk @aandis @dmk246 et al.?


mattbk:
Can merge when green.

PR #4638, opened 2017-09-16
Homepage: Clear errors on resubmission

rohitpaulk:
**Steps**

- User makes multiple errors when filling the homepage form
- User fixes one error, submits form

**Expected:** 

The error highlighting and text should only be shown for the field that still has an error. 

**Actual**: 

Errors are still visible on both fields.


whit537:
Rebased, was a372459fb7c5c1603c72a748c6ea144813ed12c0. Will merge when green ...

whit537:
Hrm, weird failure. :-/

Will have to dig deeper ...

whit537:
Timing weirdness. Restarting for now ...

PR #4650, opened 2017-09-27
Soft-launch #BackTheStack!

whit537:
This evolves the "Wait ... What? Why?" section of the homepage into a "Campaign" section, hidden behind a querystring var until October 1.

screen shot 2017-09-27 at 12 01 29 pm

PR #4651, opened 2017-09-27
Hide monthly option

whit537:
Part of https://github.com/gratipay/inside.gratipay.com/issues/1160 and includes #4650.

Producing a newsletter is going to be work! Let's not overcommit ourselves.

P.S. I also looked at https://stackoverflow.com/a/39878233 for removing the value from the enum, but it left the enum with size 4 still vs. 3 as I'd expect and it made me not want to muck with pg system tables.

- [x] change base back to master

### DIFF
--- a/gratipay/homepage.py
+++ b/gratipay/homepage.py
@@ -46,8 +46,8 @@ def _parse(raw):
         errors.append('email_address')
 
     follow_up = x('follow_up')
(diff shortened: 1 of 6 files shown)


mattbk:
LGTM, merge when green.

whit537:
Rebased now that #4650 is in, base changed to master.

PR #4658, opened 2017-09-30
Make a big mental shift

whit537:
This is it, the mental shift we [need to make](https://github.com/gratipay/inside.gratipay.com/issues/1160#issuecomment-332998643) to get out of our own heads and into our customer's.

"Pay for open source" is about _our_ problems and motivations.

"Invest in open source" is about _their_ problems and motivations.

### DIFF
--- a/tests/ttw/test_homepage.py
+++ b/tests/ttw/test_homepage.py
@@ -39,7 +39,7 @@ class Tests(BrowserHarness):
 
 
     def test_loads_for_anon(self):
-        assert self.css('#banner h1').text == 'Pay for open source.'
+        assert self.css('#banner h1').text == 'Invest in open source.'
         assert self.css('#header .sign-in button').html.strip()[:17] == 'Sign in / Sign up'
 
     def test_redirects_for_authed_exclamation_point(self):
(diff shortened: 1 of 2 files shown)


deploy/before.sql
```diff
@@ -0,0 +1,3 @@
+BEGIN;
+    ALTER TABLE payments_for_open_source RENAME TO investments_in_open_source;
```
rohitpaulk: @whit537 I understand that you're experimenting here, but aren't there quicker/cheaper/safer ways to achieve this than having to rename this throughout the codebase? 😕

rohitpaulk: My suggestion would be to only tinker with copy. Modeling investments as `payments` in code is forgivable, IMO - not worth spending your time changing

whit537: Fair enough. I've pushed 49c96e72efd71de6a362846cf695dd73f6cba5ba off, we can come back to it later if @JessaWitzel wants to. :D

mattbk: Agree with @rohitpaulk.

whit537:
@JessaWitzel Merge? 😄

whit537:
Okay stopping. [slack](https://gratipay.slackarchive.io/gratipay/page-100/ts-1506793372000009) 🐭

PR #4662, opened 2017-09-30
Replaces temp 'email2' field with 'email' when verifying emails

### LINKED ISSUE #4541: Replace temporary 'email2' param with 'email' when verifying emails
In https://github.com/gratipay/gratipay.com/pull/3965, we changed the value of the `email` parameter to be query-string encoded instead of raw text. To make this backward compatible, we changed the parameter name from `email` to `email2` and retained support for both. 

This was a **long** time ago, time to clean up. 

**Current**: We support `email2` (if the value is encoded) and `email` (if it isn't encoded. 
**End Goal**: Support only `email`, with encoded value

To achieve this without breaking existing links, we'll need an intermediary step. 

- Add support for `email`, with encoded value + send `email` in the links instead of `email2`. Don't drop support for `email2` as of yet. 
- Once the above is done, wait for the duration that verification links live - and then drop support for `email2`. 

Here's the related code: https://github.com/gratipay/gratipay.com/blob/master/www/~/%25username/emails/verify.html.spt#L14

pqrth:
* Fixes #4541

### DIFF
--- a/www/~/%username/emails/verify.html.spt
+++ b/www/~/%username/emails/verify.html.spt
@@ -13,6 +13,12 @@ participant = get_participant(state, restrict=False)
 banner = '~' + participant.username
 if participant == user.participant:
     if 'email2' in request.qs:
+
+        # As of GH-4662, we've renamed the temporary 'email2' field, introduced
+        # in GH-3965, with 'email' and this can be removed when we're confident
+        # that there are no verification links live with field 'email2' in
+        # their query string.
+
         email_field = 'email2'
     else:
         email_field = 'email'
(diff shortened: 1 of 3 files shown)


www/~/%username/emails/verify.html.spt
```diff
@@ -12,15 +12,11 @@ from gratipay.models.participant import email
...
 banner = '~' + participant.username
 if participant == user.participant:
-    if 'email' in request.qs:
-
-        # Deprecated in GH-3965. We shouldn't be generating any more of these,
-        # and this can be removed when we're confident that there are no more
-        # old-style verification links hanging out there.
-
-        email_address = request.qs['email']
+    if 'email2' in request.qs:
```
whit537: Let's add a comment similar to the one we're removing to explain what's going on here.

pqrth: @whit537 Done.

PR #4656, opened 2017-09-28
Collect logos for promoting companies

whit537:
Part of https://github.com/gratipay/inside.gratipay.com/issues/1160.

- [x] fix vertical spacing
- [x] store image files in `.insert`—maybe share code w/ `Team.save_image`?
- [x] process upload in `homepage.py` (make sure it's optional)
- [x] serve images, I guess at `/browse/payments/%uuid/logo`?
- [x] ~~change base back to master~~

PR #4671, opened 2017-10-04
Video, etc.

whit537:
Part of https://github.com/gratipay/inside.gratipay.com/issues/1160. 💃 

## Todo

- [x]  change base back to master

### DIFF
--- a/js/gratipay/homepage.js
+++ b/js/gratipay/homepage.js
@@ -5,11 +5,6 @@ Gratipay.homepage.initForm = function(clientAuthorization) {
     self.$form = $('#homepage #content form');
     self.$submit = self.$form.find('button[type=submit]');
 
-    $('a.what-why').click(function(e) {
-        e.preventDefault();
-        $('#what-why').slideToggle();
-    });
-
     function disable(e) {
         e.preventDefault();
         self.$submit.prop('disabled', true);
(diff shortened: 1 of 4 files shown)


whit537:
Started looking at pulling updates from Medium.

https://gratipay.news/feed
https://stackoverflow.com/questions/10943544/how-to-parse-an-rss-feed-using-javascript

No CORS so we'll have to do it server-side, which means either doing it manually or caching so we don't slow down page load too much.

whit537:
Actually though let's not block this on https://github.com/gratipay/gratipay.com/pull/4656. It's too cool let's get it deployed! 💃 

cc: @mattbk @JessaWitzel et al.

whit537:
Pause! @ChrisECJohns is looking at fixing up the bit where I say "$250" instead of "$250 per developer."

whit537:
Rendering one more once. [slack](https://gratipay.slackarchive.io/gratipay/page-100/ts-1507320811000295)

whit537:
Planning to merge per @mattbk's approval once right video is in place ...

whit537:
Rebased and switched to final video. Was 2e5eac74e5a2477bfc1c16fdb9d7cf041cc4303d.

whit537:
Merge approval because @mattbk is traveling: [slack](https://gratipay.slackarchive.io/gratipay/page-100/ts-1507322395000302).

PR #4676, opened 2017-10-07
set httponly=True for csrf token

### LINKED ISSUE #4561: Set httpOnly cookies.
Using the HttpOnly flag when generating a cookie helps mitigate the risk of client side script accessing the protected cookie (if the browser supports it). [[1]](https://www.owasp.org/index.php/HttpOnly#What_is_HttpOnly.3F)

rozuur:
Only instance where httponly=False is set for cookies is csrf token, I hope this fixes #4561

### DIFF
--- a/gratipay/security/csrf.py
+++ b/gratipay/security/csrf.py
@@ -69,6 +69,4 @@ def add_token_to_response(response, csrf_token=None):
     """Store the latest CSRF token as a cookie.
     """
     if csrf_token:
-        # Don't set httponly so that we can POST using XHR.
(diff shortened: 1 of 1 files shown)


0xibram:
@rozuur setting httpOnly on the `csrf_token` cookie prevents the [JS code](https://github.com/gratipay/gratipay.com/blob/4b5b0b3e287424a3e5b4179d49eddac44064f7e4/js/gratipay/forms.js#L27) from accessing it, which is required... to prevent CSRF.

PR #4678, opened 2017-10-27
Add support for account migration to Liberapay

Changaco:
Per <https://github.com/gratipay/inside.gratipay.com/issues/1196#issuecomment-336618468>. As of now this PR only adds hidden infrastructure, it doesn't touch the UI. I don't think I should be the one who writes the page explaining Gratipay's shutdown.

The new migration endpoint uses normal participant authentication as well as a shared secret which provides proof that the request is coming from Liberapay.


mattbk:
!m @Changaco

Changaco:
The Liberapay side still needs some work on the UI. I'll finish it tomorrow.

Changaco:
Here it is: https://github.com/liberapay/liberapay.com/pull/773

Changaco:
This is ready for review and merge, as is the Liberapay side.

defaults.env
```diff
@@ -108,3 +108,6 @@ PROJECT_REVIEW_TOKEN=
 RAISE_SIGNIN_NOTIFICATIONS=no
 
 LOAD_BRAINTREE_FORM_ON_HOMEPAGE=no
+
+LIBERAPAY_BASE_URL=http://localhost:8339
+LIBERAPAY_SECRET=fake
```
mattbk: Do I need something here to test? Or do I need Liberapay running locally too?

Changaco: If you want to test the full migration process then yes, you need both Gratipay and Liberapay running locally.

mattbk: Thanks! Working on it...

mattbk:
I'm still getting
```
400
Bad Request
Missing key: u'secret'
```
at /about/me/migrate. Running both Gratipay (this branch) and Liberapay (master) locally on default ports.

Changaco:
You have to add `?secret=fake` to the URL. (You don't actually need Liberapay running for this part.)

mattbk:
Okay, so that gives me a JSON of my various account data.

mattbk:
Does the form template in `migrate.html` get used?

Changaco:
@mattbk No. As I said in the original comment this PR is only hidden infrastructure.

Changaco:
I've deployed account migration on Liberapay, to whom and how should I send the real shared secret for the `LIBERAPAY_SECRET` environment variable?

mattbk:
Thanks @Changaco for your patience.

whit537:
> I've deployed account migration on Liberapay, to whom and how should I send the real shared secret for the `LIBERAPAY_SECRET` environment variable?

I guess that would be me. Is email secure enough, do you think?

whit537:
> Does the form template in `migrate.html` get used?

@mattbk Once this PR is landed we can utilize that template on the homepage in #4679.

Changaco:
> I guess that would be me. Is email secure enough, do you think?

Yes. I've just sent it.

Changaco:
Before showing off the migration button it would be a good idea to test that it actually works, all the way to charging a card in USD if possible. Any volunteers, or should I do it?

rohitpaulk:
I did a test run... Looks like I hit a 500.

screen shot 2017-11-02 at 9 40 51 am

rohitpaulk:
I like the clear explanation of what can/cannot be imported. :+1:

Changaco:
> I did a test run... Looks like I hit a 500.

Should be fixed now.

